PR #7, opened 2017-10-25
Handled minDate and maxDate props for DatePicker component

ingro:
I've seen that those props where present in DatePicker but not in DatePickerModal (which is exported), so I updated the Modal and Calendar components to handle them.


dmtrKovalenko:
👍  Thank you for contribution!
I will check it a bit and will publish new release asap

PR #8, opened 2017-10-25
[Proposal] Move moment to peerDependencies

ingro:
Hello again! Actually (as far I understand) it's impossible to set moment locale in a project that use this library and see the Datepicker localized because moment is bundled in `material-ui-pickers`. So I'm proposing to move `moment` to peerDependencies, so it will be possible to do this in the root of your project to localize all your Datepickers:

```javascript
import moment from 'moment';
import 'moment/locale/it';

moment.locale('it');
```

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "material-ui": "^1.0.0-beta.17",
+    "moment": "^2.19.1",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
     "react-scripts": "1.0.14"
--- a/package.json
+++ b/package.json
@@ -34,10 +34,10 @@
     "react-dom": "^16.0.0",
     "prop-types": "^15.6.0",
     "material-ui": "^1.0.0-beta.17",
-    "classnames": "^2.2.5"
+    "classnames": "^2.2.5",
+    "moment": "^2.19.1"
   },
   "dependencies": {
-    "moment": "^2.19.1",
     "moment-range": "^3.0.3"
   },
   "scripts": {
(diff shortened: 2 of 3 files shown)


dmtrKovalenko:
Maybe it would be more convenient to keep moment in dependencies,  because it would no be compiled several times, and also if user have no need in moment, he wouldnt required to install it.
Maybe pass locale as prop to the component?
And also I`ll check if its possible to change moment locale from project root

ingro:
I agree it's not a simple decision, it has some tradeoffs. 

Sure for people that don't use moment in their project would be easier to keep it as a direct depenency, even though they will still "pay" it's cost in kb since it's bundled with material-ui-pickers.
But moment it's widely adopted and for people already using it in their project that would be mean pay it's cost two times, and that's not good since moment isn't lightweight at all.

Another solution may be to dump moment and use something more lightweight and composable like [date-fns](https://date-fns.org/), which will result in a way more small bundle size.

marcoturi:
You should go by date-fns, it's way better for bundle size.

sakulstra:
Any progress on decision making here?
For us this library is currently not usable as we need localized dates, but already have a peer of moment in the project. So moment.locale('whatever') changes the wrong instance.

I like the idea of removing moment and switching to date-fns, but eventually we should wait till 2.0 and tz support is there. For now moving moment to peer seems like a easier solution.

dmtrKovalenko:
Hmm, I was hope that'd it should be resolved by webpack side (that moment should be used from the project deps, instead of current one) 
So if needs I will move it to peer in the next release

sakulstra:
Using this in isobuild not webpack so dunno if/how it works with other build systems, but it seems like other packages relying on moment marked it as external/peer to avoid this kind of problems with different build systems.

dmtrKovalenko:
I`m actually using this package in one of another project, and I`ve analyzed bundle and get there is just only one instance of moment, so webpack have successfully resolved duplicated moments. 
But if for some systems its causing duplicated modules - it should be placed in the peer dependencies.

dmtrKovalenko:
#59

PR #35, opened 2017-11-06
Allow for changes in TextField to toggle Picker

GaryPWhite:
Feedback from users of a tool I'm developing with material-ui-pickers is they wish they could trigger the date picker without having to click on it. Perhaps onFocus/onChange trigger could be better? I'm willing to discuss this in an issue. I didn't want to open an issue without a solution :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,16 @@ Available as npm package
 npm install material-ui-pickers -S
 ```
 
-We are using material-ui-icons font to display icons. We are working on additional way to pass icons,but for now its required. Just add this to your html 
+We are using material-ui-icons font to display icons.Just add this to your html 
 ```html
 <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
 ```
-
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
Hi, thank you for taking your time!
Yep I think it would be helpful for somehow to have an ability to open picker with tab, so I`ll merge your PR, just wanna ask you to update your target branch to develop

GaryPWhite:
sorry about the delay. Changed the case branch!

GaryPWhite:
Sidenote: If you'd like, I wouldn't mind putting in another PR (on develop :)) to add [react-keydown](https://www.npmjs.com/package/react-keydown) and record when the tab key is pressed to trigger the datepicker. Unless you meant when they tab INTO the datepicker, in which case, I think onFocus is the most relevant solution.

dmtrKovalenko:
Thank you, I think we shouldnt bind an keydown event to the datepicker, and the open on focus would be sufficient

Issue #36, opened 2017-11-07
Per day classes and DayProps prop


noah-potter:
Would you approve a PR that added a handler that could attach extra classes to each rendered day? Something like:

```jsx
dayClassName(date) {
  return classnames({
    [classes.highlight]: isWeekend(date),
  })
}

<DatePicker dayClassName={dayClassName} />
```

This function would be ran when each day's className was calculated.

Also, would you support a PR that added a `DayProps` prop that would be spread onto each day's IconButton?

sakulstra:
I would recommend to instead have a `renderDay = (day, selected) => {return <Whateveryoulike />}` function which can be passed as a prop (similar to what https://github.com/moroshko/react-autosuggest does)

dmtrKovalenko:
@noah-potter @sakulstra 
I think we can implement both, because it would not rise the bundle size, and from other side add some flexibility of rendering. I think also would be helpful to have something like `shouldDisableDate` function, allows to disable particular dates. 
Would you like to open a pull request?

noah-potter:
I'll take a stab at it

PR #44, opened 2017-11-08
Merge remote-tracking branch 'master/develop' into feature/add-custom…

noah-potter:

# 
#36 

## Description

- Add `labelFunc` prop to DateTextField. This allows for custom messages to be rendered in the TextField. This is helpful when the picker may no longer affect only one date (see week picker added to the demos) and when users need more flexibility.
- Add `renderDay` prop to Calendar. This allows a user to override how each day is rendered in the calendar.
`renderDay(date: moment, selectedDate: moment, dayInCurrentMonth: bool, dayComponent: ReactElement)`
- Add a wrapper div around each rendered day. This allows the Calendar to set the element key and listen for click events without exposing that interest when using the `renderDay` hook
- Add support for these two props to DatePicker and DateTimePicker
- Add examples using the two props to the demos page

### DIFF
--- a/docs/src/Demo/Demo.jsx
+++ b/docs/src/Demo/Demo.jsx
@@ -14,11 +14,15 @@ class Demo extends Component {
   }
 
   state = {
-    selectedDate: new Date(),
+    selectedDate: moment(),
   }
 
   handleDateChange = date => {
     this.setState({ selectedDate: date })
+  }
+
+  handleWeekChange = date => {
+    this.setState({ selectedDate: date.clone().startOf('week') })
   }
 
   scrollToContent = () => {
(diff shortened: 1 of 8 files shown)


dmtrKovalenko:
Hi, thank you so mutch.
Its the best pull request, I have seen in this project)
I'll review and test it a bit and publish release (maybe beta) on the weekend

noah-potter:
I wonder if it would be better if the `format` prop could replace the `labelFunc` I created here. So `format` would accept a string or a function. What do you think @dmtrKovalenko ?
Also, some of the stuff in the demo is incorrectly named. I went through several iterations and it looks like some of the current code references old naming. I'll submit another commit that makes the naming in the demo more consistent.

PR #79, opened 2017-11-25
Close on clickaway

alitaheri:
I believe this provides a better UX than having to explicitly close the dialog.


alitaheri:
@dmtrKovalenko Should I move this to the develop branch too?

dmtrKovalenko:
@alitaheri  
Thank you, its just really helpful to have ability not update production readme for a future release. 😄

PR #88, opened 2017-12-03
make onAccept and onDismiss optional when using ts

Stevearzh:

# 

## Description

when i use material-ui-picker in a tsx like below:

```typescript
import { DatePicker } from 'material-ui-pickers/src';

...

<DatePicker
  value={this.state.selectedDate}
  onChange={this.handleDateChange}
/>

...
```

the compiler would throw an error and told me `onAccept` and `onDismiss` is required, however if i add `onAccept` and `onDismiss` like below:

```typescript
...

handleDateAccept = (event: React.MouseEvent<HTMLElement>): void => undefined;
handleDateDismiss = (event: React.MouseEvent<HTMLElement>): void => undefined;

...

<DatePicker
  value={this.state.selectedDate}
  onChange={this.handleDateChange}
  onAccept={handleDateAccept}
  onDismiss={handleDateDismiss}
/>

...
```

then `onChange` method won't work after I pick a new date, so i checkout the `ModalWrapper.d.ts `, so modified the interface...

### DIFF
--- a/lib/src/wrappers/ModalWrapper.d.ts
+++ b/lib/src/wrappers/ModalWrapper.d.ts
@@ -5,8 +5,8 @@ import { Utils } from '../utils/utils';
 import {DateTextFieldProps} from '../_shared/DateTextField';
 
 export interface ModalWrapperProps extends DateTextFieldProps {
-    onAccept: ButtonProps['onClick'];
-    onDismiss: ButtonProps['onClick'];
+    onAccept?: ButtonProps['onClick'];
+    onDismiss?: ButtonProps['onClick'];
     dialogContentClassName?: string;
     okLabel?: ReactNode;
     cancelLabel?: ReactNode;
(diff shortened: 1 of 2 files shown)


alitaheri:
You should also update the [prop-types](https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/src/wrappers/ModalWrapper.jsx#L13) and add null check [here](https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/src/wrappers/ModalWrapper.jsx#L41) and [here](https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/src/wrappers/ModalWrapper.jsx#L46)

Stevearzh:
Sorry, just added them.

alitaheri:
@Stevearzh Add them to defaultProps with `undefined` as value. and change the `&&` syntax with conventional `if (...) { ... }`

Stevearzh:
aye, sir

alitaheri:
More lint errors 😅 😅 Try running `npm run lint` in `lib` folder before committing.

alitaheri:
Thanks, Looks good. @dmtrKovalenko Please review and merge this one. It's an important fix for the typescript users. 😅

PR #146, opened 2017-12-28
MUI version update

cherniavskii:

#


dmtrKovalenko:
Is there any breaking changes?

cherniavskii:
No :)

dmtrKovalenko:
Thus we should not force update to the new version of material-ui

cherniavskii:
Good point, you are right

Issue #161, opened 2018-01-04
Allow override for TextField component used


noah-potter:
We wrap the material ui TextField pretty heavily. Could we add a prop that allows the DateTextField to use a custom TextField instead of the imported TextField?

I think a prop like `TextField` or `TextFieldComponent` that accepted this wrapped component would work pretty well. Looks like extra props already end up on DateTextField

Thoughts before I submit a PR?

dmtrKovalenko:
Hi. no worries - feel free to open a PR - its an a lightweight and nice feature, but actually its really any chance that somewho will use material-ui and material-ui-pickers and will not use material-ui TextField? 😄

bsinibaldi:
@dmtrKovalenko Yes and no; we're using material-ui components, but would, for example, like to be able to trigger the datepicker from a dropdown (e.g. a schedule now vs. later type thing), instead of having an interstitial step with an input. If we could customize the base component, we could change it to be a dropdown or make the TextField itself hidden, etc.

ValentinH:
Any ETA for a new release bringing #163?

PR #163, opened 2018-01-05
Add `TextField` prop to DatePicker, DateTimePicker, TimePicker

noah-potter:
- [ x] I have changed my target branch to **develop** :facepunch:

#161 

## Description
Add `TextField` prop to DatePicker, DateTimePicker, TimePicker
Update prop documentation to reflect new prop
Add tests


noah-potter:
@cherniavskii Thanks for finding that. I'll rename it
@dmtrKovalenko I'll move it to default props, but similar to BackdropComponent, the prop type should be a function or a string

ValentinH:
@dmtrKovalenko when will this be released?

PR #176, opened 2018-01-12
Use absolut import of material-ui

danieljuhl:

#174

## Description
Changed all imports of material-ui to absolute paths


danieljuhl:
I did notice - I'm looking at it right now.

PR #180, opened 2018-01-14
Develop

vclav:

# 

## Description

Add keyboard shortcuts handling to calendar in its most simple form, arrows left/right changes date by 1 day, arrows up/down changes date by 7 days and all changes are propagated by `onChange` callback. Sorry for reformat of code, I'm using prettier and it did the changes automatically and since it does not break anything, I left it as is, although it hardens evaluation of changes.


vclav:
This has problem with automatic view switching when you call `onChange` in `DateTimePicker`, as in `TimePicker`, when you choose hours, you are automatically switched to minutes view, in `DateTime`, when you choose date, you are automatically switched to time.

This could be solved two ways

1. create another state for date, which is selected but not yet confirmed and confirm date on <kbd>Enter<kbd>
2. drop this automatic switch mechanism (IMHO it's not good for UX)

PR #195, opened 2018-01-26
Change keyboardIcon type from string to ReactNode.

anogr:
- [x ] I have changed my target branch to **develop** :facepunch:

# 

## Description
I noticed that TypeScript complained about using an icon for keyboardIcon due to the type being a string. I changed that to a ReactNode. This is my first pull request ever on github. I hope everything is as it should be.


dmtrKovalenko:
Thank you!

PR #281, opened 2018-03-17
 Typescript: Add locale to MuiPickersUtilsProviderProps

tstaehli:

# 

## Description
very small change; simply add the newly introduced locale prop of MuiPickersUtilsProviderProps in the corresponsinf typing file


dmtrKovalenko:
@tstaehli build failed :(

lib/src/utils/MuiPickersUtilsProvider.d.ts
```diff
@@ -5,6 +5,7 @@ import { Utils } from '../typings/utils';
 export interface MuiPickersUtilsProviderProps {
   utils: Utils;
   children: ReactNode;
+  locale: (any | string);
```
dmtrKovalenko: It should not be required

tstaehli:
changes applied. Not sure about the build though; this does not seem like a change that could break the build...

dmtrKovalenko:
@tstaehli we have typescript usage tests. And there is no locale passed — that’s why the build failed

dmtrKovalenko:
@tstaehli btw why do you use any | string - is it really needed.  Any will cover string as well

tstaehli:
@dmtrKovalenko ah, that explains the test failure. good testsuite then :-) I adapted the type from the MuiPickersUtilsProvider.jsx where the prop is defined as `PropTypes.oneOfType([PropTypes.object, PropTypes.string]),`.
I realise that `any` would suffice; however I wanted to emphasize that locale is either a string for use with MomentUtils and a more complex for DateFnsUtils.

tstaehli:
sorry, wrong button

cescoferraro:
@dmtrKovalenko please merge!

cescoferraro:
would you mind making another release?? @dmtrKovalenko thanks!

PR #288, opened 2018-03-21
Fix mask behavior

LastDreamer:


dmtrKovalenko:
Please describe actually what have you changed step-by-step. 
I think its not the best way to fix this issue

LastDreamer:
I changed check that we have valid date in text input to check that we have string that corresponds formatted date in given format.
This changes prevented premature onChange.
Now we have problem not with moment only - but with date-fns also. [Just try type 10/10/2001](https://codesandbox.io/s/535jw62ol4)

LastDreamer:
@dmtrKovalenko `everything will be fixed simply by passing 3 argument true to the moment parse function` - and what about date-fns?

LastDreamer:
Also now in my PR resolved [this bug - fill datepicker and then click clear](https://codesandbox.io/s/535jw62ol4)

dmtrKovalenko:
@LastDreamer I`m sorry but I`m closing this PR due to invalid fix. [Here](https://github.com/dmtrKovalenko/material-ui-pickers/pull/295/commits/6abce340b73a1555075803f340def62d74a79b8a) is my fix and also everything works file with date-fns and 10/10/2018.  

Please can you create separate PR related to clearing keyboard input?

PR #307, opened 2018-03-28
Fix filepaths for date-fns files from camelCase to snake_case

mmjames04:

# 

## Description

This PR changes the file path for files referenced in `date-fns-utils.js`.  Prior to this change, file names with more than one word were camelCased.  This change implements snake_case as the files within `date-fns` are snake_cased.  Prior to this change, I was unable to get `material-ui-pickers` to work in my code.  This change makes the code work as expected.


dmtrKovalenko:
Wait. Just update datefns to 2.0

Issue #310, opened 2018-03-29
How can I change the calendar day format?


abubberman:
How can I change the calendar day format? In my case, i want to change '**woe, sep 19**' to '**wo 19 sep**'.

![image](https://user-images.githubusercontent.com/13328911/38079463-40eb976c-333f-11e8-85f5-5712d5fa8555.png)

dmtrKovalenko:
Check out corresponding [documentation section](https://material-ui-pickers.firebaseapp.com/guides/formats)

abubberman:
Thnx, will have a look!

abubberman:
I'm trying to implement the sample in my code, but Typescript is complaining that 'DateFnsUtils' is not a constructor function type.  I'm probably missing something...?

```javascript
import DateFnsUtils from 'material-ui-pickers/utils/date-fns-utils';

class LocalizedUtils extends DateFnsUtils {
  getDatePickerHeaderText(date: Date) {
    return format(date, 'D MMM YYYY', { locale: this.locale });
  }
}
```

dmtrKovalenko:
But its a constructor type https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/src/utils/date-fns-utils.js

Hmm I will check it when have a chance

abubberman:
What seems to work is moving the d.ts files in a separate folder.

![image](https://user-images.githubusercontent.com/13328911/38082358-b5f71d90-3346-11e8-9ffe-c8faf41f05ce.png)

dmtrKovalenko:
@abubberman you just disable ts definitions.

abubberman:
yes.. you're correct... my conclusion was a bit to fast.

abubberman:
indeed.. typing is disabled now..

dmtrKovalenko:
Fix will appear in the nearest release

abubberman:
Thanks for fixing! When is the nearest release?

dmtrKovalenko:
@abubberman I hope at the beginning of next week

abubberman:
ok, thnx!

PR #312, opened 2018-03-29
[typescript] Fix 'utils' is not a constructor function type

dmtrKovalenko:

## Issue
#310  

## Description
Fix errors on typescript when trying to extend utils, because we have exported interfaces. And also add custom typing for moment and date-fns utils.


CharlieIGG:
@dmtrKovalenko I'm facing this issue, but I see this workaround is not available in the master branch. What do you recommend I do?

PR #317, opened 2018-04-01
Today button implementation

tkachenko-tatiana:
This PR implements new feature - 'Today' button (Issue#225)

I added 3 examples to doc for each picker. Props in doc is generated properly. Also I added unit test for Modal Dialog, where 'Today' button is implemented.


dmtrKovalenko:
Thank you for taking out that feature. :)

Issue #306, opened 2018-03-28
Controlled DatePicker shows InputMask when cleared


FranBran:
1. Manipulate input by keyboard on a controlled DatePicker
2. reset input from "outside"

result: the mask is shown and overlaps the label

Reproducable on Demo-Page:

![animation](https://user-images.githubusercontent.com/1215407/38022989-606a1ffa-3281-11e8-9ae2-454dc8dd1664.gif)

alexander-morozov:
@FranBran try to use function that return mask instead of string
It helped me.

`getMask = (value) => {
    return value ? this.props.mask : []
  }`

dmtrKovalenko:
@alexander-morozov would you like to open a PR?

alexander-morozov:
@dmtrKovalenko It doesn't need any code changes. Just pass function instead of value to mask prop.

https://github.com/text-mask/text-mask/blob/master/componentDocumentation.md#readme
See mask function section

dmtrKovalenko:
We have not fixed that, just add example how to handle such situation in docs

PR #320, opened 2018-04-03
Input mask on clear

dmtrKovalenko:

# 
#306


alexander-morozov:
@dmtrKovalenko maybe it'd be better to check if this.props.mask is a function and just use it and only if it's a value than wrap into your function? Because probably someone can want to return mask for empty value too.

alexander-morozov:
Maybe it shouldn't be fixed at all. Just add case with mask as function to documentation

dmtrKovalenko:
@alexander-morozov Thanks for point,  I have missed a mask for empty value, just add this to the docs.

Issue #333, opened 2018-04-07
Add option to disable opening of dialog when a users clicks enter in keyboard mode inside the textfield


Loktor:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers | 1.0.0-rc.5 |
| material-ui         | 1.0.0-beta.38|
| React               | 16.2.0 |
| Browser             | Chrome |
| Platform            | Windows |

## Steps to reproduce
1. Simply click enter inside a datePicker (e.q. after entering a correct date)

## Behavior

It would be nice if it would be possible to disable this behaviour, i think it should be enough to add the property + a check to https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/src/_shared/DateTextField.jsx handleKeyPress to add this functionality

## Live example

https://codesandbox.io/s/oqxj1kzk46

Loktor:
I added a pull request https://github.com/dmtrKovalenko/material-ui-pickers/pull/336 for this feature request, i think some others beside me could also benefit from it :)

Loktor:
I think this issue can now be closed :)

PR #336, opened 2018-04-07
[#333] Added feature to disable automatic opening of picker when a us…

Loktor:

# 

This pull request implements: https://github.com/dmtrKovalenko/material-ui-pickers/issues/333

## Description

I added an additional feature to allow users to disable the automatic opening of the datePicker when a user clicks enter in the keyboard mode. See: This pull request implements: https://github.com/dmtrKovalenko/material-ui-pickers/issues/333 for more details

### DIFF
--- a/lib/src/_shared/DateTextField.jsx
+++ b/lib/src/_shared/DateTextField.jsx
@@ -48,6 +48,8 @@ export class DateTextField extends PureComponent {
     keyboard: PropTypes.bool,
     /** Icon displayed for open picker button in keyboard mode */
     keyboardIcon: PropTypes.node,
+    /** enables/disable automatic opening of the picker when the user clicks enter */
+    openPickerOnEnter: PropTypes.bool,
     /** Message, appearing when date cannot be parsed */
     invalidDateMessage: PropTypes.string,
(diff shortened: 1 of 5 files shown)


lib/src/_shared/DateTextField.jsx
```diff
@@ -48,6 +48,8 @@ export class DateTextField extends PureComponent {
     keyboard: PropTypes.bool,
     /** Icon displayed for open picker button in keyboard mode */
     keyboardIcon: PropTypes.node,
+    /** enables/disable automatic opening of the picker when the user clicks enter */
+    openPickerOnEnter: PropTypes.bool,
```
dmtrKovalenko: I thought maybe it would be more convenient to name prop smth like `disableOpenOnEnter`.
It would be looks like
```jsx
<DatePicker disableOpenOnEnter />
```
instead of 
```jsx
<DatePicker openPickerOnEnter={false} />
```

Loktor: Yeah your right that is better, i will change it :)

Loktor:
So now the pr should be fine :)

Issue #334, opened 2018-04-07
When you use the datePicker + the option for showTodayButton and click on the today button inside the dialog then the field value is set instantly


Loktor:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers | 1.0.0-rc.5 |
| material-ui         | 1.0.0-beta.38|
| React               | 16.2.0 |
| Browser             | Chrome |
| Platform            | Windows |

## Steps to reproduce
1. Simply open this example https://codesandbox.io/s/ppnyk5qlm7 and klick on the datePicker icon.
2. If you select any date other than today (then the date value inside the textBox behind the picker doesn't change it's value and you can abort the save action.
3. If you select the today button and you click the abort button the date value was still selected

## Expected behavior

When i click the today button inside the datePicker dialog and the option autoOk is set to false, then the today value should only be selected in the picker (and after clicking abort the old value should still be selected)

## Actual behavior

When i click the today button inside the datePicker dialog and the option autoOk is set to false, then the today value is selected instantly without closing the dialog (e.q. if i click abort now i still selected the today value)

## Live example

https://codesandbox.io/s/ppnyk5qlm7

Loktor:
https://github.com/dmtrKovalenko/material-ui-pickers/pull/337 Should be a fix for the issue i described above :)

Loktor:
I think this issue can now be closed :)

PR #337, opened 2018-04-07
[#334] Fixed issue with today button always selection the value inste…

Loktor:
- [x ] I have changed my target branch to **develop** :facepunch:

# 

This fixes the issues described in https://github.com/dmtrKovalenko/material-ui-pickers/issues/334

## Description

For a detailed description of the problem see: https://github.com/dmtrKovalenko/material-ui-pickers/issues/334

### DIFF
--- a/docs/src/Examples/Demo/BasicDateTimePicker.jsx
+++ b/docs/src/Examples/Demo/BasicDateTimePicker.jsx
@@ -44,17 +44,6 @@ export default class BasicDateTimePicker extends PureComponent {
             showTodayButton
           />
         </div>
-
-        <div className="picker">
-          <DateTimePicker
-            value={selectedDate}
-            disablePast
-            onChange={this.handleDateChange}
-            label="With close on selection"
-            showTodayButton
-            autoOk
-          />
-        </div>
       </Fragment>
     );
   }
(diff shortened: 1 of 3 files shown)


docs/src/Examples/Demo/BasicDateTimePicker.jsx
```diff
@@ -44,6 +44,17 @@ export default class BasicDateTimePicker extends PureComponent {
             showTodayButton
           />
         </div>
+
+        <div className="picker">
+          <DateTimePicker
```
dmtrKovalenko: Please remove additional examples, just change the logic

Loktor: I added the additional examples to have examples which use the today picker + autoOk since there were no autoOk examples yet, but i can remove them :)

dmtrKovalenko: I think prop description for `autoOk` is sufficient

Loktor: okay, then i will adapt the pr :)

Loktor:
So now the pr should be fine :)

PR #343, opened 2018-04-09
Fix missing argument for setYear in typings

lookapanda:

## Description
The year argument for `setYear` was missing in the typings.


dmtrKovalenko:
@lookapanda thanks, can you change name to `count`? Thank you for help

lookapanda:
I just noticed that there are more missing typings. I will add them as a new PR when I have time.

PR #345, opened 2018-04-10
Luxon

lookapanda:

## Description
As requested, here is the PR. I had to do some changes to package.json, I hope they are okay. All tests are still passing. Tell me if something needs to be changed.


lookapanda:
```
> cross-env NODE_ICU_DATA=node_modules/full-icu jest && npm run typescript
node: could not initialize ICU (check NODE_ICU_DATA or --icu-data-dir parameters)
```
Hm, that's weird. It runs fine on my machine. I need to set this environment variable for the localization. Luxon is using native Intl

dmtrKovalenko:
@lookapanda Can you please refer build failing and change your target branch to develop

dmtrKovalenko:
@lookapanda

lookapanda:
https://github.com/dmtrKovalenko/material-ui-pickers/pull/345#issuecomment-380079928

I was talking about the failing build. I don't know why the build fails, it runs locally.

dmtrKovalenko:
> material-ui-pickers@1.0.0-rc.5 all-tests /home/travis/build/dmtrKovalenko/material-ui-pickers/lib
> cross-env NODE_ICU_DATA=node_modules/full-icu jest && npm run typescript
node: could not initialize ICU (check NODE_ICU_DATA or --icu-data-dir parameters)

dmtrKovalenko:
@lookapanda why does it need at all?

lookapanda:
Because I wanted to also test that the localization is properly working.

cherniavskii:
I'm looking into this

cherniavskii:
ok, tests are passing now

dmtrKovalenko:
Good job 👍 
Just change your target branch to develop

lookapanda:
`lookapanda wants to merge 22 commits into dmtrKovalenko:develop from lookapanda:luxon`
Didn't I already do that?

dmtrKovalenko:
@lookapanda Oh, sorry, I was from phone and misslead that you backported master to your branch. Thanks for your work!

lib/package.json
```diff
@@ -44,13 +44,15 @@
   },
   "dependencies": {
     "babel-runtime": "^6.26.0",
+    "luxon": "^1.0.0",
```
cherniavskii: hmm, shouldn't `luxon` be a dev dependency? And also be located in `externalDependencies` along with moment and date-fns?

lookapanda: Jeez, yes, I totally forgot about that. I just ran `yarn add luxon` locally which adds that as a dependency automatically (like `npm install --save`. Yes, that should be changed.

cherniavskii: no worries ;)
would you like to submit a PR?

cherniavskii: Also, it would be great to document luxon localization on our docs page

lookapanda: I just noticed that the performance with Luxon is not really that great compared to, for example, the DateFns util. I will investigate this later this week.

lookapanda:
Okay, I investigated a bit about the performance. Any idea how we could improve the performance of `getWeekArray`?
I ran a benchmark:

```
Luxon: getWeekArray x 1,081 ops/sec ±4.26% (89 runs sampled)
DateFns: getWeekArray x 67,535 ops/sec ±1.31% (91 runs sampled)
```
As you can see, DateFns is much, much faster. There is nothing really one can do about that fact as `date-fns` utilizes JS dates directly and Luxon works with wrappers, but can we improve the performance somehow?

It's very much noticeable when you open the date picker, for example. It takes some time until the modal opens.

cherniavskii:
@lookapanda from a fast look, looks like `getWeekArray` makes too much iterations:

https://github.com/dmtrKovalenko/material-ui-pickers/blob/b09fa89a92f00a3f876e1c2ce6006ef3e05a41ad/lib/src/utils/luxon-utils.js#L127-L149

There are two `map`s and `forEach`, which doesn't make a lot of sense for me.

That being said, if poor performance is caused by Luxon itself - we have nothing to do with that.

Issue #432, opened 2018-05-24
autoOk breaks keyboard input


Mario-Eis:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers |rc.9|

## Steps to reproduce
1. set autoOk to picker
2. open picker modal
3. try to navigate with keyboard

## Expected behavior
Keyboard navigation should still be possible.

## Actual behavior
The first day that is navigated to is accepted as new value. imho autoOk should only be usable with mouse input and not affect keyboard behavior.

## Live example
https://codesandbox.io/s/vq60kmlxk0

- open picker modal
- try to navigate with keyboard

Mario-Eis:
Unfortunately this issue is not solved in rc10. The calendar still closes instantly when using the keyboard to choose a date.

PR #457, opened 2018-06-08
Release rc.10

dmtrKovalenko:
We apologize for long break between releases. But we have done a lot inside ⭐️ 

Big thanks 2 contributors, that support this release ❤️ 
Here is what's changed: 

### Breaking changes 
We have redone forwarding refs logic with the new React's api. So no more `pickerRef` prop
```diff
<DatePicker
- pickerRef={node => { this.picker = node }}
+ ref={node => { this.picker = node }}
/>
```

### Features 
* We have move from inheritance to composition and providing our internal hoc for using any pickers sub-component @dmtrKovalenko
* Accept date on text field blur instead of on any successful parse. @LastDreamer 
* Introduce new TimePicker view with seconds. Use `seconds` prop @baig 
* Switched to new lifecycle components #422 
* Accept react components for messages and labels #387 
* Allow to disable keyboard event listener by setting `allowKeyboardControl` to false.

### Fixes
* Make calendar keyboard event listener disabled by default if its used not under modal #412
* Fix that autoOk prop break keyboard input #432 
* Fix not applying disabled effect for keyboard button if textfield disabled #452
* [ts] Fix missing `onKeyDown` typescript property for `DateTextFieldProps` @Ahmad Ilaiwi
* [ts] Fix missing  `todayLabel` and `showTodayButton` for pickers props @tjaskula


dmtrKovalenko:
@cherniavskii will you have any chance to briefly review it? Thank you

lib/src/DatePicker/DatePicker.jsx
```diff
@@ -24,13 +24,15 @@ export class DatePicker extends PureComponent {
...
     shouldDisableDate: PropTypes.func,
+    allowKeyboardControl: PropTypes.bool,
   }
 
   static defaultProps = {
     minDate: '1900-01-01',
     maxDate: '2100-01-01',
     disablePast: false,
     disableFuture: false,
+    allowKeyboardControl: false,
```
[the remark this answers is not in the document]
dmtrKovalenko: Actually its not a breaking change - because its enabled in the DatePickerWrapper. 
The logic is - if you are using default wrapper with modal you have events disabled. If you are using some internal component - you got its disabled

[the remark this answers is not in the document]
dmtrKovalenko: :D I think we should rename them in the next iteration

baig:
I've added support for seconds in `DateTimePicker` as well. I will open another pull request for that.

mcMickJuice:
@dmtrKovalenko I'm noticing a couple bugs with `develop`, both affecting `DatePicker`:

### Cannot select other dates

When the calendar is open, today's date (or whatever value is) is selected, but you cannot select other dates

### Selecting dates calls passed in onChange handler

When selecting dates, the passed in onChange handler is called even before clicking OK in the modal.

I updated https://github.com/dmtrKovalenko/material-ui-pickers/pull/459 with fixes for these changes since these issues are all tightly related.

However, now `autoOk` does not work with `DatePicker`.  I think we need to have a separate method in `BasePicker` for initializing the date of a picker when it's opened, **OR** we augment `getValidDateOrCurrent` in `BasePicker` to handle setting initial date value.

I prefer the latter since for https://github.com/dmtrKovalenko/material-ui-pickers/issues/450 will also likely require messing with `getValidDateOrCurrent` for setting the default date if no value is passed:

```diff
const getValidDateOrCurrent = ({ utils, value, initialFocusedDate }) => {
+  const initialDate = value || initialFocusedDate || utils.date();

-  const date = utils.date(value);
+  const date = utils.date(initialDate);

- return utils.isValid(date) && value !== null ? date : utils.date();
+  return utils.isValid(date) ? date : utils.date();
};
```

dmtrKovalenko:
@mcMickJuice thank you, your PR fixes all of these issues?

mcMickJuice:
Nope, sorry here are the DatePicker related issues and checked are what my PR fixes:

[x] Cannot select other dates while picker is open
[x] Picker is auto assigning minDate if value is null and minDate > today (https://github.com/dmtrKovalenko/material-ui-pickers/issues/445)
[ ] `autoOk` prop is not being respected

dmtrKovalenko:
I think we are ready to release 🚀

lib/src/wrappers/ModalWrapper.jsx
```diff
@@ -66,6 +67,17 @@ export default class ModalWrapper extends PureComponent {
     open: false,
   }
 
+  static getDerivedStateFromProps(nextProps) {
+    // only if accept = true close the dialog
+    if (nextProps.isAccepted) {
```
mcMickJuice: So you're handling opening/closing of Modal declaratively, using `isAccepted` as the signal to close, correct?  As opposed to imperatively using a ref to ModalWrapper?

dmtrKovalenko: @mcMickJuice you are exactly right

mcMickJuice: cool! thanks

PR #463, opened 2018-06-11
Control when you want to focus initial date in Calendar

### LINKED ISSUE #450: Allow for option to pass "initialFocusedDate" to DatePicker
## Environment

| material-ui-pickers |     1.0.0-rc.9   |
| material-ui         |      1.2.0    |
| Browser             |    Chrome - 66.0.3359.181     |
| Platform            |    OSX     |

## Steps to reproduce
1. Set value as null on DatePicker
2. Set minDate to a value greater than today
3. When you open the calendar, the focused date is today

## Expected behavior
I would like the ability to provide an "initial focused date" that would be highlighted when I open the picker and I don't have a selected value.  If a value is provided, this prop is ignored.

## Actual behavior
No ability to specify an initial focused date

## Live example

https://codesandbox.io/s/q7jxvqqqq9 - this code example uses an older version, which is where we first noticed the need for this prop (before we upgraded to rc9).  This feature is somewhat blocked by https://github.com/dmtrKovalenko/material-ui-pickers/issues/445

millnitzluan:

## Description
Created prop to control when you want to focus initial date, the default value of initialFocusedDate is true, so when you want to disable it, you only need to set to false.


millnitzluan:
@dmtrKovalenko, when I was sending initialFocusedDate in DatePicker props I was getting this error:
```
Warning: React does not recognize the `initialFocusedDate` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `initialfocuseddate` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

Do you know how to solve it?

dmtrKovalenko:
Yes. You need to spread this prop from passing to ModalWrapper and DateTextField

dmtrKovalenko:
And also the property should be named something like toFocusDateInitially or shouldFocusDateInitially. Just follow the convention for Boolean props

millnitzluan:
Okay bro, I'll fix this :+1:

PR #465, opened 2018-06-12
Control when you want to focus initial date in Calendar

### LINKED ISSUE #450: Allow for option to pass "initialFocusedDate" to DatePicker

millnitzluan:

## Screenshot

With `shouldFocusDateInitially` false:
with_false

With `shouldFocusDateInitially` true or without `shouldFocusDateInitially`:
without_prop

## Description
Created prop to control when you want to focus initial date, the default value of `shouldFocusDateInitially` is true, so when you want to disable it, you only need to set to false.


mcMickJuice:
This does not address #450. The idea of that issue is to be able to set a default "start date" whether that's in the past or future. Current behavior is if no value is provided, today's date is highlighted.

millnitzluan:
Yes @mcMickJuice. The initial idea is to open or not the datepicker with date highlight, start date is selected by date for example, if you send value as new Date (null) the starting date will be 1969 Wed, Dec 31

mcMickJuice:
@millnitzluan This is what I was thinking - https://github.com/dmtrKovalenko/material-ui-pickers/pull/466

millnitzluan:
Ah sorry, when you said focused I was thinking in the button highlighted as second screenshot.

PR #486, opened 2018-06-28
D will parsed as ordinal number

cubicwork:

## Description
D will parsed as ordinal number in date-fns 2.0.0-alpha.11
E will parsed as date in date-fns 2.0.0-alpha.11


dmtrKovalenko:
Actually we have locked version of date-fns < alpha.7 due to unsupport of some locales. Now they are not fully supported.

dmtrKovalenko:
Closed because of merge #550

PR #488, opened 2018-06-28
Add on month select

oddlyfunctional:

## Description

Hi! I added a callback to inform when the current month is changed (in my application I'm fetching an API to update the availability for that month based on recurrence rules). I tried to write tests but they wouldn't pass because the context is breaking, so I'm unsure if I should cover the other changes with tests or if should I wait for the suite to be fixed in the `develop` branch.


lib/src/DatePicker/Calendar.jsx
```diff
@@ -85,7 +87,10 @@ export class Calendar extends Component {
   };
 
   handleChangeMonth = (newMonth) => {
-    this.setState({ currentMonth: newMonth });
+    this.setState(
+      { currentMonth: newMonth },
+      () => this.props.onSelectMonth(this.state.currentMonth),
```
dmtrKovalenko: If `onSelectMonth` will be undefined there will be an error cannot call undefined as function

oddlyfunctional: That's why I defined a default value :)

lib/src/DatePicker/Calendar.jsx
```diff
@@ -40,6 +41,7 @@ export class Calendar extends Component {
     renderDay: undefined,
     allowKeyboardControl: false,
     shouldDisableDate: () => false,
+    onSelectMonth: () => {},
```
oddlyfunctional: Here @dmtrKovalenko

lib/src/DatePicker/DatePickerWrapper.jsx
```diff
@@ -114,6 +116,7 @@ DatePickerWrapper.propTypes = {
   /** Enables keyboard listener for moving between days in calendar */
   allowKeyboardControl: PropTypes.bool,
   forwardedRef: PropTypes.func,
+  onSelectMonth: PropTypes.func,
 };
```
dmtrKovalenko: Please add prop description here so it will appear in the prop types table.

dmtrKovalenko:
Closed due to inactivity

PR #492, opened 2018-06-28
Allow providing onBlur handler to DatePicker

### LINKED ISSUE #490: Providing custom onBlur handler breaks typing
## Environment

| material-ui-pickers |      1.0.0-rc.10   |
| material-ui         |     1.2.1    |
| React               |      16.4.0   |
| Browser             |   Chrome      |
| Platform            |    Windows     |

## Steps to reproduce
1. Work around #475 for a datepicker with keyboard input
2. Add an onBlur handler (required for e.g. #430 )
3. Type a date in and click away

## Expected behavior
Change handler should be called when the input is blurred

## Actual behavior
Change handler is not called

## Investigation
DateTextField ignores its own handleBlur in favour of the supplied onBlur

alexanthony:

## Description
Spreading other props was overriding the blur handler if an onBlur was passed into DatePicker


dmtrKovalenko:
1. Take a look in the Travis build. 
2. Make sure that typescript typings will support the new propert

lib/src/_shared/DateTextField.d.ts
```diff
@@ -13,6 +13,7 @@ export interface DateTextFieldProps extends Omit<TextFieldProps, 'onChange' | 'v
     maxDate?: DateType;
     maxDateMessage?: React.ReactNode;
     mask?: any;
+    onBlur?: () => void;
```
dmtrKovalenko: actually the typing is (e: SyntheticEvent) => void. 
But maybe there already typing for onBlur in TextField props. Can you test (there is specific typescript usage tests in the project)?

alexanthony:
@dmtrKovalenko it would be awesome for us to get a release with this change in the next couple of days - is this likely or shall i work on an internal version?
Also not sure if you're aware, but I don't think that this package can be installed directly from git because of the folder structure.

dmtrKovalenko:
@alexanthony OK, I will work hard to release new stable version asap

PR #512, opened 2018-07-11
Remove unnecessary spacing in MuiPickersUtilsProvider

olee:
Right now the provider generates useless space characters in the DOM.


dmtrKovalenko:
Please fix the build

olee:
Ok I'm not sure why anyone would ever add this kind of weird rule to not pass react children as props (as especially in case of  wrapping comonents etc. this this a way cleaner syntax), but I can adjust it to use the tags again I guess... (Though I'd rather suggest to disable that checking rule if possible... Oo)

PR #516, opened 2018-07-17
Build umd bundle

TrySound:
Ref https://github.com/dmtrKovalenko/material-ui-pickers/issues/503

In this diff I added umd bundle. Dependencies was slightly shaked
to simplify distribution and consuming.

`classnames` is too small project to be peer dependency. As a regular
one it will introduce less noise.

material-ui is published as stable and there is not need to keep it as
peer. Moreover it's not possible to keep it as external package in umd
because of path fragmentation.

Removed codecov and enzyme from direct dependencies.

This package will be available in `window` object under `material-ui-pickers`
namespace similar to `material-ui` one.

`keycode` is added explicitly as a direct dependency.

### DIFF
--- a/lib/.size-snapshot.json
+++ b/lib/.size-snapshot.json
@@ -1,20 +1,25 @@
 {
+  "build/dist/material-ui-pickers.umd.js": {
+    "bundled": 773228,
+    "minified": 315245,
+    "gzipped": 79507
+  },
   "build/dist/material-ui-pickers.cjs.js": {
-    "bundled": 118716,
-    "minified": 66127,
-    "gzipped": 13574
+    "bundled": 119387,
+    "minified": 66434,
+    "gzipped": 13624
   },
   "build/dist/material-ui-pickers.esm.js": {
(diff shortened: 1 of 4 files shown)


TrySound:
Hm... seems like material-ui dependency case doesn't fit well with size-limit. What do you think, can size snapshot be enough?

dmtrKovalenko:
Closing this PR due to incomprehensible changes in rollup.config

dmtrKovalenko:
I will add umd bundle by myself

TrySound:
What? Common man, what's wrong here?

TrySound:
I did the job. And I do this job for a lot of projects. Your answer looks like I'm doing shit all the day.

dmtrKovalenko:
@TrySound ohh, sorry if I offended you

dmtrKovalenko:
I have considered the way for excluding material-ui from umd bundle. And also reuse more of code. So now I see that its impossible because of relative imports.
You are good dev - I will mention you as change creator in release notes

PR #537, opened 2018-07-27
Release rc.12

dmtrKovalenko:
Thanks so match to 7 contributors, that makes this release possible! You are the best ❤️ 
And we are mostly near the **1.0.0** release 🚀 

### Breaking changes
N/A

### Features
* New awesome set of material design animations for all pickers (check the video [here](https://www.youtube.com/watch?v=4tEDNjH1MD8))
* New `onInputChange` prop, which fires on each input change
* New `initialFocusedDate` prop to make ability display initial date @mcMickJuice 
* New `container` prop to add ability change the modal container @aleliberty 
* New `mergeDateAndTime` method in utils that adds ability to more slightly control time selection @Philipp91 
* New `ViewContainerComponent` prop for DateTimePicker @Philipp91 
* Move all text field formats to utils instances, that will make ability to override text field formats globally
* UMD bundle coming `material-ui-pickers/dist/material-ui-pickers.umd.min.js` @TrySound 

### Enhancmenents
* Update to babel 7 and improve bundle size @TrySound 
* Reexport props interfaces from index modules [typescript]
* Add new moment localization example in docs @josephMG 
 
### Fixes
* Fix freezings on FireFox
* Fix incorrect format keys for default luxon utils.


dmtrKovalenko:
@alitaheri there several fields have been changed in utils interface - `material-ui-pickers-jalali-utils` should be updated as well

alitaheri:
@dmtrKovalenko Right, thanks for the heads up 🙏

TrySound:
I would add umd bundles to features

dmtrKovalenko:
@TrySound done

PR #542, opened 2018-07-29
Migrate to transition group v2

### LINKED ISSUE #541: react-transition-group issues
I'm not quite sure if this is a material-ui-pickers issue, but when upgrading to the recent rc.12 I faced some issues due to `Can't resolve 'react-transition-group/CSSTransitionGroup' in '(...)/node_modules/react-addons-css-transition-group'`. I solved it by manually installing an older version of react-transition-group(1.x). I don't know much about transitions, but I guess the solutions are:
1. use `react-transition-group` directly as `react-addons-css-transition-group` tries to access an outdated version of `react-transition-group` or
2. add `react-transition-group`@1.x to dev dependencies.

## Environment

| material-ui-pickers |  latest      |
| material-ui         |    latest     |
| React               |    latest     |

sakulstra:

### Closes # 
- closes #541 
## Description
I think this should fix the issue reported in #541 .
- not sure where to add the react-transition-group dep ... peer or dev (it will always already be available trough material-ui)?


sakulstra:
I won't be able to continue with this before the next weekend.
Feel free to pick it up (especially because I have 0 experience with react transitions).

Also I just noticed there's a @material-ui/core/Slide transition we might want to reuse instead?

sakulstra:
just a quick status update here. It's working again and using react transition group v2

------

I noticed there's a slight bug i'm not sure how to solve... when navigating in one direction and than switching the direction the content will slide out the wrong direction. I guess it's related to how i used the keys, so the onExit classes are initialized in one direction, but won't change as the key wont be accessed in the onExit. I'd be happy if someone would pick that up

dmtrKovalenko:
@sakulstra Many thanks for your work. I was pretty busy so have no time to pick it up? Can this PR be merged or there is still some issues?

sakulstra:
@dmtrKovalenko it's the first time I used transitions in react, so I guess it would be best im someone carefully reviewed this pr, but yeah for me it seems to be ready and could be merged.

dmtrKovalenko:
LGTM;

sakulstra:
https://github.com/dmtrKovalenko/material-ui-pickers/issues/573#issuecomment-413477544

Issue #534, opened 2018-07-25
Hard-coded default formats are specific to locale and library


Philipp91:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers |    v1.0.0-rc.11     |
| date-fns | v2.0.0-alpha.12 |

## Steps to reproduce
1. Use the libraries above (especially the newer date-fns version).
2. Switch your locale to `de` (this is not necessary to reproduce the more severe part of this bug).
3. A `DateTimePicker` with the current date (2018-07-25T19:58:35).

## Expected behavior
The output should be "25. Juli, 21:58" because my locale is `de`.

## Actual behavior
The actual output is: "July 206. 21:58"
Firstly, the locale is wrong (note that "Juli 25. 21:58", the part-by-part translation, would also be wrong).
Secondly, the number 206 is the day of year, and not the day of month.

## What goes wrong

Everything works fine with `date-fns v2.0.0-alpha.7`. Since then, they have changed the meaning of the symbols in their format strings. `material-ui-pickers` cannot (easily) keep up with this change because the date formats are hard-coded throughout the code base (e.g. [here](https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/src/DateTimePicker/DateTimePickerWrapper.jsx#L56)). Changing it there would break compatibility with the other two supported date libraries.
The same hard-coding also makes it difficult to globally adjust the format (without passing as a prop to every single picker instance), which is necessary to localize these formats.

## Solution idea

Ideally, the localization would work out of the box, though that is a much more difficult goal to achieve. So as a first step, it should be possible to set the default format globally, e.g. through the "utils" which users can [customize](https://github.com/dmtrKovalenko/material-ui-pickers/blob/master/lib/__tests__/DatePicker/PatchedUtils.usage.tsx#L15-L19) or the `MuiPickersUtilsProvider`.

Since the formats are not only locale-specific, but also library-specific, I suggest integrating them in the utils. The shipped version of each utils class could:

- have properties like `DateTimePicker12hFormat`, which can be overridden in sub-classes,
- if possible, retrieve a sensible locale-specific default from the underlying library (e.g. from [here](https://github.com/date-fns/date-fns/blob/master/src/locale/de/_lib/formatLong/index.js#L3-L8)),
- optionally contain locale-specific formats in the `material-ui-pickers` file, at least for the most important locales,
- fall back to the English pattern, as users can always override.

[...]

PR #550, opened 2018-08-01
Update to date-fns 2.0.0-alpha.16 and adapt format strings

Philipp91:

## Description
`date-fns` changed its date formats to match the Unicode Technical Standard (see also https://github.com/date-fns/date-fns/issues/810). Now that #534 is fixed, we can update to the latest version and adjust the format strings globally without affecting compatibility with other date libraries.

Not sure if you want all the changes in `package-lock.json`, that's just what happens when I run `npm install` on my dubiously configured machine.


dmtrKovalenko:
Yes I think we should update. So can you also change the docs and Readme.  Where specific date fan version mentioned?

Philipp91:
Done. The coverage report is a false positive, I don't think I'm changing the coverage in any way.

Philipp91:
Friendly ping

TheRusskiy:
I am beginning to think it may be worth having this as a separate package (as well as the original `date-fns-utils`) and then just reference it from the docs. Many people still use the older `date-fns` and it's gonna break material-ui-pickers for them. With so many radical changes in `date-fns` recently I wouldn't be surprised if they changed their API again.

Philipp91:
That's definitely a reasonable argument and ultimately your choice.

If `date-fns` were stable already, it would definitely be the way to go, because a `material-ui-pickers` update shouldn't imply a `date-fns` update that forces users to do major changes to their code, assuming that they use `date-fns` for other purposes than just `material-ui-pickers`. So there might be a good reason for them to stay at the old version.
However, since `date-fns` isn't stable yet, it's okay if they change their API and users can't really complain if they already depend on it a lot. If they want the stable version ultimately, they will have to update anyway. And most users will respect the status of `date-fns` and not build a lot on it yet, so updating to the latest alpha should be doable for them.

Still, putting the utils in a separate library each (possibly with exactly the same version numbers as the respective date library) can make sense. Would that make `material-ui-pickers` (slightly) more difficult to use, i.e. would users have to add multiple dependencies? Or is there a way this can work out of the box with a single dependency?

PR #555, opened 2018-08-03
Use index as key for weekdays

jarib:
The luxon utils return single-character weekdays, which gives us duplicate keys
T and S.


jarib:
IMHO using the index is perfectly fine in this case, since the order or number of weekdays never changes. 

It is certainly better than the current state, where elements may be duplicated or omitted:

> Warning: Encountered two children with the same key, `S`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

PR #562, opened 2018-08-07
fix(deps): update recompose and event-listener to resolve issues with babel.56

sakulstra:

## Description
Currently for people with @babel/runtime >=beta.56 the build breaks due to issues in recompose. Updating the deps fixes the issue.

It would be awesome if you could cut a new release with the fix included.

### DIFF
--- a/lib/.size-snapshot.json
+++ b/lib/.size-snapshot.json
@@ -5,22 +5,22 @@
     "gzipped": 14193
   },
   "build/dist/material-ui-pickers.esm.js": {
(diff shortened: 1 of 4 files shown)


sakulstra:
@dmtrKovalenko this issue is a pretty bad blocker for us as we can't adjust the babel version in our setup. is there any chance we can get another rc release soon?

lmachens:
@dmtrKovalenko same issue here. Would be great to have a new release with this change.

PR #585, opened 2018-08-19
Call mergeDateAndTime() in YearSelection

Philipp91:
With the default implementations of `mergeDateAndTime()`, this does not make any difference. But when users customize how time-less dates are represented and override the function, this call is necessary to get a valid value.

### DIFF
--- a/lib/src/DatePicker/YearSelection.jsx
+++ b/lib/src/DatePicker/YearSelection.jsx
@@ -31,7 +31,7 @@ export class YearSelection extends PureComponent {
     const { date, onChange, utils } = this.props;
 
     const newDate = utils.setYear(date, year);
-    onChange(newDate);
+    onChange(utils.mergeDateAndTime(newDate, date));
   }
 
   getSelectedYearRef = (ref) => {


dmtrKovalenko:
May you please add a comment why `mergeDateAndTime` called here. That`s can be not obvios

Philipp91:
Sorry, I'm changing my mind about this pull request -- and closing it.

Your comment that it's not obvious why this is necessary prompted me to rethink. It's desirable not to have hard-to-explain behavior checked into the library, especially as it would go untested. My use case (and similar ones) would break again in the future if someone modifies dates in a similar way (like `setYear()`) and doesn't call `mergeDateAndTime()`, which will work fine with the default date library setup, so there's no guarantee that any bug will be caught.

What's common to all these places where the conversion needs to happen is that they feed into `onChange()`. So I'm redacting this pull request and recommending folks with similar use cases to tap into `onChange()` instead and do the necessary conversions there. This should, by construction, cover all existing and future places where the main library needs to modify a date.

PR #624, opened 2018-09-14
Make classnames a direct dependency

TrySound:
I don't use classnames and it's not the best way to require installing
it from user. Peer depenencies should be used carefully. react,
material-ui are good packages for this purpose. Keeping classnames
in peer depenencies doesn't solve any problem. It's not so big package
to be a dependency bloat. And if one day there will be a different
version then I'm sure it will be a breaking change and users app will be
broken.

### DIFF
--- a/lib/package-lock.json
+++ b/lib/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "material-ui-pickers",
-  "version": "1.0.0-rc.13",
+  "version": "1.0.0-rc.14",
   "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
But as far I know material ui use it as direct dependency - so we are just requiring it

TrySound:
Not sure about npm, but yarn warns user to install it. Relying on the fact that another dependency uses this package is not the safest way. material-ui may remove it at any time and this package will be affected.

PR #628, opened 2018-09-14
Remove propTypes in production

TrySound:
This saves 8kb in production bundle.

### DIFF
--- a/lib/.babelrc
+++ b/lib/.babelrc
@@ -4,7 +4,8 @@
     "@babel/react"
   ],
   "plugins": [
-    ["@babel/proposal-class-properties", { "loose": true }]
+    ["@babel/proposal-class-properties", { "loose": true }],
+    ["transform-react-remove-prop-types", { mode: "unsafe-wrap" }]
   ],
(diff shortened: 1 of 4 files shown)


dmtrKovalenko:
But! We build the project in production mode. Somebody imports it through npm. And there propTypes is missing.

TrySound:
Why anybody need to access propTypes directly? They should be used only by react internally.

TrySound:
propTypes will validate in development mode

dmtrKovalenko:
Does this package removes actually node module of prop-types. 
Or it removes actually Component.propTypes={} definitions?

TrySound:
Not removes, wraps with NODE_ENV condition which is removed in production mode.

TrySound:
Take a look here
https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types#mode

TrySound:
The point is that prop types doesn't make sense in production code. It's a tool for developer validation.

TrySound:
So what is your thought?

PR #632, opened 2018-09-15
Remove recompose

TrySound:
This change saves 3kb in user production bundle.


TrySound:
You probably need to setup prettier.

dmtrKovalenko:
Idea is great. I filled that it must be done without recompose :)
Thats strange. Codecov says that your changes covered for 66% but the whole coverage increased. 

I will review that changes asap

TrySound:
@dmtrKovalenko Could you publish another rc?

PR #649, opened 2018-09-24
Export inline pickers from entry point

TrySound:


dmtrKovalenko:
I didnt expect to export InlinePickers from entry point in prospective of bundle increase for users that importing cjs from index.js.

TrySound:
image

TrySound:
You are already export it deeper. Users will rely on treeshakability in this case.

TrySound:
However it doesn't work well yet. Mostly because of material-ui and defaultProps. The latest will be fixed soon.
image

TrySound:
And still it doesn't matter. Every feature should be exported. The size will be increased with different ways.

Issue #661, opened 2018-09-30
Cannot read property 'date' of undefined


mcdougal:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers | 1.0.0-rc.15 |
| @material-ui/core  | 3.1.1 |
| React               |  16.5.2  |
| Browser             | Chrome        |
| Peer library        | moment@2.22.2  |

## Steps to reproduce
1. Use `MomentUtils`
2. Try to render `DatePicker`

## Expected behavior
It should render

## Actual behavior
It throws an error:
```
TypeError: Cannot read property 'date' of undefined
    at getInitialDate (https://6nwylqj9kr.codesandbox.io/node_modules/material-ui-pickers/dist/material-ui-pickers.cjs.js:1914:20)
    at new BasePicker (https://6nwylqj9kr.codesandbox.io/node_modules/material-ui-pickers/dist/material-ui-pickers.cjs.js:1932:13)
    at constructClassInstance (https://6nwylqj9kr.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:11769:18)
    at updateClassComponent (https://6nwylqj9kr.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:13491:7)
```

It seems like the `utils` context is not getting passed into the BasePicker for some reason? https://github.com/dmtrKovalenko/material-ui-pickers/blob/develop/lib/src/_shared/BasePicker.jsx#L7. This was working for me on 1.0.0-rc.14 and stopped working when I upgraded to 1.0.0-rc.15 without making any other changes, so I think it was related to the update.

I've checked the changelog and it doesn't look like there should be any breaking changes besides the new `@material-ui/core` dependency, but it's totally possible I've missed something.

## Live example
Here are two examples with the exact same code except for the version:
Working example with 1.0.0-rc.14: https://codesandbox.io/s/vj6n088935
Broken example with 1.0.0-rc.15: https://codesandbox.io/s/23o9339opy

dmtrKovalenko:
I am sorry. I will publish fix as soon as possible
Now as workaround change import from 
`import { DatePicker } from 'material-ui-pickers'` 
to 
`import DatePicker from "material-ui-pickers/DatePicker";`

mcdougal:
No problem, thanks for awesome library! The workaround fixed it so I'll use that for now

dmtrKovalenko:
[...]

PR #662, opened 2018-09-30
Remove esm build

dmtrKovalenko:
#661 Issue was opened because in esm module for some reason react context didnt work. I have removed that build to fix the issue but now needs to figure out what the problem. 
@TrySound have you any thoughts?

### DIFF
--- a/docs/config/paths.js
+++ b/docs/config/paths.js
@@ -17,9 +17,9 @@ function ensureSlash(path, needsSlash) {
     return path.substr(path, path.length - 1);
   } else if (!hasSlash && needsSlash) {
     return `${path}/`;
-  } else {
+  } 
     return path;
-  }
+  
 }
 
 const getPublicUrl = appPackageJson =>
@@ -47,7 +47,7 @@ module.exports = {
   appIndexJs: resolveApp('src/index.js'),
   appPackageJson: resolveApp('package.json'),
   appSrc: resolveApp('src'),
-  libSrc: resolveApp('../lib/src'),
+  libSrc: resolveApp('../lib/build'),
   yarnLockFile: resolveApp('yarn.lock'),
   testsSetup: resolveApp('src/setupTests.js'),
   appNodeModules: resolveApp('node_modules'),
--- a/lib/.babelrc
+++ b/lib/.babelrc
@@ -5,7 +5,7 @@
   ],
   "plugins": [
     ["@babel/proposal-class-properties", { "loose": true }],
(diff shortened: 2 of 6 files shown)


TrySound:
The problem could be because of using a few instances of context. So since we use bundles as entry points users should use only one way to import things.

```
import { DatePicker } from 'material-ui-pickers';
```
vs
```
import DatePicker from 'material-ui-pickers/DatePicker';
```

This can be also solved with removing path imports. But if you want to keep it try to add helpful message in context consumer which will just assert context value.

Removing esm is a bad solution. My bundle with esm is decreased with 12kB (not just 3kB)

TrySound:
@dmtrKovalenko So can you publish esm back?

rosskevin:
@dmtrKovalenko @TrySound I am seeing this with my local build of #695.  I have a theory:
- cjs is deployed in the package alongside the d.ts files (appears necessary for resolution of ts).  There is no other reason that cjs is exploded other than this.
- My downstream usage is a direct import `import MuiPickersUtilsProvider from '@alienfast/material-ui-pickers/utils/MuiPickersUtilsProvider'` as opposed to index
- I think the webpack/source files are resolving based on the esm.  e.g. I have errors thrown from the esm file in the stack.

@TrySound can you take a look at the rollup config that @dmtrKovalenko put together in #695, I think esm should be rolling up es2015, but I think it currently is rolling up cjs?  Also thoughts on resolving d.ts with rollup files instead of having the exploded d.ts and cjs files together?

TrySound:
I already said a few times. Import provider from index. It's part of the bundle.
```
import { MuiPickersUtilsProvider } from 'material-ui-pickers'
```

rosskevin:
That's fine, but we need to prevent a direct import.  As-is, with d.ts files they resolve.  I'll try killing exploded cjs, leaving .d.ts as-is and going to index.

rosskevin:
As long as the .d.ts files are exploded, it type checks, but will not run.  So we either need to have the .d.ts rolled up too, or stop the esm bundle and make it exploded and just rollup the cjs.

TrySound:
How esm and d.ts are related to each other? You import from just `material-ui-pickers`. Typescript will resolve `main` and found d.ts files. Bundler will look at `module` field. Everybody take what they need.

TrySound:
Ah, I see. We probably need to point `main` field to `lib/index.js` like formik does https://unpkg.com/formik@1.3.1/dist/

So node and typescript will take index.js and bundler will take esm bundle.

PR #674, opened 2018-10-07
Rewrote codebase in typescript 

dmtrKovalenko:
This PR is definitely large but makes support of typescript painless 👍


itskibo:
@dmtrKovalenko do you have intentions of implementing issue #676 in this?

I think most, if not all, of the changes are transforming the interfaces to types, and it currently prevents us from upgrading `@material-ui/core` to 3.2.0.

dmtrKovalenko:
@TrySound Hi 🥇 
Can you please take a brief look on changed rollup config in this PR? 
I do not want to break smth 🤕

docs/src/Examples/Demo/DatePicker/BasicDatePicker.jsx
```diff
@@ -1,5 +1,5 @@
 import React, { Fragment, PureComponent } from 'react';
-import { DatePicker } from 'material-ui-pickers';
+import DatePicker from 'material-ui-pickers/DatePicker';
```
TrySound: Can we do not show this usage in examples? esm bundle is relatively small and will be even smaller. Path imports can be used by advanced users.

Otherwise we will have the same confusion like in the latest issue.

dmtrKovalenko: Nice catch. I will redo it back :)

rosskevin:
Awesome @dmtrKovalenko, nice to see you going all-in on typescript!  When are you looking at an rc or `next` release of this?  

Based on comments #676 should be solved here which is preventing material-ui/core@3.2.x upgrades.  I'm trying to workaround using an augmentation externally for the moment but having no luck.

PR #690, opened 2018-10-12
migrate typography to material-ui v3.2.0

vanyakosmos:
This PR closes #678

Issue #678, opened 2018-10-08
Upgrade to @material-ui-core 3.2.0


AndreiEnache:
Upgrade dependency @material-ui-core to 3.2.0 and change typography variants to new ones according to the migration guide [here](https://material-ui.com/style/typography/#migration-to-typography-v2)

Issue #676, opened 2018-10-08
DateTextField definition causes typescript compilation to fail


janis91:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers |        rc17 (also tested with rc14 -> same result) |
| material-ui         |        3.2.0 |
| typescript | 3.1.1 |
| React               |         |
| Browser             |         |
| Peer library        |         | 

## Steps to reproduce
1. Install 3.2.0 of material-ui/core package.
2. Install material-ui-pickers.
3. Build with typescript (npm start / npm run build)

## Expected behavior

Build is successful, like with previous versions of material-ui/core

## Actual behavior

```
Failed to compile.

/***/node_modules/material-ui-pickers/_shared/DateTextField.d.ts
(7,45): An interface may only extend a class or another interface.
```

For me it seems to be an issue with extending the ``Omit`` type. But this is not allowed in TypeScript and therefore fails to compile. Moreover I wondered why this ``Omit`` is used at all, because in the comment it is declared as an ``@internal`` type definition. I think it should not be used in an external library then.

janhoeck:
Same Problem

evilass:
Same Here

tjaskula:
When do you expect a release for that ?

janhoeck:
How I temporary fixed it:
1. Run Typescript with "tsc --skipLibCheck"
2. (Optional) Add // @ts-ignore above a property, if it is not defined

veranoo:
I have same problem. After update "@material-ui/core": "3.2.2"

rosskevin:
This will be fixed in #695

bheklilr:
What version of material-ui-pickers was this bug introduced in?  I'd rather just revert back to an older, working version for now.

rosskevin:
@bheklilr this is a breaking/incompatible type change in `@material-ui/core`, not this, so you need to downgrade that to `< 3.2`.

PR #695, opened 2018-10-15
Update dependencies, convert tests to typescript, fix ts errors

rosskevin:
## Description
- Update all (except locked) dependencies 
- Incorporates and fixes ts errors from #690 (typography updates)
- Convert all tests to ts/tsx
- Fix all uncovered ts errors

## Remaining
- Some tests failing - @dmtrKovalenko will finish

Closes: #676
Closes: #678
Closes: #690


rosskevin:
@dmtrKovalenko having `__tests__ outside of `rootDir` causes a lot of unnecessary pain when trying to use typescript with IDEs.  I'm going to move `__tests__` under `src`, which is both more standard and more useful for IDEs because we can get a single set of errors for the entire codebase - both test and src.

If this is a problem for some reason, it can of course be reversed but it is causing pain right now in trying to get everything over to ts.

lib/src/_shared/DateTextField.tsx
```diff
@@ -77,7 +77,7 @@ const getError = (
 
 export interface DateTextFieldProps
   extends WithUtilsProps,
-    Omit<TextFieldProps, 'onError' | 'onChange' | 'value'> {
+    Omit<StandardTextFieldProps, 'onError' | 'onChange' | 'value'> {
```
eps1lon: I don't know how this component is used but remember that this way the variants `outlined` and `filled` won't be accepted by typescript.

You need to dissect the union into it's base types, omit and then join them again by their variant which is pretty complicated. I think the easiest solution is to join with `{ value: never }` instead of using `Omit` but this will have worse error messages if somebody actually passes the `value` prop.

rosskevin: @eps1lon I'm trying to get ts working first with all the code, then we can move forward with flexibility for other variants as an enhancement.  @dmtrKovalenko made a first pass but there is much I am uncovering as I turn tests to ts/tsx.  Trying to minimize changes in this PR but it is big already.

dmtrKovalenko:
Thanks @rosskevin for the awesome job!
I will take a look on e2e setup   

And one more time — thanks for helping in typescript integration 👍

rosskevin:
I found and fixed several downstream type errors related to usage as a library (such as https://github.com/date-fns/date-fns/issues/932).  Unfortunately these are not or cannot be uncovered locally, but I have found them when using a sample build. 

Thanks for taking care of e2e, I don't really know the setup here.

rosskevin:
The same error in e2e tests is present in runtime usage, not sure if related. e2e looks like it is breaking based on the `jest.doMock('../_shared/WithUtils'` having no effect.  Not sure though.

rosskevin:
@dmtrKovalenko I've done what I can to close the remaining gaps.  The rest of the test errors are things that I think you would better know.  Note that I added a console.error watcher so that any console errors fail.  I fixed most but anything that remains is something that I'm not comfortable changing as I don't know the code that well.  

The sooner this gets released the better as this is causing some pain for mui 3.2.x users.  Thanks for the help and finishing this up!

lib/package.json
```diff
@@ -43,15 +43,11 @@
     "luxon": "^1.0.0"
   },
   "dependencies": {
-    "@babel/preset-typescript": "^7.1.0",
-    "@babel/runtime": "^7.0.0",
```
TrySound: Bring babel runtime back. It's direct dependency.

rosskevin: runtime dependency?  will try

lib/rollup.config.js
```diff
@@ -2,11 +2,15 @@ import path from 'path';
 import resolve from 'rollup-plugin-node-resolve';
 import commonjs from 'rollup-plugin-commonjs';
 import replace from 'rollup-plugin-replace';
-import typescript from 'rollup-plugin-typescript';
+import typescriptPlugin from 'rollup-plugin-typescript';
+import typescript from 'typescript';
 import { uglify } from 'rollup-plugin-uglify';
```
TrySound: Guys, can we go with babel instead? Or do you miss some not implemented yet features?

rosskevin: I don't have any preference.  I don't think there is an issue with babel, I was just trying to get the ts tests done and am bumping into problems with a final build.

rosskevin: is there a reference project good to look at for ts + babel + esm rollup?

TrySound: It's babel + rollup usually. Typescript is only for type checking and definitions generating.

rosskevin: Is there a good reference project for rollup that outputs esm and cjs with babel?

rosskevin: found one https://github.com/rollup/rollup-starter-lib

rosskevin:
I'm stuck on the build - will have to look again in the morning.  I tried what Formik does (see https://github.com/dmtrKovalenko/material-ui-pickers/pull/662#issuecomment-430422316) but direct file imports typechecked in IDE but has ts-loader errors downstream in my storybook.   Will continue tomorrow on the build issue.  Meanwhile, if @dmtrKovalenko can fix the remaining tests we may be able to merge this tomorrow.

rosskevin:
## Rollup/.d.ts findings

I have tried mirroring Formik's build, but direct file imports don't work at runtime here.  

Succinctly:
- we produce `build/material-ui-pickers.esm.js`
- the esm only `export`s everything from index - it is unaware of file paths
- `tsc -emitDeclarationOnly` emits `.d.ts` files (exploded) to build
- Direct file imports succeed with `tsc`
- Direct file imports will fail at runtime (downstream in a project bundling this)

```js
// tsc ok, runtime failure
import MuiPickersUtilsProvider from 'material-ui-pickers/utils/MuiPickersUtilsProvider'

// tsc ok, runtime success
import { MuiPickersUtilsProvider } from 'material-ui-pickers'
```

It is clear that the rollup esm file doesn't support file paths (which is reasonable btw) by this line:

```js
export { DatePickerModal$1 as DatePicker, DatePickerInline$1 as InlineDatePicker, TimePickerModal$1 as TimePicker, TimePickerInline$1 as InlineTimePicker, DateTimePickerModal$1 as DateTimePicker, DateTimePickerInline$1 as InlineDateTimePicker, MuiPickersUtilsProvider };
```

So with that, to be fully correct, we would actually need to rollup/bundle the `.d.ts` files into a single `index.d.ts` so there is no impedance mismatch.  e.g. IDEs/typescript statically verify direct file import paths with exploded .d.ts files, yet the esm doesn't allow it.

This seems like something that rollup might want to consider as an enhancement, otherwise it is clear that rollup produced bundles with ts declarations produced separately are going to continue to have this problem.

## Why this matters

Users checking code with `tsc` will have false positives when they use direct file imports.  Analysis tools like tsc and flow should at a minimum prove that imports are good.

## Potential solutions

1. Rollup is enhanced to bundle all `.d.ts` files into a single `build/material-ui-pickers.esm.d.ts` to mirror the bundle produced for javascript (would be an enhancement)
2. Stop using rollup for `esm` specifically, produce it with `tsc` or `babel` directly and distribute exploded files so there is no impedance mismatch between the `module/esm` build and the `.d.ts` definitions.

## Where to go from here

Thoughts @TrySound @dmtrKovalenko

TrySound:
@rosskevin I still don't get. Why direct imports fails for you? This should work fine.
```js
import MuiPickersUtilsProvider from 'material-ui-pickers/utils/MuiPickersUtilsProvider';
import { DatePicker } from 'material-ui-pickers/DatePicker';

```

TrySound:
When importing from entry point typescript should resolve index.d.ts and all it's dependencies. So not difference in import ways.

TrySound:
It would be good if you will split all build configuration and refactorings into separated PRs. It's really hard to review so much changes.

rosskevin:
@TrySound here is a gist https://gist.github.com/rosskevin/79d726d2b11a419db6649ab63bf79652

- `module: 'material-ui-pickers.esm'`
- This means the webpack bundler knows nothing but this file
- assume only this file and the package.json exist in the npm package

Where in this file could a direct file import be resolved?  I only see a single export, and there is no semantic for further file resolution.  Please point to the line in the esm file that would allow `import MuiPickersUtilsProvider from 'material-ui-pickers/utils/MuiPickersUtilsProvider'` to work, because I am confused.

TrySound:
@rosskevin How webpack and typescript are related to each other? Webpack resolves `module` field. Typescript resolves `main` field. `main` field should refer to `lib/index.js` which has `lib/index.d.ts` near itself. What is the problem?

TrySound:
In case if you import from `material-ui-pickers/utils/MuiPickersUtilsProvider` you will consume cjs. It's here https://unpkg.com/material-ui-pickers@1.0.0-rc.17/utils/MuiPickersUtilsProvider.js

rosskevin:
> In case if you import from material-ui-pickers/utils/MuiPickersUtilsProvider you will consume cjs. It's here https://unpkg.com/material-ui-pickers@1.0.0-rc.17/utils/MuiPickersUtilsProvider.js

This appears to not be happening:

```
ERROR in ./packages/ui/src/ConfigProvider.tsx
Module not found: Error: Can't resolve '@alienfast/material-ui-pickers/utils/MuiPickersUtilsProvider' in '/Users/kross/projects/js/packages/ui/src'
 @ ./packages/ui/src/ConfigProvider.tsx 25:50-121
 @ ./packages/ui/src/test/MockedAppProvider.tsx
 @ ./packages/ui/src/Downshift/__tests__/Downshift.stories.tsx
 @ ./.storybook/__generated__/stories.js
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/config/polyfills.js ./node_modules/@storybook/core/dist/server/config/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true
```

TrySound:
This project is not `@alienfast/material-ui-pickers`.

rosskevin:
> This project is not @alienfast/material-ui-pickers.
It is as I am repeatedly testing these builds downstream.

Further, this would be the source of the react context not working, if someone could direct import the `MuiPickersUtilsProvider` when the `DatePicker` loads/uses the `Consumer` from a different file.

TrySound:
This will work
```js
import MuiPickersUtilsProvider from 'material-ui-pickers/utils/MuiPickersUtilsProvider';
import { DatePicker } from 'material-ui-pickers/DatePicker';
```
and this will work
```js
import { MuiPickersUtilsProvider } from 'material-ui-pickers';
import { DatePicker } from 'material-ui-pickers';
```

if you don't mix this ways then everything will work

TrySound:
Are you sure you run all build scripts, not just `rollup -c`?

rosskevin:
- This does not work in webpack. `import MuiPickersUtilsProvider from 'material-ui-pickers/utils/MuiPickersUtilsProvider';`.  I cannot say this one more time.
- Yes, I am running `npm run build`.

Sample:
material-ui-pickers_esm_js_ _lib

TrySound:
I'm sorry. I forgot about emitDeclareationsOnly flag. It's probably should be removed to enable path imports.

rosskevin:
> I'm sorry. I forgot about emitDeclareationsOnly flag. It's probably should be removed to enable path imports.

Right, I did this before and had an exploded mirror of files which caused the react context failure (because some were loaded from the rollup produced esm and some direct).  This confirms my thinking below.
 
----

The only way to ensure that the user cannot screw this up when using modules (and given our knowledge of the failing new context api when used from different files) is this approach:

1. `esm` - babel or ts to create the build for `module` - exploded and with exploded .d.ts files
2. `umd/cjs` rollup

It seems clear to me now after diving down this rabbit hole.  Thanks for sticking with this @TrySound - I thought I was going nuts yesterday.

TrySound:
No. Leave esm and umd as is. Generate cjs and d.ts with tsc.
image

rosskevin:
I'll try again now that I have a better understanding.  Should be quick.

rosskevin:
@TrySound that definitely gets me past the error and I confirmed that I was mixing imports - index for the picker and direct for the provider causing the provider context message.  

I really wish we could produce a build where this context error was not a possibility.

TrySound:
I would just remove path imports and rely on tree shaking. It's possible with typescript transpiling since it's places all class properties inside class iife. Combining with babel will make code fully treeshakable.

TrySound:
Could you split type refactorings into separate PR? After merge we will focus on build improvements.

rosskevin:
I have done very little with the build - only trying to get it to work with the runtime.  Ultimately there should not be much of any changes there.

lib/package.json
TrySound: By the way babel runtime is not used anymore

rosskevin:
Build issue is solved - it was really an issue with my understanding but I improved the error message to smack people like me in the face.

We only have a few test errors remaining, I'll take another look at them - would like to get this merged asap.

rosskevin:
Ok, I was able to do a hard compare of current tests and previous tests (regarding props) and I fixed some props the wrong way and now are working.  

The remainder I need help with @dmtrKovalenko.  I've done all I can.  Current state:

```
Test Suites: 4 failed, 26 passed, 30 total
Tests:       6 failed, 85 passed, 91 total
```

All but one are `e2e`.  The one that is not e2e mystifies me because I don't readily see a code or dependency change.

rosskevin:
@dmtrKovalenko - confirmed - all that is left is make the remainder of tests run.  I also think you need admin access to trigger the travis build without cache.

rosskevin:
I am seeing this error my downstream project - will look at it now:

```
Uncaught TypeError: format_1.default is not a function
E [20:12:38] The above error occurred in the <DateTextField> component:
    in DateTextField (created by Context.Consumer)
```

rosskevin:
Ok, I think the above error is coming from another mismatch in imports between file vs index.  Above the utils are imported directly from a file path (which is cjs) and given to the provider.  I assume we don't want to rollup the utils from the index, is that correct?

My continuing feeling is that if we stopped the rollup esm and just exploded the esm we would not have the potential for mismatched imports (and bad react context) and errors like this could not happen.

rosskevin:
The above problems with nested `default.default` on only some functions is bizarre.  I check the date-fns source for both their cjs and esm and there is no structural difference in exports of a good `addDays` vs a bad `addMonths`.  I've got no idea what is going on.

TrySound:
This problem is caused by cjs/esm interop with default export which is broken in webpack and won't be fixed. So we need to provide esm version for utils too. I can take it.

rosskevin:
Thanks @TrySound - I'm linking for reference https://github.com/webpack/webpack/issues/7973

PR #696, opened 2018-10-17
D will parsed as ordinal number

cubicwork:
D will parsed as ordinal number in date-fns 2.0.0-alpha.11
E will parsed as date in date-fns 2.0.0-alpha.11

This PR closes #


cubicwork:
W

PR #700, opened 2018-10-18
Double extension of interfaces is not allowed in TS

snk5000:
## Description

There is an error in TS: 'An interface may only extend a class or another interface' in DateTextField, because double interface extensions are not allowed in TS (https://github.com/Microsoft/TypeScript/issues/16936) Changed interface to type in this PR.


snk5000:
It's clone https://github.com/dmtrKovalenko/material-ui-pickers/pull/695

PR #701, opened 2018-10-18
Provide esm entries for utils

TrySound:
Flatten provider and utils and provided esm version via additional
package.json in each util folder.

```js
import MuiPickersUtilsProvider from 'material-ui-pickers/MuiPickersUtilsProvider';
import dateFnsUtilsOld from 'material-ui-pickers/date-fns-utilsOld';
import dateFnsUtils from 'material-ui-pickers/date-fns-utils';
import momentUtils from 'material-ui-pickers/moment-utils';
import luxonUtils from 'material-ui-pickers/luxon-utils';
```

Here's how utils folder looks like
```
date-fns-utils/
  package.json
  index.js
  index.esm.js
  index.d.ts
```


rosskevin:
This problem seems to originate from `jest.doMock('../src/_shared/WithUtils', () =>`.  I commented it out and `DatePicker.test.js` ran fine - though there were other errors with other tests. 

In my PR I have done away with this mock as I changed the nature of `mount`.  When switching to typescript there are clear type errors in the way the shallow/mount were setup/used (and these errors exist here and on `develop`) and I believe this is causing the error.  These errors aren't obvious until types come into play.

So I'm +1 for this PR as it is a good change, but it won't work until the mock error is resolved (which I have done in my PR #695).  You could try to resolve the mock error here but I don't think that is a good allocation of time since I have already done that.  

If #695 gets merged soon, we can rebase this one prior to a release - I think that's a good sequence.

TrySound:
@dmtrKovalenko I fixed all errors without big rewrites. This can be merged first to simplify review of @rosskevin's PR.

lib/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "strictNullChecks": true,
     "strictFunctionTypes": true,
     "declaration": true,
+    "esModuleInterop": true,
```
dmtrKovalenko: How did you make this works with jest tests? Awesome!

PR #708, opened 2018-10-25
Fix initialFocusedDate update on prop change

luki86:
Fix initialFocusedDate update on prop change

If calendar initial date depends on `initialFocusedDate` prop, when we update that prop the initial date will not change. This PR fix that bug


piotros:
Great job @luki86! We experienced the same bug on production in our company. @dmtrKovalenko is release for this bugfix scheduled?

PR #709, opened 2018-10-25
Treeshake classes

TrySound:
`/*@__PURE__*/` is what uglify may eliminate if no usages is found. Sadly typescript does not provide correct annotation here.


dmtrKovalenko:
Hmm 🤔
Can you provide more info about /*@__PURE__*/

dmtrKovalenko:
And also @TrySound. Do we need to use `sideEffects: false` in package.json?

TrySound:
Search for `__PURE__` here https://github.com/mishoo/UglifyJS2
Yes, sideEffects will probably help with webpack.

dmtrKovalenko:
@TrySound would you like to join as collaborator to this project? https://github.com/dmtrKovalenko/material-ui-pickers/invitations

TrySound:
Sure

PR #712, opened 2018-10-25
Change how moment is imported

ja-bravo:
## Description

Fixes TS1192: Module moment as no default export while using Typescript.

### DIFF
--- a/lib/src/utils/moment-utils/index.ts
+++ b/lib/src/utils/moment-utils/index.ts
@@ -1,4 +1,4 @@
-import defaultMoment from 'moment';
+import * as defaultMoment from 'moment';
 import { Utils } from '../../typings/utils';
 
 interface Opts {


dmtrKovalenko:
For now use `esModuleIntertop: true` in your tsconfig json.

revmischa:
Great! Please release this on npm! version 1.0.1 on npm does not have this fix.

dmtrKovalenko:
@revmischa we will not release it. Use `esModuleIntertop: true`

revmischa:
`esModuleInterop: true` breaks all of my other imports, where we have `import * as moment from 'moment'` etc. turning on that option isn't a solution. this is the only module where it's a problem.

dmtrKovalenko:
But actually import * as moment is incorrect.
You can't use namespace as function. Using of `* as` broken a lot of our codebase

Mario-Eis:
But i think I got it right now. I'll investigate.

PR #719, opened 2018-10-25
Add tslib to dependencies

TrySound:
Ref https://github.com/dmtrKovalenko/material-ui-pickers/issues/718

/cc @dmtrKovalenko


TrySound:
@dmtrKovalenko Could you make a release with this fix?

dmtrKovalenko:
Ok I will release a patch today

PR #724, opened 2018-10-28
Allow overriding the picker component in the inline/model wrappers

Philipp91:
This allows users to use a customized DateTimePicker inside the wrappers. In the simplest case, users can just supply a function that does only minor modifications to the props of the picker and then instantiates the original DateTimePicker from the library.

### DIFF
--- a/lib/src/DatePicker/DatePickerInline.tsx
+++ b/lib/src/DatePicker/DatePickerInline.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 import { Omit } from '@material-ui/core';
 import BasePicker, { BasePickerProps } from '../_shared/BasePicker';
-import DomainPropTypes from '../constants/prop-types';
+import DomainPropTypes, {ComponentType} from '../constants/prop-types';
 import InlineWrapper, { InlineWrapperProps } from '../wrappers/InlineWrapper';
 import Calendar from './components/Calendar';
 import DatePicker, { BaseDatePickerProps } from './DatePicker';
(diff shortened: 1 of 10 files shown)


dmtrKovalenko:
I don't see any real use case of that

Philipp91:
In my application I'm customizing `DateTimePicker.handleViewChange()` to add a time part to a previously time-less date when the user switches to the time tab. And I modify `DateTimePicker.handleDayChange()` to not switch to the time tab automatically in case the current value does not have a time part.

Currently, I'm duplicating the entire `DateTimePickerModal.tsx` file to be able to use my custom DateTimePicker from there.

dmtrKovalenko:
Maybe you can do that by modifying `utils` passed to pickers?

Philipp91:
Unfortunately no. It is already possible to customize a great deal using the `utils`. But for this use case I want to modify the tab switching behavior. For example, I need to replace the hard-coded `DateTimePickerView.HOUR` [here](https://github.com/dmtrKovalenko/material-ui-pickers/blob/c9bc1c811c737c66c4c898a9506ffba2ab28b698/lib/src/DateTimePicker/DateTimePicker.tsx#L153) with an expression that switches to different views depending on the `date` value. This interacts with the `state` of the component.

dmtrKovalenko:
I am rejecting this PR because thats too much overflow with customization. 
Thats an overhead because these components only renders some components. I am actually thought about making this more automatically (not repass the same props for each component) 

In your case please create your custom root component outside our internal wrapper and picker and  do everything you want

Philipp91:
Ok thanks for taking a look.

PR #727, opened 2018-10-28
[RFC] Move utils to separate package + Lerna

dmtrKovalenko:
The main idea of the requested change is to split out own pickers package and utils to be available updates them independently and reuse the logic of utils for other packages. 

The breaking change for the user will be 
```diff
- import utils from 'material-ui-pickers/utils/date-fns-utils'
+ import utils from 'date-io/date-fns-utils'
```


TrySound:
Looks good. Not sure I would keep date-io inside of this project. It will be better to create `date-io` monorepo with splitted packages which would keep there dependencies in peers.

dmtrKovalenko:
That’s a really nice idea.

dmtrKovalenko:
@TrySound çreated new repo for date-io - https://github.com/dmtrKovalenko/date-io

TrySound:
Add me on twitter to enable DM

PR #730, opened 2018-10-29
Upgrade date-fns to alpha 21

TrySound:
I don't see breaking changes before this version. But they at least fixed esm for path imports.
https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha21


dmtrKovalenko:
Closing because I have already updated the date-fns version in @date-io repo

PR #782, opened 2018-11-26
Add today and clear buttons for inline pickers

RealMarshal:
This PR closes # 

## Description
1. Adds today and clear buttons for inline date, time and datetime pickers.
2. Allows autoOk prop to affect today button behavior. Since it is true by default for inline pickers, click on today button automatically selects the value and closes the picker.

### DIFF
--- a/lib/src/DatePicker/DatePickerInline.tsx
+++ b/lib/src/DatePicker/DatePickerInline.tsx
@@ -52,6 +52,8 @@ export const DatePickerInline: React.SFC<DatePickerInlineProps> = props => {
         handleChange,
         handleTextFieldChange,
         handleAccept,
+        handleSetTodayDate,
+        handleClear,
       }) => (
         <InlineWrapper
           disableFuture={disableFuture}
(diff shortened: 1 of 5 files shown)


RealMarshal:
Why not? This is rather disappointing, it is important functionality and it should work with inline pickers too. Documentation doesn't even mention that 'showTodayButton' works only with standard pickers.

stsiarzhanau:
@dmtrKovalenko Thanks for this awsome pickers. But I have the same question. What are the reasons to not support this functionality? Having ability to clear value in inline datepicker is extremely helpful. So it'd be great to have either "X" button next to calendar icon in date input itself (like in [react-datepicker](https://reactdatepicker.com/#example-25)) or "Clear" button in calendar popover itself.

dmtrKovalenko:
@stsiarzhanau  of course that will be nice 👍

kg-currenxie:
@dmtrKovalenko you know material design is only recommendations right?

> Dialogs **should** contain a maximum of two actions.
>
> ...
> Providing a third action such as “Learn more” is **not recommended** as it navigates the user away from the dialog, leaving the dialog task unfinished.
```

There are some cases where an adornment is not the best option either. A general-purpose-community-component like this should not have these kind of restrictions imo.

PR #785, opened 2018-11-28
Prettier line width

dmtrKovalenko:


TrySound:
:(

dmtrKovalenko:
@TrySound ?

TrySound:
I usually split my screen in 2 and only ~100 may fit into it.

PR #787, opened 2018-11-28
Feature/typed overrides

### LINKED ISSUE #469: Overriding styles on createMuiTheme with Typescript
Hi,

I am using material-ui-pickers on a project here, but got a problem when trying to override all MuiPicker styles like (MuiPickersToolbar, MuiPickersCalendarHeader, ...), since they are not Typed on the posible override list of Material-Ui.

Is there a way to do it?

dmtrKovalenko:
This PR closes #469  

## Description
Adding exported file `overrides.d.ts` which contains typings for material-ui theme `overrides` option.


TrySound:
I guess this is because of syntax lack. Flow types solves this problem easily.
```js
import type { MyType } from './MyType.js';
export type {
  MyType
}
```

dmtrKovalenko:
I think that's a luck of rollup :)

TrySound:
Rollup doesn't know about typescript types. Typescript compiler without knowing about types removes only invalid js syntax. But these imports are valid in js but they does not exist in js. I don't know how to solve it. Ask in rollup-plugin-typescript maybe.

dmtrKovalenko:
Hmm, I will. Maybe then we should remove rollup, and use just typescript compiler?

dmtrKovalenko:
I think I found solution! 💡 
We can create `index.js` for rollup and `index.ts` for ts compiler that will not loose the typings 😸

TrySound:
Cool. And rollup is not a problem.

PR #799, opened 2018-12-08
add dateRangeIcon and timeIcon to docs

Dozalex:

### DIFF
--- a/docs/src/prop-types.json
+++ b/docs/src/prop-types.json
@@ -901,6 +901,28 @@
       "required": false,
       "type": { "name": "ReactNode" }
     },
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
This file is auto generated by tsdoc comments upto props in the lib folder

docs/src/prop-types.json
```diff
@@ -901,6 +901,28 @@
       "required": false,
       "type": { "name": "ReactNode" }
     },
+    "dateRangeIcon": {
```
dmtrKovalenko: Move description to the props declaration

PR #810, opened 2018-12-18
[typescript] Reexport props

dmtrKovalenko:
This PR closes # 

## Description
Optimize development bundle time by using only one rollup config.
Properly reexport typescript types


wichert:
Nice, I was just looking for this!

PR #818, opened 2018-12-25
Optimize components rerenders 

dmtrKovalenko:
Rerendering optimization to make the project be blazing fast 🔥


dmtrKovalenko:
Todo update version of  react in peer dependencies before merge

PR #838, opened 2019-01-07
[#837] Export consumer component

### LINKED ISSUE #837: Export context consumer
I've created a set of Picker components for a project using `createAutoCorrectedDatePipe` from text-mask-addons. To make this work seamlessly I'm using the `disablePast`, `disableFuture`, `minDate`, and `maxDate` props to set the correct `minYear` and `maxYear` settings for the pipe.

These components will be part of an internal component library that we are currently developing. They should not be locked to a specific date library, and instead always use the same as material-ui-pickers uses.

Using the context consumer from material-ui-pickers would ensure that I was always using the same date utils without duplicating the context myself.

**Describe alternatives you've considered**
My solution for now has been to provide the same utils object in my own context so that I can consume it. But this creates internal dependencies that I'd prefer to avoid. 

**Additional context**

skoging:
This PR closes #837

## Description

Export the MuiPickersContextConsumer used internally.

### DIFF
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -30,4 +30,4 @@ export { default as TimePickerView } from './TimePicker/components/TimePickerVie
 
 export { default as Clock } from './TimePicker/components/Clock';
 
-export { default as MuiPickersUtilsProvider } from './MuiPickersUtilsProvider';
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
I am wondering where is travis build... 
Thats strange a lot

PR #840, opened 2019-01-08
Automatically import icons from @material-ui/icons if package installed

dmtrKovalenko:
Get rid of required font-icon by default


TrySound:
Shit, please don't. Never mix cjs with esm. Burn cjs everywhere. This will probably work only with bundlers which covers all possible and impossible hacks with interop.
Rollup will require some string/replace operations. I just got rid from similar shit in react-map-gl. Don't introduce the new one.

dmtrKovalenko:
@TrySound then how we can achieve `optionalDependencies` work?

TrySound:
material-ui icons is another large shit which cannot be converted to esm. I replaced it with a downloaded set of svg files and svgr transformer.

TrySound:
Do not use optionalDependencies. material-ui core doesn't use it too.

dmtrKovalenko:
But really a lot of our users are using the @material-ui/icons. And I want to get rid of icon fonts

TrySound:
material-ui/icons are useless. svg files is the only necessary source of truth for icons IMO.

TrySound:
Icons fonts in your project? Or in pickers?

dmtrKovalenko:
But @material-ui/icons are also .svg file. They are svg components, no matter. 
But we live in the material-ui ecosystem...

Will dynamic import help?

dmtrKovalenko:
I want to make an easy way for pickers to use svg icons automatically. But don't want to make a breaking change :)

TrySound:
Nope. I'd say it's only for apps. Even core use internal icons.

TrySound:
Do you mean via strings or what?

TrySound:
I don't understand what is the problem now? You don't want to have own icons? They are just a few bytes.

dmtrKovalenko:
Now we are rendering just `<Icon> icon_name </Icon>` by default. And if somebody wants to override the icon he needs to pass the icon component to the picker via props. 
That requires new wrapper to be created and passed down all props. That sucks. 
My idea was - when you already have `@material-ui/icons` installed icons will be imported from there.

TrySound:
Let's talk in twitter DM

dmtrKovalenko:
@TrySound looks better?

TrySound:
Yeah, better. Just rename please PR

docs/src/Pages/Components/DateTimePicker/CustomDateTimePicker.example.jsx
```diff
@@ -1,7 +1,9 @@
 import React, { PureComponent, Fragment } from 'react';
 import { DateTimePicker } from 'material-ui-pickers';
-import { IconButton, Icon, InputAdornment } from '@material-ui/core';
-import { withUtilsService } from '../../../_shared/UtilsServiceContext';
+import { IconButton, InputAdornment } from '@material-ui/core';
```
TrySound: Oopsy, path imports please. Or whole mui will appear in bundle

dmtrKovalenko:
Ohhh, god, I did not save the rename 🤦‍♂️

docs/src/Pages/Components/DateTimePicker/CustomDateTimePicker.example.jsx
dmtrKovalenko: 🤦‍♂️

dmtrKovalenko: But this is a docs file! @TrySound I am bundling full material-ui in docs 🤕

TrySound: Okay then

PR #855, opened 2019-01-11
Migrate from classnames to nicer clsx

TrySound:
## Description

- esm support
- out of the box typescript support
- probably faster


TrySound:
The size is slightly bigger because material ui still uses classnames but I'm working on it.

FranBran:
`TypeError: __WEBPACK_IMPORTED_MODULE_13_clsx___default(...) is not a function`

seems like my setup can't handle *.mjs files properly?

TrySound:
@FranBran Do you use typescript?

PR #859, opened 2019-01-14
Remove keycode in favour of event.key

TrySound:
Ref https://twitter.com/olivtassinari/status/1084819202412818432

We can already see size improvement (which will be visible for end user
only after material-ui upgrade) in 1.5kB.

This is one less commonjs dependency!


dmtrKovalenko:
Tests failed

PR #874, opened 2019-01-22
Add year and month props

### LINKED ISSUE #796: Please add support pick Month like YearPicker
Duplicate
https://github.com/dmtrKovalenko/material-ui-pickers/issues/355

But maybe you can add some switch case [here](https://github.com/dmtrKovalenko/material-ui-pickers/blob/develop/lib/src/DatePicker/DatePicker.tsx#L147) by some prop

We need something like that
https://material.angular.io/components/datepicker/overview

OleksiiKukuruza:
This PR closes #796

## Description
Add year and month props similar to seconds in ```<DateTimePicker />```

Year only picker:
screen shot 2019-01-22 at 16 38 34

Year and month picker:
screen shot 2019-01-22 at 16 38 44


OleksiiKukuruza:
Most important note here is that it's necessary to add few utils

dmtrKovalenko:
linter errors
```
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/DatePicker/DatePicker.tsx:10:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/DatePicker/DatePicker.tsx:11:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/DatePicker/components/MonthSelection.tsx:2:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/DatePicker/components/MonthSelection.tsx:5:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/DatePicker/components/MonthSelection.tsx:9:3 - comment must start with a space
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/typings/overrides.ts:11:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/dmtrKovalenko/material-ui-pickers/lib/src/typings/overrides.ts:12:1 - Import sources within a group must be alphabetized
```

lib/src/DatePicker/DatePicker.tsx
```diff
@@ -48,10 +61,16 @@ export class DatePicker extends React.PureComponent<DatePickerProps & WithUtilsP
...
     maxDate: '2100-01-01',
     openToYearSelection: false,
+    year: false,
+    month: false,
   };
 
-  public state = {
-    showYearSelection: Boolean(this.props.openToYearSelection),
+  public state: DatePickerState = {
+    openView: Boolean(this.props.openToYearSelection || this.props.year)
```
dmtrKovalenko: I think here it should be the prop`openTo` the same as on date time picker. We can left `openToYear` but make a enum instead of 2 bools

OleksiiKukuruza: Maybe it will be nice to have something like 
```
initialView?: 'year' | 'month' | 'day';
maxView?: 'year' | 'month' | 'day';
minView?: 'year' | 'month' | 'day';
```
later it can be scaled to make possible to pick any precision, like pick only month and day or only month, what do you think?

dmtrKovalenko: Thats seems to be an option. But we definitely need `'year' | 'month' | 'day'` type for the views. Similar to date time picker view.

OleksiiKukuruza: So what is the final decision?

dmtrKovalenko: The final decision - for now add `openTo` property. The same as in DateTimePicker with enum value `'year' | 'month' | 'day'`. 
Also we don't want to introduce breaking changes so `openToYearSelection` should work as well but marked as deprecated. Then we will remove it. 

About minView and maxView - I am considering. The naming is not really obvious. I am considering something like. `availableView` which will be an `Array<'year' | 'month' | 'day'>`. For month picker something like `availableView=['month']`. But that can be done later

OleksiiKukuruza: Sorry but I still did not understand how to replace ```year``` and ```month``` boolean props.

```openTo``` is about initial view, it's clear.

Comment (OleksiiKukuruza) [reply lib/src/DatePicker/DatePicker.tsx]: @dmtrKovalenko so what about that year and month props?

Comment (dmtrKovalenko) [reply lib/src/DatePicker/DatePicker.tsx]: @OleksiiKukuruza instead of 2 bools describing which options to add I propose to add array of views that will be shown.

So it will be possible to just show month picker and nothing more.

Comment (dmtrKovalenko) [reply lib/src/DatePicker/DatePicker.tsx]: @OleksiiKukuruza And please - rebase your branch from up-to-date develop. Cause you're breaking coverage reporter :)

**Comment by OleksiiKukuruza** [comment]
@dmtrKovalenko we need to release date-io utils to fix failing tests and replace few functions with functions from date-io utils

**Comment by dmtrKovalenko** [comment]
@OleksiiKukuruza will do the release today. And hope will help you with tests.

File: lib/src/DatePicker/DatePicker.tsx [hunk 2]
```diff
@@ -34,12 +46,25 @@ export interface BaseDatePickerProps {
...
+export interface DatePickerProps
+  extends BaseDatePickerProps,
+    WithStyles<typeof styles>,
+    WithUtilsProps {
   date: MaterialUiPickersDate;
   onChange: (date: MaterialUiPickersDate, isFinished?: boolean) => void;
 }
 
-export class DatePicker extends React.PureComponent<DatePickerProps & WithUtilsProps> {
+interface DefaultBaseDatePickerProps {
```
dmtrKovalenko: I do not understand prospective of this interface

lib/src/DatePicker/DatePicker.tsx
```diff
@@ -48,11 +73,23 @@ export class DatePicker extends React.PureComponent<DatePickerProps & WithUtilsP
...
-  public state = {
-    showYearSelection: Boolean(this.props.openToYearSelection),
-  };
+  public state: DatePickerState;
+
+  constructor(props: DatePickerProps) {
+    super(props);
+
+    const { availableViews, openTo, openToYearSelection } = this.props as PropsWithDefaults;
+    this.state = {
```
dmtrKovalenko: move state to the public field.

lib/src/DatePicker/DatePickerModal.tsx
```diff
@@ -29,6 +44,8 @@ export const DatePickerModal: React.SFC<DatePickerModalProps> = props => {
     rightArrowIcon,
     shouldDisableDate,
     value,
+    availableViews = [DatePickerView.YEAR, DatePickerView.MONTH, DatePickerView.DAY],
```
dmtrKovalenko: should be moved to default props

dmtrKovalenko:
I think the PR if fine now.

lib/src/DatePicker/DatePicker.tsx
OleksiiKukuruza: Oh, now when tests failed I see why, that hack was added

PR #889, opened 2019-01-29
Refactor/calendar

dmtrKovalenko:


lib/src/DatePicker/components/Calendar.tsx
```diff
@@ -111,39 +116,38 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
   public validateMinMaxDate = (day: MaterialUiPickersDate) => {
     const { minDate, maxDate, utils } = this.props;
 
-    return (
+    return Boolean(
       (minDate && utils.isBeforeDay(day, utils.date(minDate))) ||
```
TrySound: Boolean casting is evil. Will minDate != null work for you?

dmtrKovalenko: Yes and no. I think here if somebody will pass `""` - boolean casting will save our ass from crash

dmtrKovalenko: But mostly I do not want to break something. Just reduce cognitive complexity of the code

PR #891, opened 2019-01-29
Redux Form integration guide

depiction:
This PR adds a Guide for integrating with Redux Form. It took awhile to get this datepicker working with Redux Form on my project because there is no documentation for it. To make it easier for other, I decided to add an example using the code I ended up using on my project.


docs/src/Pages/Guides/ReduxForm.example.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+export const required = value => {
+  if (value && typeof value === 'string') {
+    value = value.trim();
+  }
+  return value ? undefined : 'Required';
+};
+
+export class ReduxDatePicker extends React.Component {
+  state = {
```
dmtrKovalenko: But how are you getting this state on external validation? OnSubmit? It will submit anyway is there error or not?

depiction: The ReduxDatePicker's state is for presentation purposes only. It displays the errors that come from DatePicker and the required method. The validation built into DatePicker only sets a value to the date prop when it's built in validation checks pass. Then the required method passed to Field on line 73 ensures that the field is populated to determine if the form may be submitted or not.

dmtrKovalenko: This will not work for most users. I will create guide later

pixelass: Any news here? We are trying to get this done in a clean manner. An example or guide would be highly appreciated.

pixelass: bump.. any updates here?

Can you define "later"?

dmtrKovalenko:
We are working now on v3 that will significantly simplify form integration experience

taschetto:
Looking forward to using this awesome library with redux-form... :)

taschetto:
@dmtrKovalenko @pixelass you may want to look at this: https://github.com/dmtrKovalenko/material-ui-pickers/issues/147#issuecomment-487043001

PR #892, opened 2019-01-30
Add support for minutes step

etairi:
This PR closes https://github.com/dmtrKovalenko/material-ui-pickers/issues/809

## Description

This PR adds support for minutes step to the TimePicker component.


dmtrKovalenko:
That should also work for DateTimePicker? Can you please also pass the step to the dateTimePicker component

etairi:
@dmtrKovalenko Thanks for the feedback, I just did that.

etairi:
> 
> 
> That should also work for DateTimePicker? Can you please also pass the step to the dateTimePicker component

@dmtrKovalenko Thanks for the feedback, I just did that.

PR #922, opened 2019-02-17
Fix rollup warning about broken sourcemap

TrySound:
By replacing `@class` annotation with `@__PURE__` we broke rollup
internal sourcemap generator.

Found the fix here
https://github.com/ex-machine/rollup-plugin-ts-treeshaking/blob/master/index.js

image


dmtrKovalenko:
Why don't just using the `ts-treeshaking-plugin`?

TrySound:
Another one package for simple task which wraps the module we already use. Currently we see which exactly transformation it does and may manipulate it as we want.

PR #924, opened 2019-02-19
New docs

dmtrKovalenko:
Here I have recreated docs site from scratch.
@TrySound can I ask your feedback? One thing I am complaining - I love how folders splitter right now, when we have 2 separate folders, one for dev and one for lib, but that leads problems with dependency resolving. So here Ive changed development flow. 

1. Build lib to `docs/material-ui-pickers-build` folder (internal one`
2. Proxy imports of `material-ui-pickers` to `/material-ui-pickers-build` from next app.

It looks like a crutch, but I am really don't want to wrap the whole project in next app. Maybe you have better idea?


TrySound:
We use yarn workspaces and do not have problems with sharing packages. Yarn also have explicit links like `"material-ui-pickers": "link:../lib"`. Don't have an idea how to do this with npm. I'm not a fan of it.

dmtrKovalenko:
They are using links. That won’t help...
With links we will have 2 reacts in the bundle. That’s sucks

TrySound:
Hm.. yeah, then only workspaces will help :)

TrySound:
Yay!

Issue #919, opened 2019-02-12
Error validation not working with Formik and Yup


SebSchwartz:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers |     2.2.1    |
| material-ui         |      3.9.2   |
| React               |   16.8.1     |
| Browser             |   Chrome   71   |
| Peer library        |    date-fns(date-io): 1.1.0      |
| Formik             |  1.5.0   |
| Yup             |  0.26.10   |

## Steps to reproduce
1. A form using formik with an InlineDatePicker and formik configured for Yup schema validation.
2. The shema of validation validates some stuffs on the fields of InlineDatePicker. Here as schema definition for those files:
```
startDate: Yup.date().required(),
  endDate: Yup.date().when('startDate', (startDate, schema) => (startDate ? schema.min(startDate) : schema.date())),
```
3. There is no errors at all that appears. If I remove those two schema validation, it works perfectly.

## Expected behavior
The picker does not block schema validation.

## Actual behavior
Schema validation is no more working for any field on the schema.

dmtrKovalenko:
We know that - in v3 (#862) we will definitely recreate for integration process. Possibly duplicate of #836.

dan-turner:
@dmtrKovalenko are you able to please provide an example of how to do this properly now?

jdmairs:
I'm also trying to figure out how to use yup with the date-picker.

nilamsavani91:
> I'm also trying to figure out how to use yup with the date-picker.

I also want to set the validation if the endDate is less than StartDate. I am using KeyboardDatePicker from material-ui pickers.

maxmann74:
I am having this issue as well.  The "Invalid Date" error shows up, but when I clear the value, the "Required" error never shows up.

Issue #836, opened 2019-01-06
Add option to totally ignore inbound validation


benneq:
**Is your feature request related to a problem? Please describe.**
I want to use your `DatePicker` with external validation. Though I'd like to have `onChange` to trigger even for invalid inputs.

**Describe the solution you'd like**
Some new prop like `ignoreInvalid`. And if it's set to `true`, then `onChange` will always get called with either a valid or invalid value. It could even trigger for every key stroke. Just like a normal input does.
And the input should always display the value. And **not** display "Invalid value" or "unknown".

**Describe alternatives you've considered**
I already tried to use `onError` to catch the invalid value. But this will then pass this value again into the `DatePicker`, which results into the input displaying "Unknown" and then it's impossible to edit again, because `onError` will trigger again and again.

I also tried using `labelFunc` with `labelFunc={(moment) => moment._i || moment.format("D.M.YYYY")}`. But this again will make the input impossible to edit.

**Additional context**
I'm using `formik`. Though `onError={value => formik.setFieldValue(fieldName, value)}` will pass the invalid value to `<DatePicker value={...} />`. And then it won't work anymore.

Here's a Code Sandbox: https://codesandbox.io/s/1zzzkrwx7l

If you clear the DatePicker it will display "required".
If you input something invalid, it will display "must be valid"
But when it's invalid, you won't be able to edit anymore.

benneq:
That‘s not the problem. The validation integration between material-ui-pickers and formik does not work well in real world applications. `formik.setFieldError` is useless when you want use a `validationSchema`. I'll provide a better Code Sandbox in a minute.

benneq:
I just copied and slightly modified the code you referenced: https://codesandbox.io/s/645q7p8qqk

I added a `TextField` to the form and a submit button.
Then I added some extra validation for the whole form using `Formik.validate`:
1. `date` is required
2. `string` is required

Steps to reproduce the issue:
1. Make the `date` invalid. Now the error message from `DatePicker.onError` appears. `isValid` is now `false`. 
2. Blur the `TextField`. Now the "required" error message for `string` appears. `isValid` is still `false`.
3. Insert some text into the `TextField`. **Now the error on `DatePicker` disappears, because formik's `validate` method will overwrite the whole `errors` object. And `isValid` is now `true`.**

What's the issue with that?
[...]

PR #946, opened 2019-03-09
RFC: New architecture for state managment

dmtrKovalenko:
# New material-ui-pickers architecture ships in [v3](https://github.com/dmtrKovalenko/material-ui-pickers/issues/862)

## Reason 🏅 
The main reason of integrated changes are issues like #919 and #836. For now we are using internal state to manage both keyboard input and selection date flow. But there are problems under the hood > the biggest one is **validation of entered date outside**.  🤕 

Because of internal state management we are getting in the deadlock when trying to sync keyboard input with the parsed date. Anyway, not controlled components are not really *best practice*. Still, we intend on DX so high order user flow will not change. The same `value` and `onChange` props.

From now on we are using a separate control for the keyboard mode `<KeyboardDatePicker />`. That allows better tree-shaking of keyboard specific logic and better flexibility. It uses new signature for `onChange` method: `(date: Date, value: string) => void`. 
Also will be added `inputValue` prop for controlling over the pure entered string. 

### This PR changes picker state management to be:

- [x] Create a top-level custom hook for picker state management
- [x] Truly controlled `value`, without additional state under the hood.
- [x] Move `open` wrapper state up to the hook.
- [x] Move validation logic up to the hook.
- [x] **Make inline mode work via prop `variant="inline"` instead of separate component** 
- [x] Recreated from scratch keyboard mode:

- [x] Fix all components
- [x] Remove unused code
- [x] Add an ability to control opened state via prop
- [x] Fix all documentation pages
- [x] Make sure that all props documented correctly


andreypopp:
One more suggestion is to remove dependency injection machinery via context and instead inject needed date utils via props.

andreypopp:
Hope it's ok I'm posting more comments here regarding the API design.

After thinking about this more an interesting approach would be to abstract handling of date/time values into a container which will be opaque to the picker components. Such container would bundle a concrete value itself + a set of operatons on it (likely ones form date-io + some more specific to pickers).

Example:

    import * as React from 'react'
    import {Value} from 'material-ui-pickers'

    function App(props) {
      let [value, setValue] = React.useState(() =>
        Value.create(initialValue, dateUtils));

      return <DatePicker value={value} onValue={onValue} />
    }

Why this is good:

1. Value and utils are configured through the same API.

   This is the correct approach they depend on each other (you can't pass moment to a picker configured with date-fns utils).

   Furthermore this way it's easier to enfore such invariant via type system.  With the current design I don't think it's possible to do so as utils are passed via context.

   I think this point is super important.

2. No need to use context.

   I don't know why context was used in the first place. But if the reason was to prevent "prop drilling" scenario then I think it's not worth it.

   In controlled mode you still need to pass value/onChange props - you can pass utils as well too then. If you choose to bundle them in the same value container - it's just a single prop then (in addition to onChange).

   Such API without context also looks simpler to me (subjectively of course).

   Note that I'm not anti-context in general, just think this isn't the use case for it.

3. No need to introduce custom hooks.

   Again, I'm not anti-hooks in general, just that in that particular case the only hook which is needed is state.

4. (BONUS) Such API is less boilerplatly when you need differently configured pickers in the same app - one using moment and another date-fns. Unlikely but still... you don't have to wrap those pickers into different context providers.

Overall I'm very excited about this API redesign! Thank you for working on this.

dmtrKovalenko:
@andreypopp Thanks a lot for your vision. But there are a lot of reason for such design:
1. Timezones => if somebody using not default timezone, but custom one that's super simple to just pass configured luxon or moment object with particular timezone. Now it works perfectly with pickers.
2. I do think that DX is super important, so I am really aware of passing each time utils. That make additional complexety of integration I think. Also as a part of v3 I did think to use more strict typing of `value` in `onChange` - `Luxon | Dayjs | Moment | Date` and use typescript module augmentation to make it works out of scope. 
3. Thats can be really possible to add passing utils per component instance. But I still think that it needs to be an option. Partly for `less-breaking` updating and partially for DX.

andreypopp:
> 1. Timezones => if somebody using not default timezone, but custom one that's super simple to just pass configured luxon or moment object with particular timezone. Now it works perfectly with pickers.

Yeah! Sorry for not being clear. The `Value` will wrap some concrete value (moment or luxon) and utils. So you would still pass moment object with particular timezone or something else.

> 2. I do think that DX is super important, so I am really aware of passing each time utils. That make additional complexety of integration I think. Also as a part of v3 I did think to use more strict typing of `value` in `onChange` - `Luxon | Dayjs | Moment | Date` and use typescript module augmentation to make it works out of scope.

I agree about DX being important, I just don't think context helps there (in fact I think it's vice versa).

You can do for your app the following to abstract the boilerplate of passing utils each time:
```
function createValue(moment: Moment) {
  return Value.create(moment, momentUtils)
}
```

There's no reason you need to inject utils per UI component tree, it should be done in "data layer" instead, in my opinion.

dmtrKovalenko:
Agreed. But I think nobody will use 2 utils in one project

andreypopp:
> Agreed. But I think nobody will use 2 utils in one project

Yeah, sure, this is a pretty unlikely use case! I don't think it warrants using context just because of that though...

dmtrKovalenko:
I will think about that deeply. I thinks that a good catch. Thank you a lot @andreypopp

RedHatter:
I second an api that doesn't require context.

dmtrKovalenko:
Getting back to this huge PR...

dmtrKovalenko:
I think it's good to merge

andreypopp:
🎉🎉🎉 Congrats on getting this merged! This looks great and I'm trying this out right now.

The only thing I'm not sure how to approach is how to make components not perform any validations on value. Is it possible to disable?

dmtrKovalenko:
Yep, you can pass custom `helperText` and `error` and that's will totally ignore all inner validation

andreypopp:
Few more items I've found:
1. When clicking on `Clear` button in keyboard datepicker there's a traceback from inside the datepicker.
2. KeyboardDatePicker always tries to show the current date even if I pass `inputValue={null}` or `inputValue=""`

PR #992, opened 2019-04-17
Hide ok label if explicitly empty

### LINKED ISSUE #991: Hide ok label when explicitly empty
Hey, nice work. Here's a gif:

olimpgif-2

I think we maybe shouldn't show `okLabel` when its emptied on purpose. Since we have situations in which it makes sense to not have an ok label (autoOk) and I don't think it ever makes sense to show it like this: 

image
which is what happens when its given a `""` or `null` value now

A solution would be to hide it when explicitly passing these values to the `okLabel` prop.

rafael-vitor:
This PR closes #991


dmtrKovalenko:
We are not supporting `v2` anymore. Can you please create a PR to `next` branch? And maybe also do the same for `Cancel` button? 🤔

rafael-vitor:
@dmtrKovalenko done :) https://github.com/dmtrKovalenko/material-ui-pickers/pull/994

PR #1001, opened 2019-04-23
Use @material-ui/style esm imports

dmtrKovalenko:


lib/src/DatePicker/components/Calendar.tsx
```diff
@@ -5,7 +5,7 @@ import DayWrapper from './DayWrapper';
 import CalendarHeader from './CalendarHeader';
 import EventListener from 'react-event-listener';
 import SlideTransition, { SlideDirection } from './SlideTransition';
-import withStyles, { WithStyles } from '@material-ui/styles/withStyles';
+import withStyles, { WithStyles } from '@material-ui/core/styles/withStyles';
```
TrySound: You sure about that? I thought the main idea is to move all style stuff into own package.

dmtrKovalenko: Thanks that’s seems to be missing :)

PR #1027, opened 2019-05-05
Chore/update rollup

dmtrKovalenko:
## Description
Get back umd bundle :)


lib/rollup.config.js
```diff
@@ -41,6 +41,7 @@ const globals = {
   '@material-ui/core/Tabs': 'material-ui.Tabs',
   '@material-ui/styles': 'material-ui.styles',
   '@material-ui/core/SvgIcon': 'material-ui.SvgIcon',
+  '@material-ui/core/styles/colorManipulator': 'material-ui.styles.colorManipulator',
```
TrySound: I think guys don't want to provide two level path imports as public api. Only `@material-ui/core/styles` is public api.

lib/rollup.config.js
```diff
@@ -91,43 +95,43 @@ export default [
...
+      file: 'build/dist/material-ui-pickers.umd.min.js',
+    },
+    plugins: [
+      nodeResolve({ extensions }),
+      typescriptPlugin({ typescript, tsconfig }),
+      babel(babelOptions),
+      commonjs(commonjsOptions),
+      replace({ 'process.env.NODE_ENV': JSON.stringify('production') }),
+      sizeSnapshot(),
+      uglify(),
```
TrySound: Use terser instead. It's more stable.

PR #1041, opened 2019-05-13
Fix openTo of DateTimePicker

marudor:
## Description
When you reworked DateTimePicker from class to hooks you forgot to support openTo.
This fixes it (defaults to "date") and adds a test for it.


dmtrKovalenko:
Thanks a lot for your work! But it will already will be fixed in scope of #1044

PR #1043, opened 2019-05-17
Update calculation of Calendar slide direction

olexandrp:
## Description
Calculate Calendar slide direction with day precision rather than month.
If we update calendar date through the props it slides in a wrong direction.
calendar-wrong-slide-direction


dmtrKovalenko:
Thanks, but can you please change the base to “next” branch?

olexandrp:
> Thanks, but can you please change the base to “next” branch?

I add new pull request. #1048

PR #1052, opened 2019-05-21
Async month changed callbacks

lindgren-linus:
## Description

I found a need for fetching date validation information asynchronously from a server and thought that _onMonthChange_ would be the perfect callback function for fetching more information about each month.  

The property _onMonthChange_ can now return a _Promise_ (but can still return _void_, so this change won't break anything) and the Calendar component will await the callback while displaying a loading indicator. 

async_month_change


lib/src/DatePicker/components/Calendar.tsx
```diff
@@ -5,7 +5,7 @@ import DayWrapper from './DayWrapper';
 import CalendarHeader from './CalendarHeader';
 import EventListener from 'react-event-listener';
 import SlideTransition, { SlideDirection } from './SlideTransition';
-import { Theme } from '@material-ui/core';
+import { Theme, CircularProgress } from '@material-ui/core';
```
dmtrKovalenko: Make path import instead of core

lib/src/DatePicker/components/Calendar.tsx
```diff
@@ -136,12 +138,20 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
     this.props.onChange(utils.mergeDateAndTime(day, date), isFinish);
   };
 
-  public handleChangeMonth = (newMonth: MaterialUiPickersDate, slideDirection: SlideDirection) => {
+  public handleChangeMonth = async (
```
dmtrKovalenko: I am not sure that we need async/await here for one option. It will lead to some additional bundle size. 
Use .then

lib/src/__tests__/DatePicker/Calendar.test.tsx
```diff
@@ -24,6 +23,40 @@ describe('Calendar', () => {
     component.find('CalendarHeader').prop<any>('onMonthChange')(utilsToUse.date());
     expect(onMonthChangeMock).toHaveBeenCalled();
   });
+
+  it('Should not display loading indicator when synchronous', () => {
```
dmtrKovalenko: The tests must be moved to e2e. Or at least using mount.

lib/src/DatePicker/components/Calendar.tsx
```diff
@@ -284,18 +294,42 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
           transKey={currentMonth!.toString()}
           className={classes.transitionContainer}
         >
-          <div>{this.renderWeeks()}</div>
+          <>
+            {(this.state.loadingQueue > 0 && (
+              <div className={classes.progressContainer}>
+                <CircularProgress />
```
cevr: Must this render a CircularProgress? I believe this should be an option or at the very least able to be replaced by the user. @dmtrKovalenko What do you think?

lib/src/__tests__/DatePicker/Calendar.test.tsx
lindgren-linus: Thanks for the feeback! 
I could definitely move the tests but I'm wondering if they really are e2e tests? 

Except from the confusing test names (who clearly mentions visibility of elements, which is kindof misleading and should be changed), the tests really only checks so that the state of the component is changed properly.

Mount is probably more suitable for these kind of tests, I agree, but does state manipulation in components belong to e2e tests?

dmtrKovalenko: Yeep, it is more integration then e2e. 
We are on the road to testing public interface rather then internal component logic

dmtrKovalenko: mount tests just named `e2e` for some reason

lib/src/DatePicker/components/Calendar.tsx
lindgren-linus: Fixed

lib/src/DatePicker/components/Calendar.tsx
lindgren-linus: Fixed

lib/src/__tests__/DatePicker/Calendar.test.tsx
lindgren-linus: Moved the tests and modified them now

lib/src/DatePicker/components/Calendar.tsx
lindgren-linus: Good idea! I added support for an optional loadingIndicator property that would replace the CircularProgress

lindgren-linus:
Thank you for the response!
I have now implemented the changes that was requested.

dmtrKovalenko:
Sorry for long delay. Was pretty busy last week

dmtrKovalenko:
Thank you! I`ve been right now adding your branch as upstream to push it myself!
Sorry for long delay, you r the best!

PR #1075, opened 2019-05-31
Remove react-event-listener

TrySound:
This package is not really necessary with hooks. It's commonjs and its babel
helpers are bundled not efficiently by rollup.

I removed tests. Maybe it will be easier to cover with cypress.


lib/src/DatePicker/components/Calendar.tsx
```diff
@@ -70,6 +69,17 @@ export interface CalendarState {
   lastDate?: MaterialUiPickersDate;
 }
 
+const KeyDownListener = ({ onKeyDown }: { onKeyDown: (e: KeyboardEvent) => void }) => {
```
TrySound: Where is type inference? I didn't find.

lib/src/wrappers/ModalWrapper.tsx
```diff
@@ -63,21 +63,26 @@ export const ModalWrapper: React.FC<ModalWrapperProps<any>> = ({
...
-  const handleKeyDown = React.useCallback(
-    (event: KeyboardEvent) => {
-      switch (event.key) {
-        case 'Enter':
-          onAccept();
-          break;
-        default:
-          return; // if key is not handled, stop execution
-      }
+  React.useEffect(() => {
```
dmtrKovalenko: It would much clear to reuse created below `KeydownListener`

lib/src/__tests__/DatePicker/Calendar.test.tsx
```diff
@@ -50,54 +50,3 @@ describe('Calendar - disabled selected date on mount', () => {
...
-        minDate={new Date('01-01-2018')}
-        onChange={onChangeMock}
-        theme={{ direction: 'ltr' } as any}
-        allowKeyboardControl
-        {...props}
-      />
-    ));
-  });
-
-  it('Should render go to prev week on up', () => {
```
dmtrKovalenko: I am not really sure, do we have this tests in `e2e`. Anyway I planned to move them to cypress

lib/src/wrappers/ModalWrapper.tsx
TrySound: Reusing will require to use useCallback. Now handler is builtin into effect.

dmtrKovalenko: Why not to create useKeyDown for both cases?

TrySound: It has the same problem

TrySound: Okay, I will make it

TrySound: Done

PR #1091, opened 2019-06-07
Fix "todayLabel" prop type description.

NiklasMM:
## Description

Fixes description and defaultValue for prop `todayLabel`

### DIFF
--- a/lib/src/wrappers/ModalWrapper.tsx
+++ b/lib/src/wrappers/ModalWrapper.tsx
@@ -6,7 +6,7 @@ import { WrapperProps } from './Wrapper';
 import { useKeyDown } from '../_shared/hooks/useKeyDown';
 import { DialogProps as MuiDialogProps } from '@material-ui/core/Dialog';
 
-export interface ModalWrapperProps<T = {},> extends WrapperProps<T> {
+export interface ModalWrapperProps<T = {}> extends WrapperProps<T> {
   /**
    * "OK" label message
    * @default 'OK'
--- a/lib/src/wrappers/ModalWrapper.tsx
+++ b/lib/src/wrappers/ModalWrapper.tsx
@@ -6,7 +6,7 @@ import { WrapperProps } from './Wrapper';
 import { useKeyDown } from '../_shared/hooks/useKeyDown';
 import { DialogProps as MuiDialogProps } from '@material-ui/core/Dialog';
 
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
This is autogenerated. Comes from jsdoc comments

NiklasMM:
aw snap... sorry. I'll fix it

dmtrKovalenko:
1. Change description
2. run `yarn docgen`

NiklasMM:
btw, is there a reason why the default values are in all caps in the description, but the actual default values aren't?

dmtrKovalenko:
It is a limitation of react-docgen-typescript :(

NiklasMM:
`yarn docgen` puts all the output into one line. I guess that should be configured in this line: https://github.com/mui-org/material-ui-pickers/blob/ff22b9e25470549534482f1c7bc7b203638f4271/docs/scripts/docgen.js#L73

Telling `stringify` to use two spaces also generates a different format. Any trick I'm missing here?

dmtrKovalenko:
No worries. We are storing it in minfied version. 
It is github feature it is pretiffying this file

dmtrKovalenko:
Or prettier 😟

NiklasMM:
But the file is prettified on my file system and if I add it to the commit all lines are changed. Do I not add it to the commit? Sorry, I'm a bit confused here 😅

dmtrKovalenko:
Just commit it :D 
No worries

NiklasMM:
Oh, I see, there's some pre-commit magic going on. 

Thanks for walking me through it ;)

lib/src/wrappers/ModalWrapper.tsx
```diff
@@ -6,7 +6,7 @@ import { WrapperProps } from './Wrapper';
 import { useKeyDown } from '../_shared/hooks/useKeyDown';
 import { DialogProps as MuiDialogProps } from '@material-ui/core/Dialog';
 
-export interface ModalWrapperProps<T = {}> extends WrapperProps<T> {
+export interface ModalWrapperProps<T = {},> extends WrapperProps<T> {
```
dmtrKovalenko: This `,`

dmtrKovalenko:
```suggestion
export interface ModalWrapperProps<T = {}> extends WrapperProps<T> {
```

PR #1092, opened 2019-06-07
Fix Typescript override example

ygerg:
## Description

Material-ui style override definitions have changed, and the Typescript example provided is out of date.

Credit: https://stackoverflow.com/a/54890034 
Might be the same nick on Github, ping @arvymetal


degardinn:
@ygerg Indeed, this is my old Github account...
Thanks for the doc update, credits, and the remembrance that this old account still exists 😉

dmtrKovalenko:
@degardinn do you like pies? :D 🥧

degardinn:
@dmtrKovalenko That's not a pie, it's my actual face

Issue #1088, opened 2019-06-04
Cancel button shouldn't update with the current date


chriswardo:
Currently if no date is set in DatePicker (and therefore the emptyLabel is displayed) and the user opens the picker and presses the 'cancel' button, the onChange function is called with today's date for some reason.

Thanks!

bogdansoare:
also having this issue

stopak:
I'm also having this issue

pankaz:
This behaviour is observed for the "clear" button too.

dmtrKovalenko:
Fixed in 3.3.1

PR #1036, opened 2019-05-08
Check for null date object on keyboard picker change handler

njetsy:
This PR closes #1035

## Description

Check to see if the date object is null when handling a change on the keyboard picker. Otherwise an error is thrown when pressing the clear button.

PR #1095, opened 2019-06-08
Inner state for picker modal

### LINKED ISSUE #1049: Picking a date from calendar should not modify the TextInput until the user clicks OK
The issue is clearly demonstrated from the demo: https://material-ui-pickers.dev/demo/datepicker

Open a calendar -> start clicking on dates -> you will see the underlying TextInput change.

dmtrKovalenko:
This PR closes #1049 

## Description
This PR make pickers to use the inner state for displaying inside modal. To reduce side-effects for other API. It will still dispatch `onChange` for keyboard mode but will dispatch `onChange` for standard pickers only when the date is accepted.

### DIFF
--- a/lib/src/_shared/hooks/usePickerState.ts
+++ b/lib/src/_shared/hooks/usePickerState.ts
@@ -37,10 +37,10 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
 
   useEffect(() => {
     // if value was changed in closed state - treat it as accepted
-    if (!isOpen) {
+    if (!isOpen && !utils.isEqual(pickerDate, date)) {
       setPickerDate(date);
     }
-  }, [date, isOpen]);
+  }, [date, isOpen, pickerDate, utils]);
 
   const acceptDate = useCallback(
     (acceptedDate: MaterialUiPickersDate) => {
--- a/lib/src/_shared/hooks/usePickerState.ts
+++ b/lib/src/_shared/hooks/usePickerState.ts
@@ -2,13 +2,13 @@ import { useMemo } from 'react';
 import { useUtils } from './useUtils';
 import { IUtils } from '@date-io/core/IUtils';
 import { MaterialUiPickersDate } from '../..';
+import { useOpenState } from './useOpenState';
 import { BasePickerProps } from '../../typings/BasePicker';
+import { useCallback, useDebugValue, useEffect, useState } from 'react';
 import { getDisplayDate, validate } from '../../_helpers/text-field-helper';
-import { useCallback, useDebugValue, useEffect, useRef, useState } from 'react';
 
 export interface StateHookOptions {
   getDefaultFormat: () => string;
-  getValidationError?: () => React.ReactNode;
 }
 
 const valueToDate = (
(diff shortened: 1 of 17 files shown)


dmtrKovalenko:
Also closes #1088

TrySound:
Here ok button is missing but month click does not close popup. Is it intentional?
image

dmtrKovalenko:
I guess it is unrelated

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -24,112 +24,82 @@ const valueToDate = (
...
-
-    if (props.onAccept) {
-      props.onAccept(acceptedDate);
-    }
-  }, [acceptedDateRef, setIsOpen, props]);
+      if (props.onAccept) {
+        props.onAccept(acceptedDate);
+      }
+    },
+    [setIsOpen, props]
```
Philipp91: AFAIU, the `props` object always changes. It's possible, in theory, that the component that passes the props has them memoized, but in all normal circumstances (using JSX, or call sites like [this](https://github.com/mui-org/material-ui-pickers/blob/68779c7237976c8f56278322885aee980cbc3ed5/lib/src/_shared/hooks/useKeyboardPickerState.ts#L51)), the `props` object will be a new object. And `useCallback` [only memoizes](https://github.com/facebook/react/blob/425473f43fd95a564f07ffcf6dbd105cef9f78b9/packages/react-reconciler/src/ReactFiberHooks.js#L342) when the `deps` are the same according to `Object.is()`, so the `useCallback()` here is ineffective. It can either be removed (if we don't care about memoization) or be fixed:

```
const {onAccept, onChange} = props;
const acceptDate = useCallback(
  (acceptedDate: MaterialUiPickersDate) => {
    setIsOpen(false);
    onChange(acceptedDate);
     if (onAccept) {
      onAccept(acceptedDate);
    }
  },
  [setIsOpen, onAccept, onChange]
);
```

If you agree (I might be wrong, I'm very new to hooks), I can send a PR for this and similar occurrences.

dmtrKovalenko: Yep, that will be more efficient

dmtrKovalenko: It will not be changed if only internal state changed.  It anyway it will be cool to have more clear deps for top level hook

Philipp91:
Off-topic question: Is there a way I can test changes like this one in my own app before they have been released to NPM? I tried cloning the repo, running the build and then pointing the NPM dependency from my app to the `lib/build` directory. Everything has the right structure, but at runtime it complains that React was loaded twice (which breaks hooks, "Invariant Violation: Invalid hook call.").

dmtrKovalenko:
You can make a build and link your app manually for the folder lib/build. 
 Go to lib/build
 Run npm link
 Go to your app
 Run npm link @material-ui/pickers

Philipp91:
Thanks! I didn't know that `npm link` exists. The commands you gave lead to the same symlink I had before (just via the global `node_modules` directory), so it didn't resolve the duplicate-React issue. But knowing that `npm link` is the canonical way to do this allowed me to research this issue. I tried many things that try to fix the symlink situation, but none of them worked. Instead, I did this:
https://medium.com/@vcarl/problems-with-npm-link-and-an-alternative-4dbdd3e66811

```
material-ui-pickers/lib/build/ $ npm pack
my-app/ $ npm install ../material-ui-pickers/lib/build/material-ui-pickers-3.1.0.tgz
```

pankaz:
Kindly merge https://github.com/mui-org/material-ui-pickers/pull/1036 into the "next" branch. The "clear" functionality is breaking in the current "next" branch.

PR #1098, opened 2019-06-09
[Refactoring] Reorganize folder structure

dmtrKovalenko:


TrySound:
I would flatten folders as much as possible. Grouping can be achieved with name prefixed like Month and MonthView

PR #1099, opened 2019-06-09
Allow customized views inside the Picker

Philipp91:
This allows users to swap out the `YearSelection` or `Calendar` (etc.) components hard-coded inside the `Picker` with a custom component that has a compatible interface. The code is still quite hacky and I wouldn't merge it like this (either switch all to `React.createElement()` or none). As a side effect, this PR fixes the confusing situation that the `viewsMap` values were never used.
(An alternative design would be to expose separate `YearComponent`, `CalendarComponent`, etc props.)

I'm sending this out early to ask if offering this level of customization to users is something you want to do with this library.

I read somewhere in `material-ui`'s or `material-ui-picker`'s documentation that the libraries are designed to be open for customization. In the past, a related `ViewContainerComponent` prop existed (see https://github.com/mui-org/material-ui-pickers/pull/531), which I had added, but the design was a bit weird (it exposed an implementation detail) and it was rightfully removed, as far as I can tell. I think this new solution is more generic/flexible (composition over inheritance/replacement), but at the same time more type-safe (see `typeof viewsMap`) and future-proof. Users of this API would have to accept that the contract is: "you supply a component that behaves like the built-in one", which is intentionally an indirect and volatile definition. As shown in the example code, users can write custom components that make only minimal assumptions about the built-in components they replace, so in practice they're unlikely to break.

### DIFF
--- a/docs/pages/guides/CustomizeView.example.jsx
+++ b/docs/pages/guides/CustomizeView.example.jsx
@@ -1,0 +1,39 @@
+import React, { Fragment, useState } from 'react';
+import { DateTimePicker, TimePickerView, useUtils } from '@material-ui/pickers';
+import { Button } from '@material-ui/core';
+
+const CustomTimePickerView = props => {
+  const utils = useUtils();
+  return (
+    <div style={{ position: 'relative' }}>
+      <TimePickerView {...props} />
+      <Button
+        style={{ position: 'absolute', zIndex: 10, top: 0 }}
+        onClick={() => {
+          props.onHourChange(utils.mergeDateAndTime(props.date, new Date('2018-01-01 13:37')));
+        }}
+      >
+        {' '}
+        13:37{' '}
+      </Button>
+    </div>
+  );
+};
+
+function CustomizeViewExample() {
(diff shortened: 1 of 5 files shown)


dmtrKovalenko:
Actually that is smth I planned to integrate. expose top-level Picker component.

Philipp91:
Cool, then I'll implement this more cleanly? I propose using `React.createElement()`, as it's less verbose. Or did you have a different design in mind?

dmtrKovalenko:
Closing this for now, it will be reimplemented soon

dmtrKovalenko:
Closing that. Sorry but it is something that will be implemented by the core maintainers. And in a bit other way. 
Anyway thanks for your work!

Princezhm:
@dmtrKovalenko is any progress on this? thank you :)

Philipp91:
Is there a new GitHub issue for this, or any other way we can know when it's done?

Philipp91:
@dmtrKovalenko Please see the question above -- how can I know when the alternative implementation is done?

dmtrKovalenko:
Never. It is so big level of customization that we are not are not tent to support

PR #1101, opened 2019-06-09
[Refactoring] Better hook memoization

Philipp91:
Context: https://github.com/mui-org/material-ui-pickers/pull/1095#pullrequestreview-247398372

Since a mistake in this code would introduce subtle bugs, depending on how good the test coverage is for these fils, I suggest an extra-thorough review. The changes in each file are technically independent from the other files.

### DIFF
--- a/lib/src/_shared/hooks/usePickerState.ts
+++ b/lib/src/_shared/hooks/usePickerState.ts
@@ -1,11 +1,10 @@
-import { useMemo } from 'react';
 import { useUtils } from './useUtils';
 import { IUtils } from '@date-io/core/IUtils';
 import { MaterialUiPickersDate } from '../..';
 import { useOpenState } from './useOpenState';
 import { BasePickerProps } from '../../typings/BasePicker';
-import { useCallback, useDebugValue, useEffect, useState } from 'react';
 import { getDisplayDate, validate } from '../../_helpers/text-field-helper';
+import { useCallback, useDebugValue, useEffect, useMemo, useState } from 'react';
 
 export interface StateHookOptions {
   getDefaultFormat: () => string;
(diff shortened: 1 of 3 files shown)


lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -45,13 +45,10 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
   const acceptDate = useCallback(
     (acceptedDate: MaterialUiPickersDate) => {
       setIsOpen(false);
-      props.onChange(acceptedDate);
-
-      if (props.onAccept) {
-        props.onAccept(acceptedDate);
-      }
+      onChange(acceptedDate);
+      onAccept && onAccept(acceptedDate);
```
dmtrKovalenko: please leave if here

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -74,32 +71,33 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
...
+  }, [onError, validationError, value]);
 
+  const inputValue = getDisplayDate(date, format, utils, value === null, props);
   const inputProps = useMemo(
     () => ({
       validationError,
-      onClick: () => !props.disabled && setIsOpen(true),
-      inputValue: getDisplayDate(date, format, utils, props.value === null, props),
+      onClick: () => !disabled && setIsOpen(true),
+      inputValue,
```
dmtrKovalenko: here make inputValue as a first key in the object

dmtrKovalenko:
I will clone the branch and will explicitly check if all deps are working properly. But right now LGTM. Thanks for help 👍

Philipp91:
Should I address the review comments now, or are you doing that yourself on the cloned branch?

PR #1189, opened 2019-07-17
[docs] Add redux form integration example

igornfaustino:
Add a new example to the docs, It can be seen on the "Form integration" topic and refers about how to integrate the pickers with the redux-form


docs/pages/_app.tsx
```diff
@@ -30,9 +32,11 @@ class MyApp extends App<{ theme: ThemeType }> {
     const { Component, pageProps, theme } = this.props;
 
     return (
-      <PageWithContexts initialTheme={theme} pageContext={this.pageContext}>
-        <Component pageContext={this.pageContext} {...pageProps} />
-      </PageWithContexts>
+      <ReduxProvider store={store}>
```
dmtrKovalenko: Please wrap only the particular example in redux provider. So redux will be included in bundle only on the forms page

dmtrKovalenko:
Overall is cool, thank you.

docs/pages/guides/ReduxForm.example.jsx
```diff
@@ -0,0 +1,64 @@
...
+      value={value ? new Date(value) : null}
+      disabled={submitting}
+      onBlur={() => onBlur(value ? new Date(value).toISOString() : null)}
+      error={error && touched}
+      onChange={date => Date.parse(date) && inputProps.onChange(date.toISOString())}
+    />
+  );
+};
+
+class ReduxFormExample extends React.PureComponent {
```
dmtrKovalenko: Function components and hooks are more prefered

docs/pages/guides/ReduxForm.example.jsx
```diff
@@ -0,0 +1,64 @@
...
+      <form>
+        <Grid container>
+          <Grid item container justify="center" xs={12}>
+            <Field name="date" component={DateField} />
+          </Grid>
+          <Grid item xs={12} sm={12} style={{ margin: '24px' }}>
+            <Code
+              children={JSON.stringify(
+                {
+                  values: {
```
dmtrKovalenko: I think here we can pass some values from props. Actually redux-form's values and errors. 
So it will be more clear for users

docs/pages/guides/form-integration.mdx
```diff
@@ -20,3 +21,9 @@ Here are some examples!
 Here is an example of how to integrate @material-ui/pickers with Formik
 
 <Example source={FormikExample} />
+
+#### Redux Form Integration
+
+@material-ui/pickers can also be integrate with Redux Form
```
dmtrKovalenko:
```suggestion
@material-ui/pickers can also be integrated with Redux Form
```

docs/pages/guides/ReduxForm.example.jsx
igornfaustino: most of redux form props are on Field level, I can't get in a simple way the error text outside the Field component... so I add a valid props from Redux-Form, if the value is "true" no errors were find on the form

docs/tsconfig.json
```diff
@@ -19,5 +19,5 @@
     "target": "esnext",
     "resolveJsonModule": true
   },
-  "include": ["../docs/**/*.ts*"]
```
dmtrKovalenko: I think you must use .tsx for store. It should not be complicated. ASFAIK there is should not be any type definitions

docs/pages/guides/store.jsx
```diff
@@ -0,0 +1,13 @@
+import { reducer as formReducer } from 'redux-form';
+import { createStore, combineReducers } from 'redux';
+
+const rootReducer = combineReducers({
```
dmtrKovalenko: Or maybe this file even need to be inside example. So somebody can just copy and paste it to the codesandox

docs/package.json
```diff
@@ -62,12 +62,17 @@
     "react-dom": "^16.8.2",
     "react-jss": "^8.6.1",
     "react-kawaii": "^0.12.0",
+    "react-redux": "^7.1.0",
+    "redux": "^4.0.4",
+    "redux-form": "^8.2.4",
+    "redux-thunk": "^2.3.0",
```
dmtrKovalenko: I don’t think we need thunk here

docs/pages/guides/ReduxForm.example.jsx
```diff
@@ -0,0 +1,76 @@
...
+  date: selector(state, 'date'),
+  initialValues: {
+    date: new Date().toISOString(),
+  },
+});
+
+const createReduxForm = reduxForm({ form: 'example' });
+const Form = connect(mapStateToProps)(createReduxForm(ReduxFormExample));
+
+const app = () => (
```
dmtrKovalenko: Ok and a small change. Please make it capitalized `App`

docs/tsconfig.json
dmtrKovalenko: Following prev comment. Just put it inside the example

PR #1206, opened 2019-07-27
Add babel transpilation to esm

### LINKED ISSUE #663: Webpack treeshaking
It would be great to mark this library as side effect free, which is the case as far as I can see, this is the case. Then webpack 4 could leave out unused code during the build.

It is just one line in the package.json.
```
"sideEffects": false
```

See this page for more info:
https://webpack.js.org/guides/tree-shaking/

### LINKED ISSUE #1132: This library is not tree-shaking friendly
## Environment

| @material-ui/pickers |  3.1.1 |
| material-ui          |  4.1.1 |
| React                | 16.8.6 |
| Browser              | any |
| Peer library         |  date-fns 2.0.0-alpha.34 |  |

## Steps to reproduce
Tree-shaking with code splitting is problematic due to unnecessary bundling by rollup for es and cjs targets

## Expected behavior
Pickers should be packaged like material-ui core, especially pickers are under mui organisation. I am referring to 
```js
{
  "main": "./index.js",
  "module": "./esm/index.js",
}
```
they have in package.json. Notice, they give path to not bundled index.js files. However, for pickers paths target bundled library by rollup. 

Rollup is nice for UMD target, but for cjs and es this is really redundant, people using bundlers on app level will bundle vendor files anyway. Prebundling hurts tree-shaking and does not give any advantage what so ever. Webpack tree shaking is based on unused reexports and prebundling library won't have those but a flat structure.

Also, we cannot even do import like `import DatePicker from @material-ui/pickers/DatePicker` which would target a component with es module, because the only es file is bundled. This is problematic because of Webpack issue https://github.com/webpack/webpack/issues/7782 , so you cannot tree shake a given library across multiple chunks. The remedy is to import components like `import DatePicker from @material-ui/pickers/DatePicker`, but because this is not of es type, it leads to disaster because material-ui/core components are not imported as es modules, so actually they will be duplicated in date picker chunk.

So please, do not bundle for es and common js targets. And if you really want to bundle for those, do it additionally, but don't link those in `main` and `module` in package.json

## Actual behavior
As above

## Live example
n/a

dmtrKovalenko:
This PR closes #1132  

## Description
Experimenting with tree-shaking. In fact, really babel exports enable tree-shaking for webpack. 🐒

### DIFF
--- a/lib/babel.config.js
+++ b/lib/babel.config.js
@@ -1,0 +1,14 @@
+module.exports = {
+  presets: [
+    '@babel/preset-react',
+    '@babel/preset-typescript',
+    ['@babel/preset-env', { modules: false }],
+  ],
+  plugins: [
+    ['@babel/plugin-proposal-class-properties', { loose: true }],
+    ['@babel/plugin-proposal-object-rest-spread', { loose: true }],
+    ['@babel/plugin-transform-runtime', { useESModules: true }],
+    // for IE 11 support
+    '@babel/plugin-transform-object-assign',
+  ],
+};
--- a/lib/babel.config.js
+++ b/lib/babel.config.js
@@ -5,10 +5,12 @@ module.exports = {
     ['@babel/preset-env', { modules: false }],
   ],
   plugins: [
+    'optimize-clsx',
     ['@babel/plugin-proposal-class-properties', { loose: true }],
     ['@babel/plugin-proposal-object-rest-spread', { loose: true }],
     ['@babel/plugin-transform-runtime', { useESModules: true }],
     // for IE 11 support
     '@babel/plugin-transform-object-assign',
+    './remove-prop-types.js',
   ],
 };
--- a/docs/prop-types.json
+++ b/docs/prop-types.json
@@ -368,6 +368,19 @@
         "name": "\"portrait\" | \"landscape\""
       }
     },
+    "ToolbarComponent": {
+      "defaultValue": null,
+      "description": "Component that will replace default toolbar renderer",
+      "name": "ToolbarComponent",
+      "parent": {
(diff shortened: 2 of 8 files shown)


lib/babel.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  presets: [
+    '@babel/preset-react',
+    '@babel/preset-typescript',
+    ['@babel/preset-env', { modules: false }],
+  ],
+  plugins: [
+    'optimize-clsx',
+    ['@babel/plugin-proposal-class-properties', { loose: true }],
+    ['@babel/plugin-proposal-object-rest-spread', { loose: true }],
```
TrySound: This is a part of env now

lib/babel.config.js
```diff
@@ -0,0 +1,16 @@
...
+    ['@babel/plugin-transform-runtime', { useESModules: true }],
```
TrySound: Looks like this affects all code. Better to add this and remove useESModules here
```
env: {
  esm: {
    plugins: [['@babel/plugin-transform-runtime', { useESModules: true }]]
  }
}
```

lib/babel.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  presets: [
+    '@babel/preset-react',
+    '@babel/preset-typescript',
+    ['@babel/preset-env', { modules: false }],
```
TrySound: Remove modules here. Rollup adds it automatically.

lib/babel.config.js
TrySound: This issue will be fixed
https://github.com/mui-org/material-ui-pickers/issues/1219

PR #1209, opened 2019-07-27
Fix dispatching 2 onChange events with variant="inline"

### LINKED ISSUE #1202: DatePicker onChange event is fired twice!
Hello, 
I am having an issue with DatePicker component. The onChange event is fired twice, causing the setState to be called twice. I was trying to find a way to use event.preventDefault to fix it, however the onChange event only expose one parameter which the date.
Is there a way to prevent this?

## Environment

| @material-ui/pickers |^3.2.0|
| material-ui          |^4.2.0|
| React                |^16.8.6|
| Browser              |Crome 75|
| Peer library         |momentjs ^2.24.0

## Steps to reproduce

1- Click on the date picker input field
2- The inline calendar will popup
3- Click on a day
4- The date field will be updated
5- The onChange event will be fired twice

## Expected behavior
I am expecting the the onChange event to fire one time only.

## Actual behavior
The onChange event is fired twice

## Live example

https://codesandbox.io/s/material-ui-pickers-usage-demo-woym3

See in the console that the onChange event fire twice!

```
import React, { useState } from "react";
import MomentUtils from "@date-io/moment";
import moment from "moment";
import { DatePicker, MuiPickersUtilsProvider } from "@material-ui/pickers";

function App() {
  const [selectedDate, setDate] = useState(moment());

  const onDateChange = val => {
    console.log("--- date changed: ", val.toString());
    setDate(val);
  };
  return (
    <MuiPickersUtilsProvider utils={MomentUtils}>
      <DatePicker
        disableToolbar
        autoOk
        variant="inline"
        label="Select new date"
        format="MM/DD/YYYY"
        value={selectedDate}
        onChange={onDateChange}
      />
    </MuiPickersUtilsProvider>
  );
}

export default App;
```

dmtrKovalenko:
This PR closes #1202

### DIFF
--- a/lib/src/_shared/hooks/usePickerState.ts
+++ b/lib/src/_shared/hooks/usePickerState.ts
@@ -77,6 +77,8 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
 
         if (variant === 'inline' || variant === 'static') {
           onChange(newDate);
+          onAccept && onAccept(newDate);
+          return; // simulate autoOk, but do not close the mod
         }
 
         if (isFinish && autoOk) {
@@ -84,7 +86,7 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
         }
       },
     }),
-    [acceptDate, autoOk, onChange, pickerDate, variant]
+    [acceptDate, autoOk, onAccept, onChange, pickerDate, variant]
   );
 
   const validationError = validate(value, utils, props);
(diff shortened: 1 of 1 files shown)


lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -77,14 +77,16 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
 
         if (variant === 'inline' || variant === 'static') {
           onChange(newDate);
+          onAccept && onAccept(newDate);
+          return; // simulate autoOk, but do not close the mod
```
dmtrKovalenko:
```suggestion
          return; // simulate autoOk, but do not close the modal
```

PR #1218, opened 2019-07-30
Release v3.2.1

dmtrKovalenko:


dmtrKovalenko:
We need actually to make our releases more stable and automate. 
Ideally to be deployed from CI on merging to master

PR #1223, opened 2019-08-02
Include babel runtime helpers for each build type

### LINKED ISSUE #1219: Failing jest tests with added @babel/runtime dependency in 3.2.1
## Environment

| @material-ui/pickers | 3.2.1        |
| material-ui          | 4.3.0        |
| React                | 16.8.6        |
| Browser              | node@10.12.0 / jest        |
| Peer library         | date-io/moment@1.3.6, moment@2.24.0        |  |
| typescript | 3.5.2 |

## Steps to reproduce

Upgrade from 3.2.0 to 3.2.1.

## Expected behavior

Jest tests are passing like before.

## Actual behavior

All Jest tests are failing

```
 FAIL  src/App.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    ~/some_path/node_modules/@babel/runtime/helpers/esm/extends.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){export default function _extends() {
                                                                                             ^^^^^^

    SyntaxError: Unexpected token export

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:403:17)
      at Object.<anonymous> (node_modules/@material-ui/pickers/dist/material-ui-pickers.js:7:32)
```

It appears that this is caused by the new dependency `"@babel/runtime": "^7.2.0",` which made it into the bundled `package.json`.

## Live example

None.

dmtrKovalenko:
This PR closes #1219

### DIFF
--- a/lib/.size-snapshot.json
+++ b/lib/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "build/dist/material-ui-pickers.esm.js": {
-    "bundled": 102254,
-    "minified": 56597,
-    "gzipped": 15261,
+    "bundled": 103367,
+    "minified": 56989,
+    "gzipped": 15350,
     "treeshaked": {
       "rollup": {
-        "code": 45950,
-        "import_statements": 1831
+        "code": 46075,
+        "import_statements": 1795
       },
       "webpack": {
-        "code": 52389
+        "code": 52510
       }
     }
   },
   "build/dist/material-ui-pickers.umd.js": {
-    "bundled": 1028834,
-    "minified": 354290,
-    "gzipped": 97571
+    "bundled": 1072936,
+    "minified": 365383,
+    "gzipped": 100760
   },
   "build/dist/material-ui-pickers.umd.min.js": {
-    "bundled": 800771,
-    "minified": 299821,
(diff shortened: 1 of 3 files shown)


lifeiscontent:
@dmtrKovalenko when are we going to see a release?

dmtrKovalenko:
Ohh goodness. I am really sorry 😐 
I will try today. But cannot promise anything.

dmtrKovalenko:
Released in v3.2.2

PR #1248, opened 2019-08-11
[InlineWrapper]: Don't force any width on the paper

### LINKED ISSUE #1243: Inline Date Picker is too narrow on iPad landscape
The bug can be observed on the official homepage, https://material-ui-pickers.dev/demo/datepicker, when you open one of the inline datepickers on a viewport that corresponds to iPad landscape (1024x768).

One of the elements has a fixed width set to 310px, through the MuiPickersInlineWrapper-popoverPaper class. I can't quite track down who sets it under which condition.

image

wereHamster:
Fixes #1243

I don't see any reason to force a specific width on the Paper element. The other wrappers (Modal, Static) don't do it either.

[edit]: The Modal wrapper sets a min-width (actually, the ModalDialog does it). Still, I don't quite understand why min-width is needed.

### DIFF
--- a/lib/src/wrappers/InlineWrapper.tsx
+++ b/lib/src/wrappers/InlineWrapper.tsx
@@ -1,25 +1,9 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import clsx from 'clsx';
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
Can you please fix CI errors

PR #1253, opened 2019-08-13
[useKeyDown] Change key event handlers only when active state changes

wereHamster:
I noticed that the hook re-registers the keydown/up event handlers whenever it's called (React useEffect uses `==` to compare the dependencies, and `{} != {}`), but it only needs to do that when the active state changes. The code was almost correct, it was using a ref (`keyHandlersRef`) to hold the key handlers, but that ref was not being used from inside the DOM event handlers.

I also changed useIsomorphicEffect to useEffect because there is no need to run that immediately, since it doesn't affect layout.


dmtrKovalenko:
Actually it is affected layout. For the SSR purpose

wereHamster:
Maybe I'm missing something, but I fail to see how this line affects layout:

```
keyHandlersRef.current = keyHandlers;
```

Furthermore, registering key down event handlers is not something that makes sense to do during SSR.

PR #1266, opened 2019-08-17
[a11y] Open picker by space/enter click

### LINKED ISSUE #1059: Tab key does not focus controls in date picker
## Environment

| @material-ui/pickers |  3.0.0       |
| material-ui          |   4.0.0      |
| React                |  16.8.6       |
| Browser              |  Chrome 74       |
| Peer library         |  date-fns 2.0.0-alpha.27       |  |

## Steps to reproduce

1. Open a datepicker on the demo page (https://material-ui-pickers.dev/demo/datetime-picker)
2. Press the tab key

## Expected behavior

The tab key should move between the toolbar controls and days of the month, like in version 2 of material-ui-pickers.

## Actual behavior

The tab key does not seem to focus anything.

## Live example

https://material-ui-pickers.dev/demo/datetime-picker

### LINKED ISSUE #1231: Using tab if i focus on input the calendar should open or on the press of enter.
I'm using the input in a form and it'd be great to be able to tab through the fields and auto open the pickers or atleast when i hit enter key. 
I attempted to set open property on focus and on close i use to set open as false but it seems that the on focus event triggered again.
An onFocus prop for the pickers that only is called when the input is first entered OR when the user focused using tab and hit enter the picker should appear.

Apologies if I missed an existing setting.

dmtrKovalenko:
This PR closes #1059 and resolves #1231 
Also contains all changes from, so fixes #1253 

## Description
Open picker by space/enter click

### DIFF
--- a/docs/prop-types.json
+++ b/docs/prop-types.json
@@ -782,6 +782,19 @@
         "name": "(date: DateIOType) => void"
       }
     },
+    "onOpen": {
+      "defaultValue": null,
+      "description": "On open callback",
+      "name": "onOpen",
+      "parent": {
+        "fileName": "material-ui-pickers/lib/src/typings/BasePicker.tsx",
+        "name": "BasePickerProps"
+      },
+      "required": false,
+      "type": {
+        "name": "() => void"
+      }
+    },
     "format": {
       "defaultValue": null,
       "description": "Format string",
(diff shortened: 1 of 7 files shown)


TrySound:
Wait, enter? You sure about this? Enter is used to send forms. Isn't it the case?

dmtrKovalenko:
@TrySound yeah you are right, but it should also open the interactive controls? Shouldn't it?

TrySound:
Mixing behaviours is confusing IMO

TrySound:
Only space :)

PR #1366, opened 2019-10-26
Added error handling for scrollIntoView (fix for #1365)

### LINKED ISSUE #1365: TypeError: 'block' member of ScrollIntoViewOptions 'center' is not a valid value for enumeration ScrollLogicPosition
I'm working on an app that needs to support Chrome and Firefox, though because it's run on a corporate network the browsers aren't the latest. 

I'm using the DatePicker and everything works fine in Chrome, though in Firefox I get White Screen-ed when I try to select a year. In the console the error is:

`TypeError: 'block' member of ScrollIntoViewOptions 'center' is not a valid value for enumeration ScrollLogicPosition`

The version of FF my users have is version 52. From what I've found online [1,2], before Firefox 58, nearest and center values for the block option were unsupported.

For the time being I've just modded the lib to put a try catch block around this line: (not sure if there's a better fix)

https://github.com/mui-org/material-ui-pickers/blob/next/lib/src/views/Year/YearView.tsx#L47

That stops my app from crashing and allows the user to select a year. 

I also tried turning my component into a class and adding the componentDidCatch method, and that prevents the White Screen, but the date picker popup closes, preventing the user from selecting a year.

## Environment

| @material-ui/pickers |    3.2.6     |
| material-ui          |     4.5.1    |
| React                |     16.9.0    |
| Browser              |    Firefox v52     |
| Peer library         |     1.3.6    |  |

## Steps to reproduce

1. Use DatePicker in Firefox in a verison less than 58.
2. Click the date field to open up the date selection popup.
3. Click the Year in the popup to change the year.
4. App crashes.

## Expected behavior

App not to crash.

## Actual behavior

App crashes.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1389274
[2] https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

patorjk:
This PR closes https://github.com/mui-org/material-ui-pickers/issues/1365

## Description

Added error handling for the scrollIntoView method, which errors in Firefox versions before 58 (according to the [Material UI docs](https://material-ui.com/getting-started/supported-platforms/) Firefox versions >= 52 are supported).

How to test this:

* Download and use Firefox v52: https://ftp.mozilla.org/pub/firefox/releases/
* Go here to see the error: https://material-ui-pickers.dev/demo/datepicker

You'll see it gives a white screen, look in the console to see the ScrollIntoView error message.

In this PR, do the same thing except go here: http://localhost:3000/demo/datepicker - no error occurs. Open a DatePicker and click the Year. See that you can select a year just fine.

### DIFF
--- a/docs/prop-types.json
+++ b/docs/prop-types.json
@@ -62,7 +62,7 @@
     },
     "showTodayButton": {
       "defaultValue": {
-        "value": "false"
+        "value": false
       },
       "description": "If true today button will be displayed <b>Note*</b> that clear button has higher priority",
       "name": "showTodayButton",
@@ -77,7 +77,7 @@
     },
     "clearable": {
       "defaultValue": {
-        "value": "false"
+        "value": false
       },
       "description": "Show clear action in picker dialog",
       "name": "clearable",
@@ -3768,7 +3768,9 @@
       }
     },
     "minDate": {
-      "defaultValue": null,
+      "defaultValue": {
+        "value": null
+      },
       "description": "Min date",
       "name": "minDate",
       "parent": {
@@ -3781,7 +3783,9 @@
       }
     },
     "maxDate": {
-      "defaultValue": null,
+      "defaultValue": {
+        "value": null
+      },
       "description": "Max date",
       "name": "maxDate",
       "parent": {
@@ -3795,7 +3799,7 @@
     },
     "disablePast": {
       "defaultValue": {
-        "value": "false"
+        "value": false
       },
       "description": "Disable past dates",
       "name": "disablePast",
@@ -3810,7 +3814,7 @@
     },
     "disableFuture": {
       "defaultValue": {
-        "value": "false"
+        "value": false
       },
       "description": "Disable future dates",
       "name": "disableFuture",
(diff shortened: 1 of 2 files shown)


dmtrKovalenko (changes requested):
I think in this case we need just do nothing. If there is no window.scrollIntoView.

`if (!window.scrollIntoView)` 
Because try {} catch (e) can make some performance degradation for server side rendering apps

patorjk:
That test unfortunately won't work. Firefox versions 52 through 57 have a scrollIntoView method, it just doesn't work correctly. It will throw an error for inputs of "nearest" or "center" to the block property:

`TypeError: 'block' member of ScrollIntoViewOptions 'center' is not a valid value for enumeration ScrollLogicalPosition.`

patorjk:
Also, I'm totally open to changing this to work without the try/catch block, I'm just not sure how to detect the problem with "center" and "nearest" not being apart of ScrollLogicalPosition. I tried to see if I could find the enumeration via code, but wasn't able to. Another possible fix could be checking if the browser is Firefox and the version is less than 58, since this problem appears to be scoped to that Browser incorrectly implementing scrollIntoView.

dmtrKovalenko:
Closed in favor of #1393

patorjk:
While I’m happy you’re moving forward with a fix, it looks like the only difference between that PR and this one is that you’ve removed what happens in the catch statement. That means FF users 50-57 will see 1800 as their date to select and have to manually scroll to the middle to select their date.

PR #1384, opened 2019-11-07
Add Notification interface

### LINKED ISSUE #1383: CircleCI check fails
The problem is the last commit on master: https://github.com/mui-org/material-ui-pickers/commit/e76f850b82e052a20dd7cd59d71912fb62fc6f41

As the message is deleted, TypeScript can't infer the structure of the messages. Thus, the imported json array is of type `never[]`.

Thus, `yarn workspaces run build:typescript` fails, as can be seen [here](https://circleci.com/gh/mui-org/material-ui-pickers/3093).

I think a notification type should be declared and the imported json data should be casted with `as Notification`.

LorenzHenk:
This PR closes #1383  

## Description

Define TypeScript interface for notification.

### DIFF
--- a/docs/utils/NotificationManager.tsx
+++ b/docs/utils/NotificationManager.tsx
@@ -4,6 +4,11 @@ import notifications from '../notifications.json';
 import { useSnackbar } from 'notistack';
 import { makeStyles } from '@material-ui/core';
 
+interface Notification {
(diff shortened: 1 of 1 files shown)


dmtrKovalenko (approved):
Thanks

PR #1397, opened 2019-11-19
[docs] Update date-fns example

Doesntmeananything:
## Description

I would like to expand date-fns example in the docs with some approaches to customise formatting when date-fns locale object defaults are not appropriate. Since date-fns is very often used for localisation, I believe this is a natural place for more customisation examples. I have also added a link to format customisation guide to this page.

My friend and I were recently bitten by this. She was implementing a picker in tandem with date-fns, and the months were declined incorrectly for that case. Since the issue concerned a particular locale, googling around didn't really help. I then decided to review the docs again and managed to stumble upon Global format customisation guide after quite a bit of digging.

I have also added grammar and style fixes to some of the pages that I had time to  review.

Thank you very much!

### DIFF
--- a/docs/pages/localization/Date-fns-customized.example.jsx
+++ b/docs/pages/localization/Date-fns-customized.example.jsx
@@ -20,7 +20,7 @@ class RuLocalizedUtils extends DateFnsUtils {
   }
 
   getDatePickerHeaderText(date) {
-    return format(date, 'dd MMMM, EEEE', { locale: this.locale });
+    return format(date, 'dd MMMM', { locale: this.locale });
   }
 }
 
@@ -39,7 +39,7 @@ const localeUtilsMap = {
 const localeFormatMap = {
   en: 'MMMM d, yyyy',
   fr: 'd MMM yyyy',
-  ru: 'd MMMM yyyy',
+  ru: 'd MMM yyyy',
 };
 
 const localeCancelLabelMap = {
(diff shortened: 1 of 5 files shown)


dmtrKovalenko (changes requested):
Overall thank you, this is a really good addition. 
But, please create a new section with heavily customized example. It will make easier to understand what is required

Doesntmeananything:
Something like this? I've added a new section to the date-fns page and moved customised example to a separate file.

docs/pages/localization/Date-fns-customized.example.jsx
```diff
@@ -0,0 +1,106 @@
...
+const localeUtilsMap = {
+  en: DateFnsUtils,
+  fr: FrLocalizedUtils,
+  ru: RuLocalizedUtils,
+};
+
+const localeFormatMap = {
+  en: 'MMMM d, yyyy',
+  fr: 'd MMM yyyy',
+  ru: 'd MMMM yyyy',
```
dmtrKovalenko:
```suggestion
  ru: 'd MMM yyyy',
```

I think it was not expected 
image

dmtrKovalenko (changes requested):

dmtrKovalenko:
I am sorry but I cannot push to your fork. Please commit the suggestion

docs/pages/localization/Date-fns-customized.example.jsx
dmtrKovalenko: BTW I would expect чет, `21 ноября` or just `21 ноября` in the header title as a Russian speaker. But it is just like an enhancement

Doesntmeananything:
No problem! I don't think particular formatting is a big deal here for me, since I'd rather concentrate on showcasing various custmomisation options for the pickers. I wouldn't want people hacking on date-fns before trying these out.

Made both suggested changes.

dmtrKovalenko (approved):
Thanks, it looks awesome right now. And really show that we need more developer friendlier way to change formatting for all of inner date renders.

Issue #1293, opened 2019-08-30
RFC: material-ui-pickers v4


dmtrKovalenko:
## Breaking changes 

We are on the way to v4. It is required to make another major release independently of @material-ui/core team. Here what I am planning to release

### Material design v2 update 
Material design guidelines was updated with new reference for date/time pickers [reference]( https://material.io/components/pickers/#anatomy)

Here is new UI sneak peek
![design update](https://storage.googleapis.com/spec-host/mio-staging%2Fmio-components%2F1566425689922%2Fassets%2F1luixhoFfDSk7DCBpduAhfQ5IP1ojD8DI%2Fpicker-choosing-date-calendar.png)
This looks really strange:
![mobile keyboard input](https://user-images.githubusercontent.com/16926049/63997578-9108e680-cb07-11e9-8800-156d562fd96a.png)

Also as you can see this spec contains a design for the date-range picker. So possible new release should also include #364 

![image](https://user-images.githubusercontent.com/16926049/63998121-13de7100-cb09-11e9-81f0-e381ba8fe10d.png)

P.S. It looks like we need to redesign literally **everything**. So will need as many help as possible

### Improved mask and keyboard mode

Updating to the new rifm version will allow us to resolve #1144 and #1289. And also new release will make possible to make a major update of date-io library with an automatic insert of keyboard-friendly formats 

### Changes for variants 

Right now we have `inline` and `dialog` option. Due to changes in material design spec I propose to have variants `mobile` and `desktop` and accept them automatically, on the flight. 

And also we want to make `inline` or `desktop` variant to be primarily supported instead of the mobile one. 

Any discussion is appreciated :) @oliviertassinari

dmtrKovalenko:
If you do not mind please react to the issue, so I know what do you think :)

oliviertassinari:
@dmtrKovalenko Thank you for including me in the discussion. The new specification is awesome, I'm super excited about it 😍. There is an opportunity to significantly increase the adoption of the package. There is a lot of new material to cover. It might make sense to completely rethink the implementation and API at this point, given how diverse and new the specifications are. I'm not sure how to structure my thoughts on the topic, I'm gonna let it flow as a single flow of dots:
[...]

Issue #1128, opened 2019-06-17
TimePicker Style broken with Custom Theme Spacing


samuelcolburn:
## Environment

| Tech                 | Version |
| -------------------- | ------- |
| @material-ui/pickers |     3.1.1    |
| material-ui          |    4.1.1     |
| React                  |     16.8.6    |
| Browser              |  Chrome Version 74.0.3729.169 (Official Build) (64-bit)  |
| Peer library         |   date-fns 2.0.0-alpha.27  |

## Steps to reproduce

1. Add a custom ThemeProvider to material-ui using `@material-ui/styles`

2. Use a custom spacing value (in example spacing: 2)

3. Use a picker TimePicker or DateTimePicker

## Expected behavior

Clock values are properly positioned and the Clock hand properly highlights Clock values.

## Actual behavior

Clock Values are not spaced properly: they are all shifted up.

Clock hand does not position over the selected Clock value.

## Live example
https://codesandbox.io/s/material-ui-pickers-theme-error-on0tg

## Possible Solution

It looks as though this is caused by the positioning of the Clock elements being hard-coded, while the size of the Clock values is determined by the theme.

It looks like this may be solved fairly easily by changing ClockNumber.tsx to only calculate the width based on the theme spacing and leaving the height as a static `32px`. However, I didn't run any tests to see if this would break any other features, or if there was a more elegant solution.

```typescript
// ClockNumber.tsx
export const useStyles = makeStyles(
  theme => {
    // const size = theme.spacing(4);
   const width = theme.spacing(4);
    return {
      clockNumber: {
        // width: size,
       // height: size,
       width,
       height: '32px'
      }
);
```

PR #1416, opened 2019-12-05
[vNext] Redesing to material-ui 2.0 

dmtrKovalenko:
This PR is a part of v4 development (#1293)

## Description

- [x] Common design changes like toolbar, dimensions, paddings, etc...
- [x] Redesign datepicker with toolbar
- [x] Change moving views logic to be inside the calendar and change year with triangle icon
- [x] Fix tests <-- In progress right now, moving our tests from enzyme to cypress components tests
- [x] Redesign timepicker
- [x] Redesign datetime picker 

Note that new layout for mobile/desktop variant will be in the next PR, here just changing everything inside the components.

> This PR hardly in development right now, but please @oliviertassinari take a brief look of the deployment preview for how it looks and feels


lib/src/DatePicker/DatePickerToolbar.tsx
```diff
@@ -1,23 +1,28 @@
...
+export const useStyles = makeStyles<Theme, { isLandscape: boolean }>(
   {
     toolbar: {
+      color: 'white',
       flexDirection: 'column',
       alignItems: 'flex-start',
+      justifyContent: 'space-between',
+      paddingTop: 16,
+      paddingBottom: 16,
+      padding: props => (props.isLandscape ? 16 : undefined),
```
oliviertassinari: I would avoid using style function in any of the components, it's not ready for prime-time.

lib/src/DatePicker/DatePickerToolbar.tsx
```diff
@@ -26,50 +31,51 @@ export const useStyles = makeStyles(
...
-      <ToolbarButton
-        variant={isYearOnly ? 'h3' : 'subtitle1'}
-        onClick={() => setOpenView('year')}
-        selected={openView === 'year'}
-        label={utils.getYearText(date)}
+    <PickerToolbar isLandscape={isLandscape} className={classes.toolbar}>
+      <Typography
+        color="inherit"
+        variant="overline"
+        children={title}
```
oliviertassinari: A funny coding style, was it on purpose 🙃?

lib/src/_shared/icons/PenIcon.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import SvgIcon, { SvgIconProps } from '@material-ui/core/SvgIcon';
+
+export const PenIcon: React.FC<SvgIconProps> = props => {
```
oliviertassinari: You might want to look at the `createSvgIcon()` approach we have. It can be useful. For instance, to forward ref.

lib/src/views/Calendar/Calendar.tsx
```diff
@@ -356,9 +71,152 @@ export const styles = (theme: Theme) => ({
...
+  const moveToDay = React.useCallback(
+    (day: MaterialUiPickersDate) => {
+      if (day && !shouldDisableDate(day)) {
+        handleDaySelect(day, false);
+      }
+    },
+    [handleDaySelect, shouldDisableDate]
+  );
+
+  useKeyDown(Boolean(allowKeyboardControl && variant !== 'static'), {
```
oliviertassinari: There is some issues with this useKeyDown helper:
- It doesn't use the synthetic event, so he can't use `event.key` (React polyfills it)
- It doesn't listen to the active frame, I believe it wouldn't work when rendered in an iframe, popup

lib/src/views/Calendar/CalendarHeader.tsx
```diff
@@ -1,124 +1,178 @@
...
+export const useStyles = makeStyles<Theme, Pick<CalendarWithHeaderProps, 'view'>>(
   theme => ({
     switchHeader: {
       display: 'flex',
-      justifyContent: 'space-between',
       alignItems: 'center',
-      marginTop: theme.spacing(0.5),
+      marginTop: theme.spacing(2),
       marginBottom: theme.spacing(1),
+      paddingLeft: theme.spacing(3),
```
oliviertassinari: I would suggest not to use the spacing helper in these components unless if somebody customizes it, it has an expected outcome.

lib/src/views/Calendar/CalendarHeader.tsx
```diff
@@ -127,15 +181,11 @@ CalendarHeader.displayName = 'CalendarHeader';
 CalendarHeader.propTypes = {
   leftArrowIcon: PropTypes.node,
   rightArrowIcon: PropTypes.node,
-  disablePrevMonth: PropTypes.bool,
-  disableNextMonth: PropTypes.bool,
 };
 
 CalendarHeader.defaultProps = {
```
oliviertassinari, line 228: Have you considered to use the prop spread default values instead of default props?

lib/src/views/Calendar/CalendarView.tsx
```diff
@@ -0,0 +1,198 @@
...
+export const CalendarView: React.FC<CalendarViewProps> = ({
+  date,
+  view,
+  onChange,
+  changeView,
+  onMonthChange,
+  minDate: unparsedMinDate,
+  maxDate: unparsedMaxDate,
+  loadingIndicator = <CircularProgress />,
+  ...rest
```
oliviertassinari: For consistency with the core?

```suggestion
  ...other
```

lib/src/DatePicker/DatePickerToolbar.tsx
dmtrKovalenko: It allows to get rid of unnecessary spaces when passing down children props.

lib/src/_shared/icons/PenIcon.tsx
dmtrKovalenko: I do think about using @material-ui/icons again. I removed that dependency because there were a lot of problems with esm build. 
But AFAI everything should be ok right now? cc @TrySound

lib/src/views/Calendar/CalendarHeader.tsx
dmtrKovalenko: It is primarily used for typescript typings. This is just so old code, will move it. (But there is a problem with auto generating default props values)

lib/src/DatePicker/DatePickerToolbar.tsx
dmtrKovalenko: :( I was happy that it will be possible to get rid of clsx 😢. Do you know when it will be ok? Because docs are saying that everything is stable enough.

lib/src/_shared/icons/PenIcon.tsx
TrySound: @oliviertassinari What is the approach you are talking about?

TrySound: Got it. It's about avoiding boilerplate. Looks good to me.
https://github.com/mui-org/material-ui/blob/89687f38cae750650555772ba4d821c9084d8dfc/packages/material-ui/src/internal/svg-icons/ArrowDropDown.js#L7

lib/src/DatePicker/DatePickerToolbar.tsx
oliviertassinari: I wish we were able to use them in the core components too. We don't as we are speaking. I think that it would either require the styled-components migration or to use react-jss internally. Performance is OK but not great. It's fine for the Box, but I wouldn't use them as a systemic solution.

lib/src/DatePicker/DatePickerToolbar.tsx
oliviertassinari: So it's to optimize for readability when developing?

lib/src/_shared/icons/PenIcon.tsx
oliviertassinari: I don't think that the picker should be importing from @material-ui/icons either. It's too big of a dependency, especially for people that would want to customize the picker. We don't use the icons package either in the @material-ui/core package.

lib/src/views/Calendar/CalendarHeader.tsx
oliviertassinari: We might have a solution for this. The default props generation for API purpose was solve in the mui-org/material-ui repository. @eps1lon took care of it.
It's a good reminder that we need to look into how we can extract the API from TypeScript sources, and not only JavaScript ones.

lib/src/DatePicker/DatePickerToolbar.tsx
dmtrKovalenko: Both. But actually this 
```jsx
<div> {children} </div>
```

Will render 2 spaces before and after the children. And also our behaviour looks more like spreading things down instead of rendering.

lib/src/views/Calendar/CalendarHeader.tsx
dmtrKovalenko: Right now I created custom script that reads jsdoc tags on props like `@default "someValue"` and primarily use it and getting rid from default props.

eps1lon: Our handler for `react-docgen` and default props via default values in object rest spread can be found in https://github.com/mui-org/material-ui/blob/6f7318cdd5c1c2887860c0d56212a3e42d107082/docs/src/modules/utils/defaultPropsHandler.js#L8

dmtrKovalenko: @eps1lon don't you use `react-docgen-typescript`?

lib/src/views/Calendar/CalendarHeader.tsx
dmtrKovalenko: @oliviertassinari it is a nice point. I am really not clear about what spacing means. But we have some issues like #1141. How do you think should we support them or just ignore?

oliviertassinari: Yes, #1128 is exactly why I would discourage using the spacing helper unless a different spacing has a clear expected outcome (we have very few cases on this on the main repository). I can give you one example `<Grid spacing={1} />`. I doubt there are many cases like this in the date picker.

dmtrKovalenko:
@oliviertassinari could you take a look to deployment preview and visual changes? There are quite a lot of code changes :)

oliviertassinari:
@dmtrKovalenko Yes, I'm having a look! Well, I will try 🙃 

Capture d’écran 2019-12-20 à 00 50 13

docs/pages/demo/datepicker/ServerRequest.example.jsx
```diff
@@ -26,7 +26,7 @@ function ServerRequest() {
       <DatePicker
         label="With server data"
         value={selectedDate}
-        onChange={handleDateChange}
+        onChange={date => handleDateChange(date)}
```
oliviertassinari, line 29: I'm wondering if we shouldn't make it clear that we don't expect a returned value.

```suggestion
        onChange={date => {
          handleDateChange(date)
        }}
```

docs/pages/guides/CssOverrides.example.jsx
```diff
@@ -2,8 +2,7 @@
 import React, { useState } from 'react';
 import lightBlue from '@material-ui/core/colors/lightBlue';
 import { DatePicker } from '@material-ui/pickers';
-import { createMuiTheme } from '@material-ui/core';
-import { ThemeProvider } from '@material-ui/styles';
+import { createMuiTheme, MuiThemeProvider } from '@material-ui/core/styles';
```
oliviertassinari: We have removed `MuiThemeProvider` from the documentation.

```suggestion
import { createMuiTheme, ThemeProvider } from '@material-ui/core/styles';
```

docs/pages/guides/CssThemeSpacing.example.jsx
```diff
@@ -1,19 +1,18 @@
 import React, { useState } from 'react';
-import { createMuiTheme } from '@material-ui/core';
-import { ThemeProvider } from '@material-ui/styles';
 import { DateTimePicker } from '@material-ui/pickers';
+import { createMuiTheme, MuiThemeProvider } from '@material-ui/core';
 
-const defaultMaterialTheme = createMuiTheme({
+const muiTheme = createMuiTheme({
   spacing: 2,
```
oliviertassinari, line 6: Should this be a visual regression test rather than a demo? But I'm not sure a visual test is needed. On the main repository, we had very few bug report / concern on this end. In https://material-ui.com/customization/density/#explore-theme-density, people can tweak the density.

lib/src/DatePicker/DatePickerToolbar.tsx
```diff
@@ -26,50 +23,32 @@ export const useStyles = makeStyles(
 export const DatePickerToolbar: React.FC<ToolbarComponentProps> = ({
   date,
   views,
-  setOpenView,
   isLandscape,
-  openView,
+  title = 'SELECT DATE',
```
oliviertassinari, line 27: This would make a good new addition to https://material-ui.com/guides/localization/.
What about using lower case, so that people can more easily tweak the capitalization with CSS. For instance, they might have globally disabled the uppercase of the button.

```suggestion
  title = 'Select date',
```

lib/src/DateTimePicker/DateTimePicker.tsx
```diff
@@ -14,9 +14,9 @@ import {
   WithPureInputProps,
 } from '../Picker/makePickerWithState';
 
-export type DateTimePickerView = 'year' | 'date' | 'month' | 'hours' | 'minutes';
+export type DateTimePickerView = 'year' | 'date' | 'month' | 'hours' | 'minutes' | 'seconds';
```
oliviertassinari, line 17: 😍

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
```diff
@@ -33,81 +45,55 @@ export const DateTimePickerToolbar: React.FC<ToolbarComponentProps> = ({
   hideTabs,
   dateRangeIcon,
   timeIcon,
-  onChange,
+  title = 'SELECT DATE & TIME',
```
oliviertassinari, line 48:
```suggestion
  title = 'Select date & time',
```

I'm wondering if we should rename the prop to `titleText` to something that starts to look like a convention:
- backIconButtonText
- labelRowsPerPage
- labelDisplayedRows
- nextIconButtonText
- getLabelText
- clearText
- closeText
- loadingText
- noOptionsText
- openText

oliviertassinari: We also need to be careful when using DOM attribute names.

lib/src/TimePicker/TimePickerToolbar.tsx
```diff
@@ -139,33 +133,34 @@ export const TimePickerToolbar: React.FC<ToolbarComponentProps> = ({
...
-      {ampm && (
+      {showAmPmControl && (
         <div
           className={clsx(classes.ampmSelection, {
             [classes.ampmLandscape]: isLandscape,
-            [classes.ampmSelectionWithSeconds]: arrayIncludes(views, 'seconds'),
           })}
         >
           <ToolbarButton
+            data-test-id="toolbar-am-btn"
```
oliviertassinari: For reference, we have a `data-mui-test` convention in the main repository to solve this problem (+ a babel plugin to remove it before releasing). It would be great to unify the approach.

@eps1lon should we rename `data-mui-test` to `data-testid`, to match @testing-library default behavior?
```suggestion
            data-testid="toolbar-am-btn"
```

lib/src/_shared/ModalDialog.tsx
```diff
@@ -38,76 +63,56 @@ export const ModalDialog: React.SFC<ModalDialogProps & WithStyles<typeof styles>
...
-    {...other}
-  >
-    <DialogContent children={children} className={classes.dialog} />
-
-    <DialogActions
+}) => {
+  const classes = useStyles();
+  return (
+    <Dialog
+      role="dialog"
```
oliviertassinari: Is the role needed?

lib/src/_shared/ModalDialog.tsx
```diff
@@ -38,76 +63,56 @@ export const ModalDialog: React.SFC<ModalDialogProps & WithStyles<typeof styles>
...
+      <DialogContent children={children} className={classes.dialog} />
 
-      {showTodayButton && (
-        <Button color="primary" onClick={onSetToday}>
-          {todayLabel}
-        </Button>
-      )}
+      <DialogActions
+        classes={{
+          root: clsx({
```
oliviertassinari: Prefer className when possible :)

lib/src/wrappers/ModalWrapper.tsx
```diff
@@ -65,7 +65,7 @@ export const ModalWrapper: React.FC<ModalWrapperProps<any>> = ({
   ...other
 }) => {
   useKeyDown(open, {
```
oliviertassinari: I'm wondering. Why not using a onKeydown listener on the element?

lib/src/views/Clock/Clock.tsx
```diff
@@ -167,8 +60,163 @@ export const styles = (theme: Theme) =>
...
+
+  const handleMouseMove = (e: React.MouseEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
+    // MouseEvent.which is deprecated, but MouseEvent.buttons is not supported in Safari
+    const isButtonPressed =
+      typeof e.buttons === 'undefined' ? e.nativeEvent.which === 1 : e.buttons === 1;
+
+    if (isButtonPressed) {
+      setTime(e.nativeEvent, false);
```
oliviertassinari, line 143: It's the first time I see the usage of `nativeEvent`. Do you have a specific use case over using `event`?

lib/src/views/Clock/Clock.tsx
```diff
@@ -167,8 +60,163 @@ export const styles = (theme: Theme) =>
...
+      isMoving.current = false;
+    }
+  };
+
+  const handleMouseMove = (e: React.MouseEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
+    // MouseEvent.which is deprecated, but MouseEvent.buttons is not supported in Safari
+    const isButtonPressed =
+      typeof e.buttons === 'undefined' ? e.nativeEvent.which === 1 : e.buttons === 1;
```
oliviertassinari, line 140: I believe `event.button === 0` could work cross platform.

lib/src/views/Clock/Clock.tsx
```diff
@@ -167,8 +60,163 @@ export const styles = (theme: Theme) =>
       left: '50%',
       transform: 'translate(-50%, -50%)',
     },
-  });
+    amButton: {
+      zIndex: 24,
```
oliviertassinari: Does 24 answer to a specific logic? Do you need the property in the first place?

lib/src/_shared/ModalDialog.tsx
```diff
@@ -38,76 +63,56 @@ export const ModalDialog: React.SFC<ModalDialogProps & WithStyles<typeof styles>
...
 
-      {okLabel && (
-        <Button color="primary" onClick={onAccept}>
-          {okLabel}
-        </Button>
-      )}
-    </DialogActions>
-  </Dialog>
-);
+        {cancelLabel && (
```
oliviertassinari, line 97: I'm wondering about a new name for the localization wordings (for consistency, as detailed above).

```suggestion
        {cancelText && (
```

lib/src/_shared/ToolbarButton.tsx
```diff
@@ -1,24 +1,31 @@
...
+export interface ToolbarButtonProps extends ExtendMui<ButtonProps, 'variant'> {
   variant: TypographyProps['variant'];
   selected: boolean;
   label: string;
   align?: TypographyProps['align'];
   typographyClassName?: string;
 }
 
+export const useStyles = makeStyles(
+  _ => ({
```
oliviertassinari: No theme?

```suggestion
  {
```

lib/src/views/Calendar/DayWrapper.tsx
```diff
@@ -23,10 +23,9 @@ const DayWrapper: React.FC<DayWrapperProps> = ({
       role="presentation"
       onClick={dayInCurrentMonth && !disabled ? handleClick : undefined}
       onKeyPress={dayInCurrentMonth && !disabled ? handleClick : undefined}
+      children={children}
```
oliviertassinari, line 26: For the convention in the React community and for people reading the source, I think that the previous approach would be less destabilizing.

lib/src/views/Calendar/FadeTransitionGroup.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+
+interface FadeTransitionProps {
+  transKey: React.Key;
+  className?: string;
+  children: React.ReactChild;
+}
+
+const animationDuration = 500;
```
oliviertassinari, line 13: Should we use a transition duration from on the of the specification (theme.transitions.duration.x)?

oliviertassinari (commented):
This is a first batch of review, I have only looked at the source code. I cover some generic questions, and details.

Tomorrow, I will take more time to look at the product DX and UX, I think that I will be able to do more important feedback with this second phase :).

Well done so far 👌

lib/src/views/Clock/Clock.tsx
dmtrKovalenko: We need e.buttons === 1

docs/pages/demo/datepicker/ServerRequest.example.jsx
dmtrKovalenko: I dont think it will help somebody. I think 99% of js developers will be confused by such examples

docs/pages/guides/CssOverrides.example.jsx
dmtrKovalenko: oh god :)

docs/pages/guides/CssThemeSpacing.example.jsx
dmtrKovalenko: Actually there were some issues with it in pickers #1128. So this is more like a proof that we are not completely broken with custom spacing. But I agree that it needs to be a test rather than a demo

lib/src/TimePicker/TimePickerToolbar.tsx
dmtrKovalenko: Actually I don't like @testing-library so now working on improving cypress with component tests https://github.com/cypress-io/cypress/pull/5923. 
Going to try use it with pickers.
> P.S `data-testid` is incorrect word for spell checkers, but I don't care about it

lib/src/_shared/ModalDialog.tsx
dmtrKovalenko: Haha, this line was added more than 2 years ago, sure it is not needed

lib/src/views/Clock/Clock.tsx
dmtrKovalenko: Yes, react is not including some props from event like `changedTouches` 
```
offsetX = e.changedTouches[0].clientX - rect.left;
offsetY = e.changedTouches[0].clientY - rect.top;
```

same for `which`, which (🤣) is working only in safari

lib/src/_shared/ModalDialog.tsx
dmtrKovalenko: Let's pull this stuff out to separate PR with `localization support`. This one is so giant. I have a lot of thoughts about it, like providing localized formats for dates out of the box with localization. It will be a killer feature

lib/src/_shared/ToolbarButton.tsx
dmtrKovalenko: Agree. 
But don't like such approach. Let me explain. From the perspective of changing styles from using theme or not. To change it you need 1) remove function 2) remove the `)` from the end of expression. 
I would like to use only 1 way to create theme. 
> Also for reference from the algebraic type system perspective we should avoid such polymorphic functions, cause they are giving more friction.

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
dmtrKovalenko: I thought this stuff should be filled from `label` coming to input. And then be overridable with custom prop

lib/src/views/Calendar/FadeTransitionGroup.tsx
dmtrKovalenko:
```
shortest: 150
shorter: 200
short: 250
standard: 300
complex: 375
enteringScreen: 225
leavingScreen: 195
```

All these durations are too fast for this animation.

lib/src/TimePicker/TimePickerToolbar.tsx
oliviertassinari: So, what do you think that we should do here? `data-testid` seems to match https://docs.cypress.io/guides/references/best-practices.html.

lib/src/views/Clock/Clock.tsx
dmtrKovalenko: Thanks its debugging stuff

lib/src/views/Clock/Clock.tsx
oliviertassinari: I don't have the whole context, it might be something to dig into. It doesn't seem to be an issue with the `Slider` component that uses `changedTouches` too.

lib/src/views/Clock/Clock.tsx
oliviertassinari: I'm wondering because react-router checks `event.button` to only consider the main click (0).

lib/src/_shared/ModalDialog.tsx
oliviertassinari: Yes, I agree!

lib/src/_shared/ToolbarButton.tsx
oliviertassinari: Remembers me when we were switching between class components and functions components, eww, it so much better with hooks now.

lib/src/TimePicker/TimePickerToolbar.tsx
dmtrKovalenko: Hehe, my bad :DDDD will rename it

lib/src/views/Clock/Clock.tsx
dmtrKovalenko: Just checked with pickers, it doesn't works. BTW we are using also 
```
const rect = e.target.getBoundingClientRect();
```
which is more likely to be a problem. I don't remember why it was added. But it is required (it is crashing with event)

oliviertassinari: Should it be `event.currentTarget.getBoundingClientRect()` here?

dmtrKovalenko: image

`e.changedTouches` is undefined for synthetic event

oliviertassinari (commented):
A second path, only looking at the end result. I assume the objective is to match the latest specification regarding the experience on mobile. I have noticed the following opportunities:

- When switching to the year view, the scroll starts at the top and animates down to the active year. 
I think that starting at the active year scroll position would improve the UX. It's the behavior that material.io demonstrates, that I can see on Google Calendar and Gmail (Android).
- When picking a year, like 1900, I think that it would be better to slide the year window around it, so we can choose between 1800 and 2000. It's not ideal to pick a date around 1500, hopefully, the windowing allows moving 200 years at the time is good enough for a marginal problem (very rare).
- Should a tap on the year display the year view? I'm not sure that the dropdown icon surface nor discoverability is good enough.
- It seems that the year chip should be 36px height (OK) and have around borders, what about increasing the border-radius from 16px to 18px?
- The month transition animation on Android is KO. I suspect the view is too heavy. What about removing the ButtonBase from it? I think that it could also greatly help when rendering the month view independently (or when we will need to render 3 months on the same view). Alternatively, we could remove the animation, it's would be fine. Right now, it best benefits iOS users while it harms Android (and harms desktop, but we mentioned that it's a topic for another effort) users.

I haven't noticed anything else, it's so cool to see it making progress :). I haven't looked at the API, but I guess it's too early for that?

lib/src/views/Clock/Clock.tsx
oliviertassinari: For a mouse event, this sounds correct (mouse up). Should we use clientX and clientY for mouse events?

dmtrKovalenko:
@oliviertassinari 
1. There is a specific prop called `animateYearScrolling` which is by default `false`. It is confusing that it is enabled for the first example, I will remove it. By default it is starting from the year position, but can be usefull for something. 
2. What do you mean? Increse min-max range to include 200 years back and forwards? 
3. Nice questions. I think yes for desktop users click on month should open `MonthView`, click on year open year, but I am not sure about mobile, maybe just tapping on the header should also move to year?
( I have created this functionality on the very first try, but for some reason removed it) 
4. will fix 
5. I propose to create new prop called `reduceAnimations=/(android)/i.test(navigator.userAgent)`. This will actually improve default user experience, and will show animations for desktop and ios. What do you think?

Api changes will appear in the next PR. Here we just moving to new design reference

lib/src/views/Clock/Clock.tsx
dmtrKovalenko: Yes, because we need to get mouse position relative to the container for dragging. `offsetX` and `offsetY` will not work for all browsers

oliviertassinari:
1. I didn't know about the `animateYearScrolling` prop. What problem does the prop solve? Do we have an opportunity to reduce the complexity of the component?
2. I was wondering about using the selected year, and not the current year (2019, soon 2020) to generate the list of years people can click on.
3. I have noticed that the date & time picker has an interactive year. I don't have any solution in mind. We might get an issue about this problem in the future. We will see.
5. I like the solution you have proposed. I also think that we should seriously consider giving up on the ButtonBase usage in the day cells. We have to balance between these dimensions:
- can we render 3 months in a reasonable amount of time, on a 4 years old desktop? For instance, I have a 5 years old Macbook Air 🙈. I would have concerns if it takes more than 250ms, for instance, to shift one month.
- can we have instant feedback on keyboard navigation, right now, it feels low. I suspect we won't have any other choice than to render outside React. Try keeping the right arrow down, the UI freeze.
- can we have a smooth animation? Taking into account that no animation might be better than a freezing one.

dmtrKovalenko:
1. Yeah I think we can remove this prop. It looks useless.
2. Related to the years. They are generating from the `minDate` to `maxDate`. By default the range is `1900-01-01` to `2100-01-01`. 
3. Yes, it looks like a problem, but still, you can switch to year view from toolbar and also (needs to be implemented) from calendar header 
4. Will try to use `ButtonBase` and check is it reduces performance 

(haha I am seeing that keyboard navigation is stopping after the animation, I suppose the problem is losing focus. Looks like we need a global useKeyDown)

oliviertassinari:
2. Thanks for the context. I thought the date picker had no limits. What happens if we change the props to remove the min and max date limits? I have tried `<DatePicker minDate={null} maxDate={null}` but it didn't seem to have an impact. https://blueprintjs.com/docs/#datetime/datepicker behaves like the current implementation. (Oh, I see that blueprints went with http://react-day-picker.js.org/, it sounds like its a good signal that we can benchmark with it :)).
What do you think of removing the default limits, and having ±100 years sliding window around the "active" date?

Unrelated, we might have an opportunity to improve a11y. I was wondering about the usage of grid roles for the day grid: https://www.w3.org/TR/wai-aria-practices/#table.

4. I had a closer look at the "calendar" implementation. Do we need the `<div role="presentation">` around the cells? A lighter DOM node could help with performance.

dmtrKovalenko:
Addressed some feedback:

1) Moving to button base saved 2ms of rendering on my machine (from 10ms to 8ms of rendering dialog with transition group)
2) My concern about the dynamic range by default - will it be obvious for developers? Right now it is clear that `minDate` will not render anything below. Moreover, it will require additional calculations on the rendering step (honestly it should not be a problem) 
3) Yes, we need this wrapper with `role="presentation"`. The main reason is that the day component is overridable, so we need a way to be sure that all event handlers are attached even when the component is overridden. 
4) Keyboard animation was broken by moving to the relative keydown listeners. Right now fixed and will not freeze after switching to the next month. 
5) https://axesslab.com/accessible-datepickers/ I do think that primary way of input day should be a keyboard. I suppose that's why google added an input button even into the mobile datepicker view. I am not sure how users can properly understand the selected date from the calendar day list. I am planning also to significantly improve accessibility by adding proper aria-labels to the date values and properly handle keyboard input. But also it needs to be done after keyboard input from mobile and to be a separate PR

oliviertassinari:
1. I have tried the following diff, to benchmark a month switch in dev mode

```diff
diff --git a/lib/src/views/Calendar/Day.tsx b/lib/src/views/Calendar/Day.tsx
index 100cad6..d28a1e8 100644
--- a/lib/src/views/Calendar/Day.tsx
+++ b/lib/src/views/Calendar/Day.tsx
@@ -70,9 +70,9 @@ export const Day: React.FC<DayProps> = ({
   });

   return (
-    <IconButton className={className} tabIndex={hidden || disabled ? -1 : 0} {...other}>
-      <Typography variant="body2" color="inherit" children={children} />
-    </IconButton>
+    <span>
+      {children}
+    </span>
   );
 };
```

**before**
Capture d’écran 2019-12-22 à 18 56 49

**after**
Capture d’écran 2019-12-22 à 18 56 41
2. My main concern would be about: if somebody needs to significantly increase the min/max range. How long does it take to render the whole year list (>1s?) and how much to scroll to find the value the user needs?
3. For this problem, we frequently ask for developers to spread the props to the root element, it could be an alternative. I have tried without the wrapping div, it doesn't seem to have a significant impact as we have in 1. It seems that we can ignore performance on the mater. 
4. Nice
5. +1 for keyboard and a11y exploration

dmtrKovalenko:
1. I am not seeing such significant performance impact with `ButtonBase` + `<span>` instead of IconButton + Typography. But still removing it in favor of simplicity and fewer dom nodes 
2. Here is a snapshot switching to year selection with rendering a list from 1700 to 2300. Pretty fast, and thanks to animation it is even more smooth. 
image

Also, I am planning to render 4 years in the column for desktop pickers. For mobile it is too much cause touch targets will be too small. 
> From [material.io](https://material.io/components/pickers/#mobile-pickers): Mobile calendar pickers. They aren’t ideal for selecting dates in the distant past or future that require more navigation, such as entering a birth date or expiration date.
3. Haha, I remember why we need an additional day wrapper :)
```jsx
<DatePicker
  // We are passing already rendered day node right to the override function. Super usefull for adjustign rendering
  renderDay={(day, selectedDate, isInCurrentMonth, dayComponent) => {
     const isSelected = isInCurrentMonth && selectedDays.includes(date.getDate());

     return <Badge badgeContent={isSelected ? '🌚' : undefined}>{dayComponent}</Badge>;
   }}
 />
```

5. Will move this work to separate PR

dmtrKovalenko:
I think this should be ready to go? Moving to some work in date-io 💪

oliviertassinari (approved):
One step at the time 👍

PR #1433, opened 2019-12-25
Mobile & desktop components

dmtrKovalenko:
A part of vNext #1293


docs/pages/demo/timepicker/KeyboardTimePicker.example.jsx
```diff
@@ -1,11 +1,11 @@
 import React, { useState } from 'react';
-import { KeyboardTimePicker } from '@material-ui/pickers';
+import { TimePicker } from '@material-ui/pickers';
```
oliviertassinari, line 2: I'm wondering, should we document this approach instead? 

```suggestion
import TimePicker from '@material-ui/pickers/TimePicker';
```

At the same time, I was wondering about changing the demos of http://material-ui.com/, to use the Pull Request's approach. I'm curious to hear your thoughts on this.

lib/src/Picker/Picker.tsx
```diff
@@ -134,8 +160,6 @@ export const Picker: React.FunctionComponent<PickerProps> = ({
       </div>
     </div>
   );
-};
+}
 
-Picker.defaultProps = {
-  ...datePickerDefaultProps,
-};
+Picker.displayName = 'Picker';
```
oliviertassinari, line 185: What's the motivation for setting a display name? The current tradeoff we use in the main repository is to ensure a correct display name in development and to optimize for bundle size in production (a marginal gain).

lib/src/wrappers/WrapperVariantContext.tsx
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+import { WrapperVariant } from './Wrapper';
+
+// consider getting rid from wrapper variant
```
oliviertassinari, line 4: What alternative do you have in mind?

lib/src/wrappers/ResponsiveWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+  cancelLabel,
+  clearLabel,
+  todayLabel,
+  showTodayButton,
+  clearable,
+  DialogProps,
+  PopoverProps,
+  ...other
+}) => {
+  const isDesktop = useMediaQuery<Theme>(theme => theme.breakpoints.up(desktopModeBreakpoint));
```
oliviertassinari, line 26: We are lucky date pickers, on the server, aren't relevant :). Maybe we could add a logic to render null on the server, to clear up any potential confusion (I'm wondering) and save CPU cycles (I'm not too sure about this one because if it then render a Modal, it will be empty)?

lib/src/views/Year/Year.tsx
```diff
@@ -0,0 +1,77 @@
...
+      })}
+      {...other}
+    />
+  );
+};
+
+Year.displayName = 'Year';
+
+export default React.forwardRef<HTMLDivElement, YearProps>((props, ref) => (
+  <Year {...props} forwardedRef={ref} />
```
oliviertassinari: We can save the intermediary component, unless there is some TypeScript issue I'm not aware of.

lib/src/views/Year/Year.tsx
```diff
@@ -0,0 +1,77 @@
...
+      className={clsx(classes.root, {
+        [classes.yearSelected]: selected,
+        [classes.yearDisabled]: disabled,
+      })}
+      {...other}
+    />
+  );
+};
+
+Year.displayName = 'Year';
```
oliviertassinari: We use the following in the main repository to avoid the need for setting a display name.

```js
const Year = React.forwardRef(function Year(props, ref) {
```

lib/src/views/Year/Year.tsx
```diff
@@ -0,0 +1,77 @@
...
+  onSelect,
+  forwardedRef,
+  value,
+  selected,
+  disabled,
+  children,
+  ...other
+}) => {
+  const classes = useStyles();
+  const handleClick = React.useCallback(() => onSelect(value), [onSelect, value]);
```
oliviertassinari: What's the use case for useCallback?

lib/src/views/Year/Year.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+    yearDisabled: {
+      pointerEvents: 'none',
+      color: theme.palette.text.hint,
+    },
+  }),
+  { name: 'MuiPickersYear' }
+);
+
+export const Year: React.FC<YearProps> = ({
```
oliviertassinari: Should it match the name of the style sheet? (MuiPickersYear -> PickersYear)

lib/src/views/Year/Year.tsx
```diff
@@ -0,0 +1,77 @@
...
+        fontWeight: theme.typography.fontWeightMedium,
+      },
+    },
+    yearSelected: {
+      margin: '10px 0',
+      fontWeight: theme.typography.fontWeightMedium,
+    },
+    yearDisabled: {
+      pointerEvents: 'none',
+      color: theme.palette.text.hint,
```
oliviertassinari: Maybe?

```suggestion
      color: theme.palette.text.disabled,
```

lib/src/views/Clock/Clock.tsx
```diff
@@ -27,7 +27,7 @@ export const useStyles = makeStyles(
       justifyContent: 'center',
       alignItems: 'flex-end',
       position: 'relative',
-      margin: `16px 0 8px`,
+      margin: `16px 0 16px`,
```
oliviertassinari:
```suggestion
      margin: '16px 0',
```

lib/src/views/Calendar/Calendar.tsx
```diff
@@ -199,7 +206,7 @@ export const Calendar: React.FC<CalendarProps> = ({
         transKey={currentMonth!.toString()}
         className={classes.transitionContainer}
       >
-        <div>
+        <div style={{ overflow: 'hidden' }}>
```
oliviertassinari, line 209: a stylesheet instead of inline style?

lib/src/typings/helpers.ts
```diff
@@ -1,5 +1,3 @@
-import { Omit } from '../_helpers/utils';
```
oliviertassinari, line 1: I see that the helper was introduced in the TypeScript (v3.5) directly: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html. However, I'm wondering about what version of TypeScript we should support and the timing. cc @eps1lon. (TypeScript doesn't follow semver).

On a side note, we have this helper in `@material-ui/types` too. https://unpkg.com/browse/@material-ui/types@4.1.1/index.d.ts

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -81,34 +84,43 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
...
+      rawValue: value,
       validationError,
       openPicker: () => !readOnly && !disabled && setIsOpen(true),
     }),
-    [disabled, inputValue, readOnly, setIsOpen, validationError]
+    [disabled, format, inputValue, onChange, readOnly, setIsOpen, validationError, value]
   );
 
   const pickerState = { pickerProps, inputProps, wrapperProps };
+  useDebugValue(pickerState, () => ({
```
oliviertassinari, line 128: Interesting, I have almost never used it. I'm wondering, in the dev tools, don't they already scope the state under the name of the hook, do we need the `MuiPickerState` key?

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -62,17 +63,19 @@ export function usePickerState(props: BasePickerProps, options: StateHookOptions
       onClear: () => acceptDate(null),
       onAccept: () => acceptDate(pickerDate),
       onSetToday: () => setPickerDate(utils.date()),
-      onDismiss: () => {
-        setIsOpen(false);
-      },
+      onDismiss: () => setIsOpen(false),
```
oliviertassinari, line 70: Haha, we have two schools of thought. I appreciate the previous approach as when we refactor, it's obvious the method doesn't return anything and we can move on. On the other hand, the new approach is shorter.

oliviertassinari (commented):
I haven't looked at the "essence" of the changes, I struggle a bit to follow it all 🙃. Maybe I should wait for the out of draft state? Or when the demos are updated and running.

Instead, my comments are related to generic patterns I have noticed. I hope it's not too much bikeshedding.

docs/pages/demo/timepicker/KeyboardTimePicker.example.jsx
dmtrKovalenko: We have different build system. esm build allowed only by importing from the root. P.S tree shaking works

oliviertassinari: OK, good to know.

lib/src/Picker/Picker.tsx
dmtrKovalenko: For some reason our components are missing displaynames in a lot of cases. I need to investigate the issue

lib/src/wrappers/WrapperVariantContext.tsx
dmtrKovalenko: Just passing through the props. But now I tent to left a context.

lib/src/typings/helpers.ts
dmtrKovalenko: We are already using `Omit` from typescript in v3. It was accidentally a problem, so in the release notes we asked to update typescript

lib/src/_shared/hooks/usePickerState.ts
dmtrKovalenko: I am always using the shorter way because the code itself in typescript and I always know if method returns `void` or not

dmtrKovalenko:
@oliviertassinari Yeah, please wait. Today I will try to update the documentation and you can compare documentation changes to see `grand` API changes. I`d like to know what do you think about that. 

I`ll ask your review and move this PR from draft. But anyway thanks for some points that I missed :)

lib/src/Picker/Picker.tsx
oliviertassinari: If it's related to forward ref, we have this solution:
```jsx
const Rating = React.forwardRef(function Rating(props, ref) {
```

lib/src/wrappers/WrapperVariantContext.tsx
oliviertassinari: Context sounds great :)

docs/package.json
```diff
@@ -12,19 +12,21 @@
   "keywords": [],
   "author": "",
   "license": "ISC",
+  "engines": {
+    "node": "12.x"
```
oliviertassinari, line 16: I'm wondering if we need to mention the engine. For instance, we don't mention it on the main documentation. But we do for the published packages ("node": ">=8.0.0")

docs/pages/demo/datepicker/AdvancedKeyboard.example.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { Fragment, useState } from 'react';
+import EventNoteIcon from '@material-ui/icons/EventNote';
+import { DesktopDatePicker } from '@material-ui/pickers';
+
+function AdvancedKeyboardExample(props) {
+  const [selectedDate, handleDateChange] = useState(new Date());
+
+  return (
+    <Fragment>
```
oliviertassinari, line 9:
```suggestion
```

docs/pages/demo/datepicker/AdvancedKeyboard.example.jsx
```diff
@@ -0,0 +1,28 @@
...
+        label="Advanced keyboard"
+        placeholder="2018/01/01"
+        format={props.__willBeReplacedGetFormatString({
+          moment: 'YYYY/MM/DD',
+          dateFns: 'yyyy/MM/dd',
+        })}
+        mask="____/__/__"
+        keyboardIcon={<EventNoteIcon />}
+        value={selectedDate}
+        onChange={date => handleDateChange(date)}
```
oliviertassinari, line 22: For consistency, I think that we should always expose the event as the first argument.

```
  /**
   * Callback fired when the value changes.
   *
   * @param {object} event The event source of the callback
   * @param {DateIOType} value
   */
  onChange: PropTypes.func,
```

https://github.com/mui-org/material-ui-pickers/issues/1293#issuecomment-538374383

docs/pages/getting-started/usage.mdx
```diff
@@ -13,6 +13,8 @@ import { CODESANDBOX_EXAMPLE_ID } from '_constants';
 
 #### Quick start
 
+**Note** `onChange` has a second parameter. So if you will not do `onChange={date => handleDateChange(date)}` you will get the console warning that [useState hook does not accept a second callback argument.](https://github.com/facebook/react/issues/14174)
```
oliviertassinari, line 16: I don't understand. What's the purpose of the note?

docs/pages/getting-started/usage.mdx
```diff
@@ -28,9 +30,9 @@ function App() {
 
   return (
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
-      <DatePicker value={selectedDate} onChange={handleDateChange} />
-      <TimePicker value={selectedDate} onChange={handleDateChange} />
-      <DateTimePicker value={selectedDate} onChange={handleDateChange} />
+      <DatePicker value={selectedDate} onChange={date => handleDateChange(date)} />
```
oliviertassinari, line 33: For reference, we use the following approach in the other components:
```suggestion
      <DatePicker value={selectedDate} onChange={(event, date) => { handleDateChange(date) }} />
```
The main incentive for not using implicit returns is to reduce the cost of changes, it avoids having to look at what the logic does (not to break a potentially returned value) before introducing a new line in the handler (I believe that even if using TypeScript, people have to do the effort to check).

lib/.size-snapshot.json
```diff
@@ -1,26 +1,26 @@
 {
   "build/dist/material-ui-pickers.esm.js": {
-    "bundled": 109092,
-    "minified": 62241,
-    "gzipped": 16402,
+    "bundled": 108303,
+    "minified": 61808,
+    "gzipped": 16662,
```
oliviertassinari, line 5: I don't think that this value represents reality. It seems to ignore the cost of the core component dependencies. I strongly think that it should take the cost of @material-ui/core and date-io into account. Otherwise, we can blindly include a large number of dependencies. I suspect the "real" size is more in the order of 80kB (16 pickers + 20 styles + 35 the other components + 10 date-io) for somebody only using the pickers. Which is about x4 more than we can hope for looking at https://bundlephobia.com/result?p=react-datepicker. I suspect we will be able to find leverages to improve this part :).

What about React? Given that people don't have the choice to use it, I think that, this time, we can remove it from the equation.

What about prop-types, I suspect its cost is also not taken into account because it's marked as a peer dependency. Which, I think that we should also fix. Now, the good news is that @eps1lon has built a more advanced bundle size tracking in https://github.com/mui-org/material-ui (we were using size-limit) before that. They we will be able to leverage once we host the pickers in the mono-repository.

oliviertassinari: This also raised another point I don't understand. Why is @date-io/core a peer dependency?
I think that they are only two cases that justify a peer dependency: 1. the dependency has a singleton, 2. the dependency is optional. I believe 1. and 2. don't apply. 

Note that @material-ui/core should be a peer dependency because of 1. (the styles's theme propagation: context).

lib/src/__tests__/test-utils.tsx
```diff
@@ -34,7 +36,11 @@ export const shallow = <P extends WithUtilsProps>(element: React.ReactElement<P>
...
 
 export const mount = <P extends WithUtilsProps>(element: React.ReactElement<P>) =>
-  enzyme.mount(<MuiPickersUtilsProvider utils={UtilClassToUse}>{element}</MuiPickersUtilsProvider>);
+  enzyme.mount(
+    <ThemeProvider theme={createMuiTheme()}>
+      <MuiPickersUtilsProvider utils={UtilClassToUse}>{element}</MuiPickersUtilsProvider>
+    </ThemeProvider>
+  );
 
 export const shallowRender = (render: (props: any) => React.ReactElement<any>) => {
```
oliviertassinari, line 45: I think that it would be great to migrate all the shallow tests to e2e tests. I haven't looked at how they are used.

oliviertassinari (commented):
### The small details I have noticed:

- Regarding the codesandbox iframe of the getting started page. It's slow to load. What's the advantage over an embedded demo?
- The "Custom spacing" demo, I think that we could remove it. I believe checking for theme.spacing in the source yields +80% of the value. I doubt we will see a regression anytime soon. I imagine having it has helped to solve the underlying issue. But now? 
- cmd + click on the docs sidenav links doesn't work (open new tab). But it's already handled in http://material-ui.com/, so we should be fine.
- `disabled` seems to be the default prop value (false) in the API docs.
- core and pickers use a different naming convention for the descriptions. It would be great to unify them.
- There is an alignment issue in the /demo/datetime-picker#basic-usage
alignment issue
- Something feels off with the outer spacing: /demo/datepicker#basic-usage
outer-spacing
- In /demo/timepicker#basic-usage, it would be awesome to include a demo based on this approach: https://material-ui.com/components/autocomplete/#disabled-options. 

### The important chunks I have noticed:

- The desktop version blocks the scroll and moves the focus away from the input.
- I think that we should support a `defaultValue` prop. We have recently introduced a new utility to help you do so: [useControlled](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/utils/useControlled.js). Once we do, I don't see any reason for `value` and `onChange` to be required.
- `autoOk`: it seems that the majority of the libraries I can benchmark with have it enabled by default, including the default date picker on chrome. Shouldn't we change the default value? And if we do then, we need to rename the prop: to prefix it with `disable` -> `disableAutoOk` so we can keep false as the defaut value (a naming convention implemented for all the component, and inspired by the HTML 5 spec for attributes). Now, because the Autocomplete has the same feature, I would propose that we name the prop identically (to increase consistency). Hence my proposal for the following in https://github.com/mui-org/material-ui-pickers/issues/1293#issuecomment-538374383, you can find 2 other renaming proposals I would love to get your feedback on (`allowKeyboardControl` and `initialFocusedDate`).
  ```
  /**
   * If `true`, the popup won't close when a value is selected.
   *
   * from (autoOk)
   * default false
   */
  disableCloseOnSelect: PropTypes.bool,
  ```
- They are a couple of props that don't enforce the boolean naming approach (to always have false as a default). I think that it would be great to change them. We have it documented in https://material-ui.com/guides/api/#property-naming.
- I suspect the Autocomplete and the desktop version of the Datepicker have the potential to share a bunch of the same logic. We could even consider making a shared hook, but maybe we should wait for a 3 use case, like a color picker or tree-select.
In any case, the parts I see that should/could have the same answer for: Popper, `disableCloseOnSelect`, `disableOpenOnFocus`, keyboard escape, `autoComplete`, `onInputChange`, how we expose the TextField. The objective is both to increase consistency and improve the UX.
- The keyboard feels slow to move, in dev and with a single month display (I fear what it would be with 2 or 3 months displayed at once). I think that we should try to use a pure DOM-based approach (bypassing React), like http://react-day-picker.js.org/ is doing. https://airbnb.io/react-dates/ has the same issue. It's not great (but I can't find any open issue related to it on react-dates; so I would vote for ignoring the problem, for now, unless it would require an important change of the logic, in which case, as we are diving deep into it, now, might be the best time).
- I strongly think that we should remove `onMonthChange` support for promises. If the user quicky change between months, x requests will be fired, we have no guarantees on the order the network will resolve them (unless I have missed something). We might very well display information for the wrong month.
In practice, removing the support seems to be 90% about updating the demo.
- For the spinning issue, I think that a `loading` prop would be better, enabling the support of more use cases. speaking of the loading state. I think that we could center it (it would solve the scrolling issue).
On the other hand, I think that it would be better to leave the loading implementation with a customization capability. I don't think that it's a frequent use case. Does it justify bundling the circular progress for it?
loading

By the way, you mentioned that you would need 90 hours to make significant progress.
Man, if you want to dedicate more time to it, over another one, it's perfectly fine. This is the type of component that people tend not to write themselves and is very frequently used, so having something very polished can make a huge difference 😍.
I'm wondering if working on different components in parallel, alone, is a better strategy than working on fewer components at the same time, but in small teams of two.

### The initiatives I have seen and I have loved

- The responsive approach is cool.
- The second picker support is cool.
- The progress made with the desktop support

docs/package.json
dmtrKovalenko: It is `docs` package.json. It is required for zeit now deployments

docs/pages/demo/datepicker/AdvancedKeyboard.example.jsx
dmtrKovalenko: We cannot use event in `onChange` because this feature can be fired from
1) keyboard input (we have event)
2) clicking `ok` button (we have event, but it is `MouseEvent` so I think we already cannot pass 2 different event types)
3) pressing enter with open datepicker (again different event type)
4) selecting date in the last available view <-- **Here** we don't have an event 

Again passing event will make more confusion

docs/pages/getting-started/usage.mdx
dmtrKovalenko: There were 5 issues opened because of this warning. I decided to put this as a note

docs/pages/getting-started/usage.mdx
dmtrKovalenko: I don't think it is will help somebody, but I don't care I can change the examples

lib/.size-snapshot.json
dmtrKovalenko: 1. Make sure that we have a timepicker - which is entierly different control and requires (I think more then 20% of bundlesize)
2. `@date-io/core` has no bundle size. We have optional dependency `@date-io/{library}` and @date-io/core is used to orchestrate these dependencies and make sure that some of the library is installed. Users don't need to install it -> it is a dependency of each @date-io/{lib} and has no bundle size

PR #1460, opened 2020-01-19
Fix copy and paste typo: "desktop" instead of "mobile"

Skn0tt:
Thanks for this great package!
See title.

### DIFF
--- a/docs/pages/api/props.tsx
+++ b/docs/pages/api/props.tsx
@@ -76,7 +76,7 @@ const Docs: React.FC<WithRouterProps> = ({ router }) => {
 
           <h4> Desktop Wrapper </h4>
           <Typography gutterBottom>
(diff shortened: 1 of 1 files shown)


dmtrKovalenko (changes requested):
Or with `Mobile{componentName}`

docs/pages/api/props.tsx
```diff
@@ -76,7 +76,7 @@ const Docs: React.FC<WithRouterProps> = ({ router }) => {
 
           <h4> Desktop Wrapper </h4>
           <Typography gutterBottom>
-            Props available on mobile device with `{componentName}` or with `Desktop{componentName}`
+            Props available on desktop device with `{componentName}` or with `Desktop{componentName}`
```
dmtrKovalenko:
```suggestion
            Props available on desktop device with `{componentName}` or with `Mobile{componentName}`
```

PR #1466, opened 2020-01-23
Change the typings of Clock to use a static type as ClassKey

### LINKED ISSUE #1465: Type compatibility issues with material-ui 4.9.0
Due to the changes made in https://github.com/mui-org/material-ui/pull/19320 a type error occurs in pickers

```
node_modules/@material-ui/pickers/views/Clock/Clock.d.ts:4:48 - error TS2344: Type '(theme: Theme) => Record<"container" | "clock" | "squareMask" | "pin", CSSProperties | (() => CSSProperties)>' does not satisfy the constraint 'string | Record<any, CSSProperties | CreateCSSProperties<any> | ((props: any) => CreateCSSProperties<any>)> | StyleRulesCallback<...>'.
  Type '(theme: Theme) => Record<"container" | "clock" | "squareMask" | "pin", CSSProperties | (() => CSSProperties)>' is not assignable to type 'StyleRulesCallback<any, any, any>'.
    Type 'Record<"container" | "clock" | "squareMask" | "pin", CSSProperties | (() => CSSProperties)>' is not assignable to type 'Record<any, CSSProperties | CreateCSSProperties<any> | ((props: any) => CreateCSSProperties<any>)>'.
      Property 'container' is incompatible with index signature.
        Type 'CSSProperties | (() => CSSProperties)' is not assignable to type 'CSSProperties | CreateCSSProperties<any> | ((props: any) => CreateCSSProperties<any>)'.
          Type '() => CSSProperties' is not assignable to type 'CSSProperties | CreateCSSProperties<any> | ((props: any) => CreateCSSProperties<any>)'.
            Type '() => CSSProperties' is not assignable to type '(props: any) => CreateCSSProperties<any>'.

export interface ClockProps extends WithStyles<typeof styles> {
                                                 ~~~~~~~~~~~~~
```

## Environment

| @material-ui/pickers |3.2.9         |
| material-ui          |4.9.0         |
| TypeScript           |3.7.5         |

mvestergaard:
This PR closes #1465

## Description
This fixes some compatibility issues with material-ui 4.9.0
It should also be backwards compatible

~~The changes to `docs/prop-types.json` was done automatically by some sort of tooling when i committed... Not sure if the tooling isn't compatible with windows or something.~~
I managed to circumvent the tooling that generates and adds `prop-types.json` to the commit.

### DIFF
--- a/lib/src/views/Clock/Clock.tsx
+++ b/lib/src/views/Clock/Clock.tsx
@@ -5,7 +5,7 @@ import ClockType, { ClockViewType } from '../../constants/ClockType';
 import { getHours, getMinutes } from '../../_helpers/time-utils';
 import { withStyles, createStyles, WithStyles, Theme } from '@material-ui/core/styles';
 
-export interface ClockProps extends WithStyles<typeof styles> {
+export interface ClockProps extends WithStyles<keyof ReturnType<typeof styles>> {
   type: ClockViewType;
   value: number;
   onChange: (value: number, isFinish?: boolean) => void;


dmtrKovalenko:
That’s weird. We neeed to fix a problem but not a symptom. Why it was broken? There is no more way to infer style types?

mvestergaard:
See the linked issue, changes were made in @material-ui/styles@4.9.0

mvestergaard:
This is blocking upgrade to material-ui 4.9.0 so I'm just suggesting this small fix to make them work together.
Otherwise maybe @eps1lon can weigh in with how it's intended to work with the changes made to types in 4.9.0

dmtrKovalenko:
Let’s wait a bit, I will investigate this issue when be at office. 1-2 hours

mvestergaard:
An alternative may be to change it to `WithStyles<typeof styles, true>` or `WithStyles<typeof styles, false>` same as in `Calendar`. That overload doesn't appear to cause an error. What do you think?

eps1lon:
> Otherwise maybe @eps1lon can weigh in with how it's intended to work with the changes made to types in 4.9.0

It shouldn't unless your types were unsound to begin with. We had one union member that wasn't used in our test suite though. So this might help finding a test that makes use of that particular member.

I'll probably won't have time looking into this particular issue this week. Though I am working on the picker types currently so this might include a useful side-effect.

dmtrKovalenko:
Yeah let's fix it for now with working overload. Thanks.

mvestergaard:
My suggested alternative doesn't work after all, the reason it works on `Calendar` and not `Clock` is that on `Calendar` the styles are defined as:
```tsx
export const styles = (theme: Theme) => ({
  transitionContainer: {
...
```
while on `Clock`:
```tsx
export const styles = (theme: Theme) =>
  createStyles({
    container: {
...
```

I'm not sure if there are any other consequences from removing `createStyles` than loosing intellisense on the style definition.

mvestergaard:
Removing createStyles results in all sorts of other build issues.
Seems like the original proposed change on this PR is the smallest change needed to make this work.
Can we go with that for now, and at least make it work, then work on making it "correct" later?

mvestergaard:
I've updated the approach to infer the ClassKey using keyof, this should at least be a middle ground you can tolerate?

eps1lon:
So yeah we're encountering this as well. Something else is going on though. It seems like it's using an outdated signature of `createStyles`.

Edit: Pretty sure this is only an issue because typescript inlined the type. It expanded `StyleRules` to `CSSProperties | () => CSSProperties`. In 4.9 we changed the signature of `StyleRules` to just `CSSProperties` but since typescript inlined the older on this change wasn't propagated. 

We just need to bump the used version of core in this repository and republish. 

I would consider this a bug in typescript though. `/core` is a peer and therefore no types should be inlined.

dmtrKovalenko:
I fixed the root of this issue and released v2.3.10, so this can be closed. 
But anyway, big thanks to @mvestergaard for quick contribution 💪

PR #1488, opened 2020-02-03
Add showLeftArrowButton and showRightArrowButton props to CalendarHeader

PickleyD:
Add ability to hide left and right arrow buttons in calendar header. Useful if you want to control the calendar externally.

Still needs tests writing but is this feature something that would be considered?

### DIFF
--- a/docs/prop-types.json
+++ b/docs/prop-types.json
@@ -417,6 +417,36 @@
         "name": "(day: DateIOType) => boolean"
       }
     },
+    "showLeftArrowButton": {
+      "defaultValue": {
+        "value": "true"
+      },
+      "description": "Show left arrow button?",
+      "name": "showLeftArrowButton",
+      "parent": {
+        "fileName": "material-ui-pickers/lib/src/views/Calendar/Calendar.tsx",
+        "name": "CalendarProps"
+      },
+      "required": false,
+      "type": {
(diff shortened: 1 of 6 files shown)


dmtrKovalenko:
Why not to make them just disabled? Why not just do `leftArrowButtonProps={{ hidden: true }}`?

PickleyD:
Am I wrong in thinking that the props must match those specified here: https://material-ui.com/api/icon-button/ ?

dmtrKovalenko:
I mean that your case is really rare. It will be easier for you and for us just to hide your buttons with css. 
`leftArrowButtonProps={{ disabled: true, style: { visibility: 'hidden' }}}` Or pass classname that will hide the button, or anything

PickleyD:
Understood. You're right. Thanks for your help!

On a related note. Here's something that originally gave me unexpected results...
Currently the props are in this order: 

```
{...leftArrowButtonProps}
disabled={isPreviousMonthDisabled}
```
which means that any `disabled` prop you pass in may be overridden by the `isPreviousMonthDisabled` / `isNextMonthDisabled` props. Is this intentional behaviour or should their order be swapped?

dmtrKovalenko:
Yes, to prevent unexpected behavior of opening a month with all days disabled

PR #1503, opened 2020-02-10
Feature/accessibility

### LINKED ISSUE #1462: [DatePicker] Improve accessibility
Testing!

I'm not an a11y expert but I was comparing the performance of the MUI datepicker with an accessible version (https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html) with a couple of different screenreaders, and it doesn't seem to be very usable with the following screenreader/browser combinations:

VoiceOver & Safari, Mac OS Catalina 10.15.2
ChromeVox Classic Extension 53.0.2784.6 & Chrome 79.0.3945.130, Mac OS Catalina 10.15.2
JAWS 2019.1912.1 & Chrome 79.0.3945.130, Windows 10
NVDA 2019.2.1 & Firefox 68.4.1esr, Windows 10

## Environment

https://material-ui.com/components/pickers/#material-ui-pickers
VoiceOver & Safari, Mac OS Catalina 10.15.2
ChromeVox Classic Extension 53.0.2784.6 & Chrome 79.0.3945.130, Mac OS Catalina 10.15.2
JAWS & Chrome, Windows 10
NVDA & Firefox, Windows 10

## Steps to reproduce

1. Test https://material-ui.com/components/pickers/#material-ui-pickers "Date picker inline" example with the screenreader combinations. 
2. Test https://material-ui.com/components/pickers/#material-ui-pickers "Date picker dialog" example with the screenreader combinations. 

## Expected behavior

I'd expect to see behaviour like:
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html
as per the standards described in https://www.w3.org/TR/wai-aria-practices/#dialog_modal

### MUI datepicker - inline option:

JAWS + Chrome, Windows 10: 

- Tab navigation to the calendar button does not announce the chosen date. 
- Selecting the calendar button does not announce that a modal dialog has been opened. 
- Selecting the calendar button announces all the text and day buttons in the calendar immediately.
- Next month buttons are announced as “unlabelled button”.
- Using SPACE or ENTER key on the number buttons does not close the datepicker dialog window. 

NVDA + Firefox, Windows 10:

- Tab navigation to the calendar button does not announce the chosen date. 
- Selecting the calendar button does not announce that a modal dialog has been opened, in fact, it announces nothing. 
- Next month buttons are announced as “button” only.
- Using SPACE or ENTER key on the number buttons does not close the datepicker dialog window. 

VoiceOver + Safari, Mac OS:

- Selecting the calendar button announces “and 10 more items (group)“, which doesn’t make sense. 
- The focus is not brought to the current date when opening the table by selecting the calendar button. 
- Next month buttons are announced as “button” only.
- Using SPACE or ENTER key on the number buttons does not close the datepicker dialog window. 

VoiceOver + Chrome, Mac OS:

- Selecting the calendar button announces “and 4 more items (group)“, which doesn’t make sense. 
- The focus is not brought to the current date when opening the table by selecting the calendar button. 
- Next month buttons are announced as “button” only.
- Using SPACE on the number buttons does not close the datepicker dialog window. 

ChromeVox + Chrome, Mac OS:

- Selecting the calendar button announces nothing. 
- The focus is not brought to the current date when opening the calendar by selecting the calendar button, instead the focus is on the whole dialog.
- Next month buttons are announced as “button” only.
- Using arrow keys changes the date but does not announce anything or change the focus.
- Using SPACE on the number buttons does not close the datepicker dialog window. 

### MUI datepicker - dialog option:

JAWS + Chrome, Windows 10: 

- Tab navigation to the calendar button does not announce the chosen date. 
- Selecting the calendar button announces a dialog has been opened but does not bring the focus to the current date. 
- Next month buttons are announced as “unlabelled button”.
- Using SPACE or ENTER key on the number buttons does not close the datepicker dialog window. 

NVDA + Firefox, Windows 10:

- Tab navigation to the calendar button does not announce the chosen date. 
- Selecting the calendar button announces a clickable dialog, but focus is not on the current date, only the whole dialog. 
- Next month buttons are announced as “button” only.
- Using SPACE or ENTER key on the number buttons does not close the datepicker dialog window. 

VoiceOver + Safari, Mac OS:

- Selecting the calendar button announces “2014 and 14 more items (group)“, which doesn’t make sense. 
- The focus is not brought to the current date when opening the table by selecting the calendar button. 
- Next month buttons are announced as “button” only.
- Using SPACE key on the number buttons does not close the datepicker dialog window. 

VoiceOver + Chrome, Mac OS:

- Selecting the calendar button announces that the current date is selected, but doesn’t announce a dialog has been opened. 
- The focus is not brought to the current date when opening the table by selecting the calendar button. 
- Next month buttons are announced as “button” only.
- Using SPACE on the number buttons does not close the datepicker dialog window. 

ChromeVox + Chrome, Mac OS:

- Selecting the calendar button announces “enter dialog, dialog” only. 
- The focus is not brought to the current date when opening the calendar by selecting the calendar button, instead the focus is on the whole dialog, and the first tab navigation brings you to the year, which is announced as a button, but ENTER on this button closes the dialog. 
- SPACE on the year button allows you to change the year, but the focus is not brough to the current year option, and tab navigation brings you to the first year entry option - 1899. 
- Tabbing from the year button brings you to the current date, which is announced as a button, but SPACE does nothing and ENTER closes the dialog. 
- Next month buttons are announced as “button” only.
- Using arrow keys changes the date but does not announce anything or change the focus.
- Using SPACE on the number buttons does not close the datepicker dialog window. 

## Live example

https://material-ui.com/components/pickers/#material-ui-pickers

dmtrKovalenko:
This PR closes #1462 

## Description
PR is giantly improving accessibility of pickers.

### DIFF
--- a/lib/src/DateTimePicker/DateTimePickerTabs.tsx
+++ b/lib/src/DateTimePicker/DateTimePickerTabs.tsx
@@ -71,8 +71,8 @@ export const DateTimePickerTabs: React.SFC<DateTimePickerTabsProps> = ({
         className={classes.tabs}
         indicatorColor={indicatorColor}
       >
-        <Tab value="date" icon={<>{dateRangeIcon}</>} />
-        <Tab value="time" icon={<>{timeIcon}</>} />
+        <Tab value="date" aria-label="pick date" icon={<>{dateRangeIcon}</>} />
+        <Tab value="time" aria-label="pick time" icon={<>{timeIcon}</>} />
       </Tabs>
     </Paper>
   );
--- a/lib/src/_helpers/utils.ts
+++ b/lib/src/_helpers/utils.ts
@@ -1,3 +1,5 @@
+import * as React from 'react';
+
 /** Use it instead of .includes method for IE support */
 export function arrayIncludes<T>(array: T[] | readonly T[], itemOrItems: T | T[]) {
   if (Array.isArray(itemOrItems)) {
@@ -6,3 +8,13 @@ export function arrayIncludes<T>(array: T[] | readonly T[], itemOrItems: T | T[]
 
   return array.indexOf(itemOrItems) !== -1;
 }
+
+export const onSpaceOrEnter = (innerFn: () => void) => (e: React.KeyboardEvent) => {
+  if (e.key === 'Enter' || e.key === ' ') {
+    innerFn();
+
+    // prevent any side effects
+    e.preventDefault();
+    e.stopPropagation();
+  }
+};
--- a/lib/src/views/Calendar/Year.tsx
+++ b/lib/src/views/Calendar/Year.tsx
@@ -17,16 +17,16 @@ export interface YearProps {
 
 export const useStyles = makeStyles(
   theme => ({
-    yearContainer: {
+    yearButtonContainer: {
       flexBasis: '33.3%',
       display: 'flex',
       justifyContent: 'center',
       padding: '8px 0',
     },
-    yearContainerDesktop: {
+    yearButtonContainerDesktop: {
       flexBasis: '25%',
     },
-    yearLabel: {
+    yearButton: {
       height: 36,
       width: 72,
       borderRadius: 16,
@@ -35,15 +35,15 @@ export const useStyles = makeStyles(
       justifyContent: 'center',
       cursor: 'pointer',
       outline: 'none',
-      '&:focus': {
+      '&:focus, &:hover': {
         backgroundColor: fade(theme.palette.action.active, theme.palette.action.hoverOpacity),
       },
     },
     yearSelected: {
       color: theme.palette.getContrastText(theme.palette.primary.main),
       backgroundColor: theme.palette.primary.main,
-      '&:focus': {
-        backgroundColor: theme.palette.primary.light,
+      '&:focus, &:hover': {
+        backgroundColor: theme.palette.primary.dark,
       },
     },
     yearDisabled: {
@@ -79,8 +79,8 @@ export const Year: React.FC<YearProps> = ({
       role="button"
       ref={forwardedRef}
       onClick={() => onSelect(value)}
-      className={clsx(classes.yearContainer, {
-        [classes.yearContainerDesktop]: wrapperVariant === 'desktop',
+      className={clsx(classes.yearButtonContainer, {
+        [classes.yearButtonContainerDesktop]: wrapperVariant === 'desktop',
       })}
     >
       <Typography
@@ -90,7 +90,7 @@ export const Year: React.FC<YearProps> = ({
         color={selected ? 'primary' : undefined}
         children={children}
         onKeyPress={onSpaceOrEnter(() => onSelect(value))}
-        className={clsx(classes.yearLabel, {
+        className={clsx(classes.yearButton, {
           [classes.yearSelected]: selected,
           [classes.yearDisabled]: disabled,
         })}
--- a/lib/src/views/Calendar/Year.tsx
+++ b/lib/src/views/Calendar/Year.tsx
@@ -69,7 +69,6 @@ export const Year: React.FC<YearProps> = ({
 
   React.useEffect(() => {
     if (focused && ref.current) {
-      console.log(ref.current);
       ref.current.focus();
     }
   }, [focused]);
--- a/lib/src/views/Calendar/Year.tsx
+++ b/lib/src/views/Calendar/Year.tsx
@@ -66,10 +66,10 @@ export const Year: React.FC<YearProps> = ({
   const classes = useStyles();
   const ref = React.useRef<HTMLSpanElement>(null);
   const wrapperVariant = React.useContext(WrapperVariantContext);
-  const handleClick = React.useCallback(() => onSelect(value), [onSelect, value]);
 
   React.useEffect(() => {
     if (focused && ref.current) {
+      console.log(ref.current);
       ref.current.focus();
     }
   }, [focused]);
@@ -77,8 +77,8 @@ export const Year: React.FC<YearProps> = ({
   return (
     <div
       role="button"
-      onClick={handleClick}
       ref={forwardedRef}
+      onClick={() => onSelect(value)}
       className={clsx(classes.yearContainer, {
         [classes.yearContainerDesktop]: wrapperVariant === 'desktop',
       })}
@@ -87,9 +87,16 @@ export const Year: React.FC<YearProps> = ({
         ref={ref}
         variant="subtitle1"
         tabIndex={selected ? 0 : -1}
-        onKeyPress={handleClick}
         color={selected ? 'primary' : undefined}
         children={children}
+        onKeyPress={e => {
+          if (e.key === 'Enter' || e.key === ' ') {
+            onSelect(value);
+
+            e.preventDefault();
+            e.stopPropagation();
+          }
+        }}
         className={clsx(classes.yearLabel, {
           [classes.yearSelected]: selected,
           [classes.yearDisabled]: disabled,
--- a/lib/src/views/Calendar/Year.tsx
+++ b/lib/src/views/Calendar/Year.tsx
@@ -89,7 +89,7 @@ export const Year: React.FC<YearProps> = ({
         tabIndex={selected ? 0 : -1}
         color={selected ? 'primary' : undefined}
         children={children}
-        onKeyPress={onSpaceOrEnter(() => onSelect(value))}
+        onKeyDown={onSpaceOrEnter(() => onSelect(value))}
         className={clsx(classes.yearButton, {
           [classes.yearSelected]: selected,
           [classes.yearDisabled]: disabled,
--- a/lib/src/views/Calendar/Year.tsx
+++ b/lib/src/views/Calendar/Year.tsx
@@ -8,7 +8,8 @@ export interface YearProps {
   children: React.ReactNode;
   disabled?: boolean;
   onSelect: (value: any) => void;
-  selected?: boolean;
+  selected: boolean;
+  focused: boolean;
   value: any;
   forwardedRef?: React.Ref<HTMLDivElement>;
 }
@@ -59,27 +60,36 @@ export const Year: React.FC<YearProps> = ({
   selected,
   disabled,
   children,
+  focused,
   ...other
 }) => {
   const classes = useStyles();
+  const ref = React.useRef<HTMLSpanElement>(null);
   const wrapperVariant = React.useContext(WrapperVariantContext);
   const handleClick = React.useCallback(() => onSelect(value), [onSelect, value]);
+
+  React.useEffect(() => {
+    if (focused && ref.current) {
+      ref.current.focus();
+    }
+  }, [focused]);
 
   return (
     <div
       role="button"
       onClick={handleClick}
+      ref={forwardedRef}
       className={clsx(classes.yearContainer, {
         [classes.yearContainerDesktop]: wrapperVariant === 'desktop',
       })}
     >
       <Typography
+        ref={ref}
         variant="subtitle1"
-        tabIndex={disabled ? -1 : 0}
+        tabIndex={selected ? 0 : -1}
         onKeyPress={handleClick}
         color={selected ? 'primary' : undefined}
         children={children}
-        ref={forwardedRef}
         className={clsx(classes.yearLabel, {
           [classes.yearSelected]: selected,
           [classes.yearDisabled]: disabled,
--- a/lib/src/views/Calendar/Year.tsx
+++ b/lib/src/views/Calendar/Year.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import clsx from 'clsx';
 import Typography from '@material-ui/core/Typography';
+import { onSpaceOrEnter } from '../../_helpers/utils';
 import { makeStyles, fade } from '@material-ui/core/styles';
 import { WrapperVariantContext } from '../../wrappers/WrapperVariantContext';
 
@@ -88,14 +89,7 @@ export const Year: React.FC<YearProps> = ({
         tabIndex={selected ? 0 : -1}
         color={selected ? 'primary' : undefined}
         children={children}
-        onKeyPress={e => {
-          if (e.key === 'Enter' || e.key === ' ') {
-            onSelect(value);
-
-            e.preventDefault();
-            e.stopPropagation();
-          }
-        }}
+        onKeyPress={onSpaceOrEnter(() => onSelect(value))}
         className={clsx(classes.yearLabel, {
           [classes.yearSelected]: selected,
           [classes.yearDisabled]: disabled,
--- a/docs/layout/components/navigationMap.ts
+++ b/docs/layout/components/navigationMap.ts
@@ -38,6 +38,7 @@ export const navItems = [
   {
     title: 'Guides',
     children: [
+      { title: 'Accessibility', href: '/guides/accessibility' },
       { title: 'Form integration', href: '/guides/form-integration' },
       { title: 'CSS overrides', href: '/guides/css-overrides' },
       { title: 'Global format customization', href: '/guides/formats' },
--- a/docs/layout/styleOverrides.ts
+++ b/docs/layout/styleOverrides.ts
@@ -43,6 +43,21 @@ export const createOverrides = (theme: Theme): StyleRules<any> => ({
     color: theme.palette.text.primary,
   },
   li: theme.typography.body1,
+  '.mui-pickers-markdown-table': {
+    boxShadow: theme.shadows[3],
+    backgroundColor: theme.palette.background.paper,
+    borderCollapse: 'collapse',
+
+    '& th, td': {
+      padding: 16,
+      ...theme.typography.body1,
+      border: `1px solid ${theme.palette.divider}`,
+    },
+
+    '& th': {
+      ...theme.typography.h6,
+    },
+  },
   code: {
     fontSize: 16,
     lineHeight: 1.4,
--- a/docs/next.config.js
+++ b/docs/next.config.js
@@ -4,6 +4,7 @@ const withImages = require('next-images');
 const withTypescript = require('@zeit/next-typescript');
 const rehypePrism = require('@mapbox/rehype-prism');
 const headings = require('./utils/anchor-autolink');
+const tableStyler = require('./utils/table-styler');
 const withTM = require('next-transpile-modules');
 const slug = require('remark-slug');
 const webpack = require('webpack');
@@ -13,7 +14,7 @@ const withMDX = require('@zeit/next-mdx')({
   extension: /\.(md|mdx)?$/,
   options: {
     hastPlugins: [rehypePrism],
-    mdPlugins: [slug, headings],
+    mdPlugins: [slug, headings, tableStyler],
   },
 });
 
--- a/docs/package.json
+++ b/docs/package.json
@@ -31,6 +31,7 @@
     "@types/prismjs": "^1.9.1",
     "@types/react": "^16.8.13",
     "@types/react-kawaii": "^0.11.0",
+    "@types/react-redux": "^7.1.7",
     "@types/sinon": "^7.0.13",
     "@zeit/next-bundle-analyzer": "^0.1.2",
     "@zeit/next-css": "^1.0.1",
@@ -71,8 +72,7 @@
     "sinon": "^7.3.2",
     "styled-jsx": "^3.2.4",
     "ts-loader": "^6.0.2",
-    "typescript": "^3.4.4",
-    "@types/react-redux": "^7.1.7"
+    "typescript": "^3.4.4"
   },
   "devDependencies": {
     "dotenv": "^7.0.0",
--- a/docs/pages/guides/Accessibility.example.jsx
+++ b/docs/pages/guides/Accessibility.example.jsx
@@ -6,7 +6,9 @@ function BasicDatePicker() {
 
   return (
     <DatePicker
-      label="Basic example"
+      disableMaskedInput
+      placeholder="mm/dd/yyyy"
+      label="Accessible"
       value={selectedDate}
       onChange={date => handleDateChange(date)}
     />
--- a/docs/pages/guides/accessibility.mdx
+++ b/docs/pages/guides/accessibility.mdx
@@ -3,7 +3,7 @@ import Example from '_shared/Example';
 import PageMeta from '_shared/PageMeta';
 import * as Accessibility from './Accessibility.example';
 
-<PageMeta title="Accessiblity" />
+<PageMeta title="Accessiblity" description="Accessibility guideline for @material-ui/pickers" />
 
 ## Accessibility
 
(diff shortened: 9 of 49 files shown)


lib/src/DateTimePicker/DateTimePickerTabs.tsx
```diff
@@ -71,8 +71,8 @@ export const DateTimePickerTabs: React.SFC<DateTimePickerTabsProps> = ({
         className={classes.tabs}
         indicatorColor={indicatorColor}
       >
-        <Tab value="date" icon={<>{dateRangeIcon}</>} />
-        <Tab value="time" icon={<>{timeIcon}</>} />
+        <Tab value="date" aria-label="pick date" icon={<>{dateRangeIcon}</>} />
```
dmtrKovalenko, line 74: Todo make replaceable

dmtrKovalenko:
@ahayes91 if you could take a look on the deployment preview of this PR to check accessibility changes - you will really help!

ahayes91:
> @ahayes91 if you could take a look on the deployment preview of this PR to check accessibility changes - you will really help!

Sure, I'd be very happy to test this out - I'll go with a few SR/browser combinations. ~~I'll start a comment here but it'll probably take me all day to fill it out so I'll remove this WORK IN PROGRESS comment whenever I'm done 😁~~ 

# Overall issues:

- Mobile version is not compatible with VoiceOver & Safari on iOS or with Talkback on Android at all
- The placeholder of `01/01/2019` is confusing - something like `mm/dd/yyyy` should be used instead like in https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html. (My preference is American format like the examples, but perhaps this could/should be configurable).
- Enter key cannot be used to interact with the buttons on the datepicker dialog, instead the Enter key will just close the dialog. 
- When the datepicker is opened focus is not always set correctly - it should be set on the currently chosen date in the dialog. 
- Typing/reading out the date is too literal and includes the underscores - it should read out the date in a date format (like in https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html) 
- The year dropdown functionality is unreliable across the screenreaders - nothing is announced when it is opened on NVDA, there is no way to choose a year with JAWS using only the keyboard...
- Choosing the date does not close the dialog or update focus correctly with NVDA and FireFox. 

## VoiceOver + Chrome, MacOS

- Tab navigating to the input shows focus correctly on the date input field and reads: "02/10/2020, contents selected, 01/01/2019, edit text. You are currently on a text field. To enter text in this field, type."
- Typing in the field will read out underscores even though the user hasn't typed any underscores.
- If the user types in an invalid date (e.g. 02/31/2020) the "Invalid date" message appears but it is not read out to the user. It's also not clear what date format should be followed (it's mm/dd/yyyy here but there are no audio or visual indicators about this).
- There's probably an issue here with the placeholder text - when I cleared the input with my keyboard backspace keys, "01/01/2019" appeared and that explains what's being read out on first tab. 
- Tab navigation to the calendar button shows focus correctly on the calendar button and reads: 
"Choose date, the selected date is 28th of February, 2020, button. You are currently on a button. To click this button, press Control-Option-Space."
- Control Option Space opens the dialog, but focus stays on the calendar button, and reads "February and 2 more items, dialog, February and 2 more items, group, February February, 2020 2020, You are currently on a group, inside web content. To interact with items in this group press Control Option Shift Down Arrow, To exit this web area, press Control Option Shift Up Arrow". Pressing Control Option Shift Down Arrow does nothing. Pressing Control Option Shift Up Arrow reads "Out of Datepicker - @material-ui/pickers component, web content" but doesn't close the dialog, and moves the focus to the whole screen.
- The Space key will click the button and open the date dialog, but the focus ring disappears, and reads "02/10/2020 selected, February February, 2020, 2020, You are currently on a text field, inside web content. To enter text in this field, type. To exit this web area, press Control Option Shift Up arrow". Hitting the Space key again does nothing. On https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html, the focus is moved automatically to the current date button, and hitting the Space key again chooses that same date and closes the dialog. 
- Tab navigating when the calendar is open is causing Chrome to crash - I restarted my laptop but this is still happening when VoiceOver is on 💥 ❌ 
- The same thing just happened for me on https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker too 💥 This could be a Chrome issue with the latest update to Version 80, so I'll have a look online. 

## VoiceOver + Safari, Mac OS
- Tab navigation to the basic example brings focus correctly to the input and reads: "02/10/2020 contents selected, 01/01/2019, Edit text. 01.01/2019, You are currently on a text field. To enter text in this field, type." The placeholder is being read out incorrectly here again. 
- Typing "2" into the field reads out: "2_/__/____, edit text, invalid data. 01.01/2019, You are currently on a text field. To enter text in this field, type.". Again, underscores being read out incorrectly here and no indication of what a date should look like. 
- Tab navigation will bring focus to the calendar button and read "Choose date, button. You are currently on a button. To click this button, press Control Option Space.". Pressing Control Option Space will open the dialog, bring the focus to "February", and read "February and 11 more items, web dialog, February and 11 more items, group. You are currently on a heading level 6, inside web content. To exit this web area, press Control-Option-Shift-Up arrow.". Pressing Control Option Space again will read "heading level 6, February". Pressing the space key does nothing, and pressing Enter will close the dialog. 
- Tab navigation when the dialog is open will bring focus to the year dropdown button and read "Calendar view is open, switch to year view, button. You are currently on a button, inside web content. To click this button, press Control-Option-Space. To exit this web area, press Control-Option-Shift-Up arrow." Pressing Control Option Space will read "Press year view is open, switch to calendar view button". Tab navigation will bring you to "2020, button". Tab navigation again will only let you move between the 2020 button and the calendar dropdown button. Arrow navigation allows you to move between year options, which are read out as "2021, button". Pressing Space or Control Option Space will choose that year, reopen the calendar view, but the focus is still visually in the same place as the year button. The Februrary 2021 group is read out again. Here, I think the focus should be brought back to the chosen date. 
- Tab navigation allows you to get to the previous and next month buttons correctly, but hitting Space or Control Option Space on them will read "January, January. You are currently on a button inside web content. To click this button, press Control Option Space. To exit this web area, press Control Option Shift Up arrow."
- Pressing the Enter key on any of the year dropdown or next/previous month buttons will close the dialog instead of interacting with that button. 
- Find next control mode (Control Option Cmd J) will find the input field correctly. 
- Pressing ESC will close the dialog correctly but it will also minimize the Safari window itself. 

## VoiceOver + Safari, iOS
- Swipe navigation will bring you to "Basic example" label, and then swiping again will bring you to the input which reads: "02 slash 10 slash 2020, text field, read-only. Double tab to edit. Use the rotor to access misspelled words." Double-tapping on the screen will open the dialog with the focus still in the shape of the input, as per the screenshot below, and reads "02 slash 10 slash 2020, text field, read-only.":
IMG_7352
Left and right or up and down swiping won't allow you to do anything. 
- Unless you can actually see the screen, you can't interact with the dialog at all. ❌ 

## ChromeVox + Chrome, Dell Chromebook
- Tab navigating to the input moves the focus on the input correctly and announces: "01 slash 01 slash 2019, 02 slash 02 slash 2020, selected, Edit text numeric only." (Looks like the same issue with the placeholder being read out here).
- Typing "03" will read out: "01 slash 01 slash 2019, 0_ slash 2 underscores slash 4 underscores Edit text numeric only". (Same issue with both the placeholder being read out, and the underscores, and no indication of what input is expected). The invalid date format label also appears but nothing is actually read out to the user either. 
- Tab navigating to the calendar button will move focus to it correctly and read out: "Choose date, button, press Search + Space to activate". Pressing "Search + Space" will open the dialog, but the focus is still on the calendar button. "February, Heading 6, 2020 Heading 6" is read out." Pressing Search + Space again doesn't do anything.
- Hitting tab will move the focus to the dropdown picker beside "February 2020" and read "Dialog, calendar view is open, switch to year view Button Press Search plus Space to activate". Pressing Search + Space will switch the visual direction of the arrow on the button and read "year view is open, switch to calendar view, Button Press Search plus Space to activate". Using the right arrow key will move the focus to the next year and read "2021, Heading 6". I would think that this should be read out as a button or something interact-able. Pressing Space will close the year view and open the calendar view, with focus on the whole dialog and read: "February 2021 calendar view is open, switch to year view previous month next month S M T W T F S jan 31, 2021. Feb 1, 20201. Feb 2, 2021....." and continue reading out all the elements in the calendar. 
I think the focus should probably move back to the current date in this case for a better experience. 
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html doesn't have a year dropdown picker in the same way so it's not really comparable in this case... https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker is another pretty a11y example that also doesn't have a year dropdown... 
- Tabbing from the year dropdown will bring the focus to the previous month button and read "previous month, Press Search + Space to activate". Pressing Search + Space will leave the focus on the previous month button and read "January, heading 6". Pressing the Enter key on this button will close the dialog, and move the focus back to the calendar button. That doesn't match the behaviour of https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html , where both "Enter" and "Space" keys can be used to interact with the previous/next month buttons. 
- Tabbing from the next month button will bring the focus to the current date and read "Feb 10, 2020 Button, row 3 column 2 Table, 5 by 7, Press Search plus Space to activate. Press Search plus Ctrl plus Alt with arrows to navigate by cell". Pressing "Search plus Space" doesn't do anything at all. Pressing "Search plus Ctrl plus Alt with right arrow" will move the focus to the next date and read "Feb 11, 2020, Button, row 3 column 3, Press Search plus Space to activate". Pressing Search plus Space will move the green date focus to the "11" button, but nothing is read out to the user and the dialog is still open. Clicking the Space button only will close the dialog, with the focus on the calendar button and "Chose date, selected date is February 11, 2020, Button Main Press Search plus Space to activate". If you open the dialog again, the green focus is on 11 but there is still a circle on 10: 
Screenshot 2020-02-10 at 12 02 55 PM
- Hitting the ESC key when the dialog is open will correctly close the dialog. 
- Using form navigation (Search + f) will bring you correctly to the date input. 

## JAWS 2019 + Chrome, Windows 10
- Tab navigating to the input shows the focus correctly and reads: "01/01/2019 edit, 02/10/2020, type in text". Typing in the input will only read out each character typed. Typing "31/02/2020" and then tab navigating away and back to the input will read: "01/01/2019 edit invalid entry, 31/02/2020, type in text". 
- Tab navigating to the calendar button shows focus correctly and reads "Choose Date, selected date is Feb 2, 2020, button. To activate press space bar". Pressing the Space bar will read out "Space", open the dialog, the focus will disappear, and reads "dialog, February 2020 calendar view is open, switch to year view previous month next month S M T W T F S heading level 6". Pressing Space again does nothing other than read "Space". Pressing Enter does nothing other than read "Enter". Pressing ESC reads "ESC", closes the dialog, and reads "Main region, Choose date, selected date is Feb 10 2020, button, to activate press space bar." with the focus on the calendar button.  
- Tab navigating while the dialog is open moves the focus to the year dropdown and reads "Calendar view is open, switch to year view button, to activate press space bar". Pressing Space reads "Space" but nothing else, even though visually the picker has updated to show year options. Tabbing will bring you to "2020 button". "Alt Shift arrow keys" can be used to move between the year options but the focus doesn't seem to match what's happening visually - it's read out as "2021 button, year view is open, switch to calendar view button, to activate press space bar". There doesn't seem to be any way to choose another year option. ❌ 
-Tab navigation brings you to the previous month button with focus "previous month button, to activate press space bar", and pressing Space on previous month button will just announce "space" and nothing else even though the calendar has updated visually. 
- Alt + Arrow keys can be used to move between date options - Enter or Space will choose that date and close the dialog. 
- Form mode (F key) will bring you correctly to the date input. 


## NVDA + FireFox, Windows 10
- Tab navigation to the input brings the focus correctly to the input (and focus mode is activated with the focus mode sound) and reads "01/01/2019 edit has autocomplete. Selected 02/10/2020." Typing "2" into this field reads "2, invalid entry".
- Tab navigating to the calendar button brings focus to the button and reads "Choose date, selected date is February 2nd, 2020, button". Pressing space opens the dialog, but the focus is not apparent, and announces "Dialog, February heading level 6 2020 heading level 6. Calendar view is open. switch to year view button previous month button S M T W T F S..." and reads out all the rest of the buttons on the calendar. Focus is incorrect here again and would be better on the current date. Hitting the space button will stop the announcements, but nothing happens. Tab will bring you to the year dropdown button and announce "calendar view is open, switch to year view button". Hitting Space on this button will change the calendar visually but nothing is announced to the user. Hitting Space again will change back to the date view. 
- When the date dialog (calendar view) is open, tabbing can be used to get to the previous and next month buttons. Hitting "Enter" on the previous month button will change the visual state of the component but it won't announce anything to the user. Hitting "Space" on the previous month button is the same. 
- Tabbing past the next month button brings you to the currently chosen date and reads "table with 5 rows and 7 columns, row 3 column 2 clickable February 10 2020 button". Pressing "Space" or "Enter" on this button doesn't do anything. Alt and right arrow key will not change the focus, but "column 3 clickable February 11th 2020 button" will be announced. Pressing "Space" or "Enter" will change the green selection, but a grey circle still remains on the 10 option and the dialog is still open. Pressing ESC will close the dialog and show that the selected date is whatever was green (February 11 2020). 
- ESC will correctly close the dialog and move focus to the calendar button. 
- Form mode (F) will bring focus to the input and announce "Clickable 01/01/2019 edit has autocomplete. 02/11/2020". It will not bring you to focus mode until you click "Enter" or "Space".  


## Talkback + Chrome, Moto G6 Android V9
- Swiping brings you to the input, but double tapping does nothing. You can't interact with the input at all ❌

oliviertassinari:
@ahayes91 Oh wow, thank you so much for diving deep in the accessibility of the component! This is one of the best reviews I have seen from the community to date! I remember similar excellent feedback grom @ryancogswell :)

lib/src/views/Calendar/Year.tsx
```diff
@@ -59,27 +61,35 @@ export const Year: React.FC<YearProps> = ({
...
       <Typography
+        ref={ref}
         variant="subtitle1"
-        tabIndex={disabled ? -1 : 0}
-        onKeyPress={handleClick}
+        tabIndex={selected ? 0 : -1}
         color={selected ? 'primary' : undefined}
         children={children}
-        ref={forwardedRef}
+        onKeyPress={onSpaceOrEnter(() => onSelect(value))}
```
oliviertassinari: What motivates keypress over keydown?

lib/src/_helpers/utils.ts
```diff
@@ -6,3 +8,13 @@ export function arrayIncludes<T>(array: T[] | readonly T[], itemOrItems: T | T[]
...
   return array.indexOf(itemOrItems) !== -1;
 }
+
+export const onSpaceOrEnter = (innerFn: () => void) => (e: React.KeyboardEvent) => {
+  if (e.key === 'Enter' || e.key === ' ') {
+    innerFn();
+
+    // prevent any side effects
+    e.preventDefault();
+    e.stopPropagation();
```
oliviertassinari, line 18: I think that we should be cautious about the usage of stopPropagation. I think that it would be great to add a comment on the "why?" we use it here for our future self. https://css-tricks.com/dangers-stopping-event-propagation/

dmtrKovalenko:
@ahayes91 Thank you so much for your testing! This really helps, I am in progress fixing issues. Wondering do you have twitter? I'd like to tweet about your help with pickers accessiblity 💪 

But some things I noticed:
1. I have no idea why voiceover is duplicating year and month on chrome. The same I am seeing here https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html so I think it is chrome bug 
2. About text input that reads mask chars `1_/__/____` I have no idea how to fix it. This looks like a real problem that requires us to remove mask at all. I think we need to make an accessibility guideline that will cover all the issues. 
> Can we detect screen readers and disable masked input dynamically? 
3. Pretty same with placeholder - developer needs to pass the proper placeholder manually, because right now we are using localized formats and it looks like `Pp` so just fallback to the example of formatted date
4.  Enter works like in exampe and submitting focused date, then closing dialog

oliviertassinari:
Capture d’écran 2020-02-10 à 21 29 01

@dmtrKovalenko I have [used](https://github.com/mui-org/material-ui/pull/19612) the Pagination as a workbench to experiment around the implementation of the latest ["state" guidelines](https://material.io/design/interaction/states.html). I think that it's something we can leverage for the days in the calendar. Actually, we could almost use the same logic :D.

For instance, it would fix the following issue: consider the selected day, there is no visual clue that it has the focus. We can't distinguish the "selected" to the "selected + focus" state as we can in the [WAI-ARIA example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html).
Another detail would be about the "selected + hover" state, the current implementation makes the background lighter while the [material design spec](https://material.io/design/interaction/states.html#selected) seems to suggest to make it darker.

ahayes91:
> @ahayes91 Thank you so much for your testing! This really helps, I am in progress fixing issues. Wondering do you have twitter? I'd like to tweet about your help with pickers accessiblity 💪

Yes indeed - @theaislinnhayes  👍 

> But some things I noticed:
> 
> 1. I have no idea why voiceover is duplicating year and month on chrome. The same I am seeing here https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html so I think it is chrome bug

If you're seeing the same on that example or on another accessible example https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker then I'd say don't worry about it - Chrome's got its own problems anyway given that version 80 is definitely broken with VoiceOver 🙄 

> 2. About text input that reads mask chars `1_/__/____` I have no idea how to fix it. This looks like a real problem that requires us to remove mask at all. I think we need to make an accessibility guideline that will cover all the issues.
> 
> > Can we detect screen readers and disable masked input dynamically?

I'd be inclined to ditch the mask altogether to be honest. Neither of those few accessible datepicker examples that we have (https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html or https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker) use one - putting the desired format as a placeholder or in the label would be enough for context for all users, and while you'd save a couple of key presses with a mask, it really does hinder the a11y experience. To me it feels like a case of sacrificing a11y user experience for the sake of convenience for non-SR-users. 
It also might not be a good idea (or even possible) to detect screenreaders and disable certain features - even though the most recent WebAim survey indicated quite a few users wouldn't mind if websites could detect use of a screenreader (https://webaim.org/projects/screenreadersurvey8/#detection), there are lots of arguments against doing so for privacy and discriminatory reasons: 
https://marcozehe.wordpress.com/2014/02/27/why-screen-reader-detection-on-the-web-is-a-bad-thing/
https://www.boia.org/blog/should-you-have-a-separate-accessible-website

> 3. Pretty same with placeholder - developer needs to pass the proper placeholder manually, because right now we are using localized formats and it looks like `Pp` so just fallback to the example of formatted date

Can we update the demo examples to show the proper placeholder, so that developers have a better idea of what should be passed?

> 4. Enter works like in exampe and submitting focused date, then closing dialog

👍 - just retested this quickly with VoiceOver on Safari, and hitting Enter on a chosen date in the calendar view is good. The enter button doesn't work on the other buttons though (like previous/next month/ year dropdown) and it also seems like there isn't a way to choose a year from the dropdown either with Space or Enter.... I can retest this again in a few days anyway 💪

attilavago:
I agree with @ahayes91, the inputs masks are not a great idea from an accessibility perspective, so they are required to be removed. 

On the note of browsers detecting screen readers being used, there are much larger issues around it, like privacy and discrimination. The entire standard is built around creating user interfaces and user experiences that work for all. This is why for example keyboard accessibility results in allowing all other input methods used by the disabled communities to work very well, as they emulate the keyboard and thus are backwards compatible with all other user experiences.

dmtrKovalenko:
Masks will not be removed from the datepicker. But there is an option to disable masked input

oliviertassinari:
If we remove the mask support, would we save this dependency 800 B? https://bundlephobia.com/result?p=rifm@0.11.0

dmtrKovalenko:
@oliviertassinari yes, but I have a strong opinion that mask is required

oliviertassinari:
OK, fair enough. Maybe somebody will open an issue about it, and as the component gain adoption and the issue upvotes, we will re-evaluate in the future?
In any case, I recall that we had to set the mask disabled by default to handle localization? So it's not that much a concern?

dmtrKovalenko:
Mask will be disabled for any other from en-US locale (example is here https://dev.material-ui-pickers.dev/localization/date-fns). For the other locales it needs to be enabled manually

dmtrKovalenko:
We keep loosing code coverage percentage, my experiments with collecting coverage from cypress tests are not working. Moved experiments to #1507

oliviertassinari:
@dmtrKovalenko Great iteration on the accessibility of the component! Well done 😍.

I have a couple of feedback based on https://next.material-ui-pickers.dev/demo/datepicker:

1. What about we replace the pulsating ripple for focused days in the calendar with the focus-visible class name? It would solve this case when the popup is open with the keyboard:
Capture d’écran 2020-02-14 à 12 23 07
2. When the popup is open with the keyboard, about 40% of the time, the active day isn't initially focused. I suspect a race condition, somewhere:
Capture d’écran 2020-02-14 à 12 27 02
3. When a date is clicked on, the contrast goes wild
Capture d’écran 2020-02-14 à 12 28 47
4. Should the popup behave like a modal? Blocking all interactions on the rest of the page? It's it a [common](https://trello.com/c/mOYNtDsH/2308-datepicker) approach? I personally feel frustrated by this behavior. There is a related issue about it in https://github.com/mui-org/material-ui/issues/17353.
5. We might have a responsiveness issue. I have recorded the following for the open interaction. In this run, it takes 237ms between the mouse up and the beginning of the appearance of the calendar (transitioning).  

Capture d’écran 2020-02-14 à 12 42 29

This is, for instance, to be compared with jQuery UI (134ms):

Capture d’écran 2020-02-14 à 12 46 13

http://react-day-picker.js.org/ seems to perform great with about 90ms of delay.

While jQuery UI feels a bit slow, I think that it would be great to look into the performance of our date picker. [Google recommends 100ms](https://developers.google.com/web/fundamentals/performance/rail) as the upper boundary for a great experience :D.

ahayes91:
Have the accessibility improvements for datepicker been released yet? I couldn't figure out if this is linked to any of the latest releases 🤓

mhchen:
@ahayes91 not sure if you're still waiting on an answer, but my understanding is that this went out in the 4.x alpha releases. I recently had to upgrade to 4 Alpha 10 to address these a11y issues (and thank you for being so involved in the process!). I picked 10 because it is the last one before requiring an upgrade to Material-UI core 5 Alpha. I wish these changes could've been published in a stable version, but I'm glad to be able to use them now anyway.

PR #1504, opened 2020-02-11
Add new prop for custom date format on toolbar

### LINKED ISSUE #1345: New property to change the date/time format from the toolbar

I was thinking than maybe you should include a new property to define the property of the toolbar, right now the `format` property from `KeyboardDatePicker`, `KeyboardTimePicker` and `KeyboardDateTimePicker` changes the date pattern of the input text field but does not affect the toolbar format. 

So, basically you have at least 2 different formats in the picker component. One, which is the one the user sees when the datepicker is closed in the input text field and the other one corresponds to the format of the date in the calendar itself.

So, my idea was either affect even the date format from the toolbar with the `format` property or to create a new property, let's say `toolbarFormat` to support the change of its format.

**Additional context**
Screen Shot 2019-10-09 at 10 54 32 AM

ghost:
- MuiPickerUtilsProvider can now receive an object of formats
- Closes #1345

This PR closes #1345  

## Description
- Providing an object of formats that can override the default initial formats in each of the @date-io library util interfaces
example for moment library:
```javascript
import React from "react"
import moment from "moment"
import "moment/locale/ko"
import MomentUtils from "@date-io/moment"
import { MuiPickersUtilsProvider, DatePicker } from "@material-ui/pickers"

moment.locale("ko")

const CustomFormattedCalendar = () => {
  return (
    // Providing normalDate format key as "MMM Do, dddd" to override defaultFormat "D MMMM" in @date-io/moment utils
    // as DatePicker uses `utils.format(date, "normalDate")` to display date on tollbar for non english locale
    <MuiPickersUtilsProvider utils={MomentUtils} libInstance={moment} locale="ko" libFormat={{ normalDate : "MMM Do, dddd" }}>
         ......
```
Please review :)

### DIFF
--- a/lib/src/MuiPickersUtilsProvider.tsx
+++ b/lib/src/MuiPickersUtilsProvider.tsx
@@ -10,6 +10,7 @@ export interface MuiPickersUtilsProviderProps {
   children: React.ReactNode;
   locale?: any;
   libInstance?: any;
+  libFormat?: any;
 }
 
 export const MuiPickersUtilsProvider: React.FC<MuiPickersUtilsProviderProps> = ({
@@ -17,12 +18,12 @@ export const MuiPickersUtilsProvider: React.FC<MuiPickersUtilsProviderProps> = (
   children,
   locale,
   libInstance,
+  libFormat,
 }) => {
-  const utils = React.useMemo(() => new Utils({ locale, instance: libInstance }), [
-    Utils,
-    libInstance,
-    locale,
-  ]);
+  const utils = React.useMemo(
+    () => new Utils({ locale, instance: libInstance, formats: libFormat }),
+    [Utils, libInstance, locale, libFormat]
+  );
 
   return <MuiPickersContext.Provider value={utils} children={children} />;
 };


lib/src/MuiPickersUtilsProvider.tsx
```diff
@@ -10,19 +10,20 @@ export interface MuiPickersUtilsProviderProps {
   children: React.ReactNode;
   locale?: any;
   libInstance?: any;
+  libFormat?: any;
```
dmtrKovalenko, line 13: Its something I am planned to do! Thanks 👍 
But I think better naming will be something like `formats`, the same what we are passing down to adapter

dmtrKovalenko (changes requested):

lib/src/MuiPickersUtilsProvider.tsx
oliviertassinari: I wonder, have we considered this API?

```jsx
import React from "react"
import moment from "moment"
import "moment/locale/ko"
import MomentUtils from "@date-io/moment"
import { MuiPickersUtilsProvider, DatePicker } from "@material-ui/pickers"

moment.locale("ko")

const utils = new MomentUtils({
  locale: 'ko',
  instance: moment,
  format: {
    normalDate : "MMM Do, dddd",
  },
});

const CustomFormattedCalendar = () => {
  return (
    <MuiPickersUtilsProvider utils={utils}>
```

PR #1507, opened 2020-02-13
[WIP] Cypress coverage

dmtrKovalenko:
Try to collect coverage from cypress tests


oliviertassinari:
@dmtrKovalenko What's the objective of this pull request? What's the status of it?

dmtrKovalenko:
Can be closed, anyway will be supressed by other PR

oliviertassinari:
@dmtrKovalenko It seems that the objective is to get test coverage reporting. If it's the case, we can defer the resolution of the problem to https://github.com/mui-org/material-ui/issues/19706, where it's already set up in the main repository. Is that correct?

dmtrKovalenko:
No, I intend to keep cypress tests so it needs to be done anyway.
> It will be much easier after migrating all tests to components

oliviertassinari:
@dmtrKovalenko Would you mind expanding? From https://github.com/mui-org/material-ui-pickers/pull/1730#discussion_r419114709, I thought that the policy was to replace cypress with testing-library, and introduce it back or an alternative tool like [playwright](https://github.com/mui-org/material-ui/pull/21449) as we hit a roadblock (depending on the pain points and what can best solve it). In such a roadmap, coverage with cypress has little value.

dmtrKovalenko:
Haha, I had absolutely different view

oliviertassinari:
@dmtrKovalenko What do you think about it?

PR #1525, opened 2020-02-21
Implement min/max time validation

### LINKED ISSUE #519: Add minTime & minDateTime props to disable period of time
Just a suggestion, I have a case where I need the earliest time selectable to be exactly 24 hours ago.

### LINKED ISSUE #821: Add minTime and maxTime properties to TimePicker
Hi ,  I suggest adding two poperties, minTime and maxTime. the input value would be checked if are in the range ( if it's greater than minTime or it's less than maxTime or it's in between minTime and maxTime properties) after filling the input; If it is not in range the function prop onError would be trigged and the input highlighted in red. Thk you.

dmtrKovalenko:
This PR closes #519  

## User facing changes

`TimePicker`: Add new props:
* minTime: `ParsableDate`
* maxTime: `ParsableDate` 
* shouldDisableTime : `(timePoint: number, clockType: 'hours' | 'minutes' | 'seconds') => boolean`

`DateTimePicker`: Add new props:
* minDateTime: `ParsableDate` – minimal selectable point of time
* maxDateTime: `ParsableDate` – minimal selectable point of time

Disabled parts of clock are visually highlighted by clock numbers and not selected nor from mouse nor from touch events 
image

If the time point is disabled clock pointer will not be displayed 
image


oliviertassinari (commented):
A great feature!

I wonder about the API. From my understanding of all the possible use cases, there is no way we can save the usage of `minTime`, so it's perfect we have it. For instance, somebody might want to **disable all booking slots in the morning**. It makes sense both when using the DateTimePicker or the TimePicker.

> `minDateTime`: `minDate` + `minTime` combined. 

However, I'm confused by the description of `minDateTime`. Could the wording be counter-intuitive? 
I would expect the following: the prop is used to set a minimum date-time. For instance, with `minDateTime="2020-02-21T18:00:00"` I expect any date-time after this point in time can be valid, but the description makes me think that `2020-02-22T12:00:00` is invalid.

What if we had 2 props `minTime` and `minDate` instead of 3?

lib/src/__tests__/e2e/TimePicker.test.tsx
```diff
@@ -124,3 +129,120 @@ describe('e2e - TimePicker with seconds', () => {
...
+  });
+
+  it('Should not select disabled second', () => {
+    component.find('button[data-mui-test="seconds"]').simulate('click');
+    component.find('Clock div[role="menu"]').simulate('touchMove', clockTouchEvents['--:20']);
+
+    expect(component.find('button[data-mui-test="seconds"] h3').text()).toBe('00');
+  });
+
+  it.only('Should select enabled second', () => {
```
oliviertassinari:
```suggestion
  it('Should select enabled second', () => {
```

dmtrKovalenko:
Just to clarify `minDateTime` vs `minTime` 

`minDateTime` - can be used to set selectable range from `1st January 12:00` to `1st February 12:00` for example. 
`minTime` - can be set to disable a time range in each day, for example, when it needs to disable night hours (schedule of work from 10:00 to 19:00) and also select a date.

oliviertassinari:
@dmtrKovalenko Ok, it's very clear and a great behavior.
So now I wonder. From a user perspective, when is it better to use `minDate` than `minDateTime`?

dmtrKovalenko:
There is no technical difference for calendar.  But minDate will not be used for time validation. 
I think we need to keep them different. 

Because `minDate` can be used like `addDays(new Date(), -15)` — in this case time will not be disabled, because it will be used only for dates. But `minDateTime= addDays(new Date(), -15)` will restrict minTime of -15 day as well.

oliviertassinari:
Ok, so If I understand correctly, the usage of two props is motivated to ease the usage of the following API:
```jsx
import { addDays } from 'date-fns';

<DateTimePicker minDate={addDays(new Date(), -15)} />
```

which is simpler than?

```jsx
import { addDays, startOfDay } from 'date-fns';

<DateTimePicker minDateTime={addDays(startOfDay(new Date()), -15)} />
```

dmtrKovalenko:
The main idea is to provide an easier way to migrate for the current user. `minDate` will continue works as in v3. 
Plus using `minDateTime` can be confusing, requiring to annulate time in the date can hide issues. `minDateTime` will be used much rare than just `minDate`

oliviertassinari:
@dmtrKovalenko Ok, thanks. I haven't benchmarked with the other libs, but it sounds coherent. In any case, we will learn more once we ship 🚢.

bartosz-lobejko-trabr:
I just pulled new Alpha release and try to use minTime prop but got this problem:

`Attempted import error: 'MuiPickersUtilsProvider' is not exported from '@material-ui/pickers'.`

```
    "@material-ui/core": "^4.9.7",
    "@material-ui/icons": "^4.9.1",
    "@material-ui/lab": "^4.0.0-alpha.46",
    "@material-ui/pickers": "^4.0.0-alpha.4",
```

Before that I had error about react not recognizing minTime prop, that's why I updated all Material-ui

dmtrKovalenko:
Read the release notes. There are breaking changes in each alpha release. 

We will make an update guide and provide scripts closer to the stable v4 release.

PR #1534, opened 2020-02-23
Api changes

dmtrKovalenko:
This PR is a part of vNext #1293 development 

### New props 
* `toolbarFormat` - easily customize date string displaying in toolbar (#1345)
* `disableHighlightToday` - disable highlighting of today date with a circle 
* `showDaysOutsideCurrentMonth` - show days outside of current month in calendar
* `disableMargin` - (only `Day` prop) disable margin between days, useful for range displaying when days are linked together with background color

### Remove props
* `labelFunc` - completely unusable function with current keyboard input design
* `invalidLabel` - pretty same, unusable with keyboard input, use `error` and `helperText` instead

### Renames
* `format` => `inputFormat` 
* `emptyLabel` => `emptyInputText` 
* `initialFocusedDate` => `defaultHighlight` 
* `title` => `toolbarTitle` (+ make it accept value from `label` as default)

### Change signatures
* `renderDay` 
```ts
(date: DateIOType, selectedDate: DateIOType, dayInCurrentMonth: boolean, defaultDay: React.ReactNode) => React.ReactNode 
```
```ts
(date: DateIOType, selectedDate: DateIOType,  DayComponentProps:  DayProps) => React.ReactNode 
```
#### Motivation:
Better customization of the day displaying, by allowing to spread down and override any props we need, before it was possible only by `React.cloneElement`, but it was not so powerful like spreading props down. This also better from performance perspective when overriding days, because we are not rendering day bides ourself if user will render something different.
```tsx
  renderDay={(day, selectedDate, DayComponentProps) => {
     return (
        <Badge overlap="circle" badgeContent={isSelected ? '🌚' : undefined}>
          <Day {...DayComponentProps} isToday disableMargin aria-label="Something"  />
        </Badge>
     );
  }}
```

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -1,0 +1,7 @@
+version: 1
+snapshot:
+  percy-css: |
+    #codefund {
+      display: none;
+    }
+  |
--- a/docs/_shared/Ad.tsx
+++ b/docs/_shared/Ad.tsx
@@ -36,6 +36,10 @@ const Ad: React.FC = () => {
     }
   }, []);
 
+  if (process.env.VISUAL_TESTING) {
+    return null;
+  }
+
   return (
     <Grid container>
       <span id="codefund-script-position" />
--- a/docs/_shared/PropTypesTable.tsx
+++ b/docs/_shared/PropTypesTable.tsx
@@ -1,6 +1,7 @@
 import clsx from 'clsx';
 import Code from './Code';
 import FuzzySearch from 'fuzzy-search';
+import ReactMarkdown from 'react-markdown';
 import PropTypesDoc from '../prop-types.json';
 import SearchBar from 'material-ui-search-bar';
 import React, { useMemo, useState } from 'react';
@@ -141,7 +142,7 @@ const PropTypesTableLazy: React.FC<PropTypesTableProps> = ({ disableHeader, src 
                 </TableCell>
 
                 <TableCell className={classes.description}>
-                  <span dangerouslySetInnerHTML={{ __html: prop.description }} />
+                  <ReactMarkdown source={prop.description} />
                 </TableCell>
               </TableRow>
             ))}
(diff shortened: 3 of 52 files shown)


oliviertassinari:
Regarding `toolbarTitle`, should we call it `toolbarText`? "Text" being a common suffix we have for strings that should/can be localized?

oliviertassinari (approved):
Do you keep track of the API breaking changes somewhere? Maybe we should start a migration guide?

dmtrKovalenko:
I am following material design guidelines, there it is called `title`, maybe call it `toolbarTitleText`? 
Keeping breaking changes only in releases, just need to combine all releases together. 
But yeah starting migration guide is a nice idea, I think maybe we need to provide codemod as well at least for prop renames?

oliviertassinari:
> I am following material design guidelines, there it is called title, maybe call it toolbarTitleText?

I'm not sure the developers coming to the component will be aware of how the material design specification calls the label. I think that as long as it's intuitive and consistent, it will be great. No strong preference, `toolbarTitleText`, `toolbarText`, `titleText` they sounds great.

> codemod as well at least for prop renames?

In terms of user experience, I think that:

1. a note in each release is a good
1. a migration guide is great
1. a codemod is excellent

I think that 2. is a nice tradeoff between the DX and the time required to make it happen. What do you think?

dmtrKovalenko:
The migration guide will be done anyway. Like we do for v3 https://material-ui-pickers.dev/guides/upgrading-to-v3
I have no experience with codemod scripts so have no idea how much time it will take

oliviertassinari:
> I have no experience with codemod scripts so have no idea how much time it will take

I would expect a codemod to take about an afternoon of work or more. But I have never done prop renaming for a codemod (at least I only remember doing import renaming). My estimate could be wrong.

dmtrKovalenko:
I have no idea what's wrong with percy 🤔

PR #1537, opened 2020-02-25
LocalizationProvider

dmtrKovalenko:
Closes #1504 

#### LocalizationProvider
Rename and changes in API for `MuiPickersUtilsProvider`. It is renamed to `LocalizationProvider`. In future this provider may be moved to the `@material-ui/core`

```diff
- <MuiPickersUtilsProvider utils={DateFnsUtils} />
+ <LocalizationProvider dateAdapter={DateFnsUtils}  />
````

Also, `LocalizationProvider` now accepts special `dateFormats` prop that allows globally override formats used for displaying dates

```tsx
const formats = {
  normalDate: 'd MMM yyy',
  keyboardDate: 'd MMM yyy',
};

<LocalizationProvider
  dateAdapter={DateFnsAdapter}
  locale={frLocale}
  dateFormats={formats}
/>
```

Also, `libInstance` from now renamed to `dateLibInstance` which is a more meaningful name in case of reusable `LocalizationProvider` 

#### dateAdapter prop

There is a new prop `dateAdapter` in each component which allows passing configured date adapter object without context 

```tsx
const memoizedDateAdapter = useMemo(() => {
    return new DateFnsAdapter({ locale });
  }, [locale]);

<DatePicker
  value={selectedDate}
  onChange={date => handleDateChange(date)}
  dateAdapter={staticDateAdapter}
/>;
```


README.md
```diff
@@ -43,19 +43,19 @@ npm i luxon @date-io/luxon
...
+Then teach pickers which library to use with `LocalizationProvider`. This component takes an adapter property, and makes it available down the React tree thanks to React context. It should preferably be used at the root of your component tree.
 
 ```jsx
-import MomentUtils from '@date-io/moment';
-import DateFnsUtils from '@date-io/date-fns';
-import LuxonUtils from '@date-io/luxon';
-import { MuiPickersUtilsProvider } from '@material-ui/pickers';
+import Luxon from '@date-io/luxon';
+import Moment from '@date-io/moment';
+import DateFns from '@date-io/date-fns';
```
oliviertassinari: Should the import come from?  
```suggestion
import dateFns from '@material-ui/pickers/adapter/date-fns';
```

But I wonder about keeping the file system flat?

```suggestion
import dateFnsAdapter from '@material-ui/pickers/dateFnsAdapter';
```

README.md
```diff
@@ -43,19 +43,19 @@ npm i luxon @date-io/luxon
...
+import { LocalizationProvider } from '@material-ui/pickers';
 
 function App() {
   return (
-    <MuiPickersUtilsProvider utils={DateFnsUtils}>
+    <LocalizationProvider adapter={DateFns}>
```
oliviertassinari: "adapter" might not make it clear that it's about handling dates, maybe?

```suggestion
    <LocalizationProvider date={dateFnsAdapter}>
```

docs/pages/guides/DateAdapterProp.example.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
```
oliviertassinari, line 1: What do you think of using `React.useState`?

```suggestion
import React from 'react';
```

docs/pages/guides/Formats.example.jsx
```diff
@@ -1,31 +1,29 @@
...
+  normalDate: 'd MMM yyy',
+  keyboardDate: 'd MMM yyy',
+};
 
 function DateFnsLocalizationExample() {
   const [selectedDate, handleDateChange] = useState(new Date());
 
   return (
-    <MuiPickersUtilsProvider utils={LocalizedUtils} locale={frLocale}>
+    <LocalizationProvider adapter={DateFnsAdapter} locale={frLocale} libFormats={formats}>
```
oliviertassinari: Maybe to make it clear it's about formating dates?

```suggestion
    <LocalizationProvider adapter={DateFnsAdapter} locale={frLocale} dateFormats={formats}>
```

(If we were to anticipate a future use case where we would handle numbers and localization strings?)

README.md
dmtrKovalenko: I think the main point is to make library names the same (`date-fns`) and not `dateFnsAdapter`? What do you think about it, IMHO it will be more clear for users

README.md
dmtrKovalenko: I think date is even more confusing, maybe like `dateLib={DateFns}` or  `dateLibAdapter={DateFns}`

docs/pages/guides/DateAdapterProp.example.jsx
dmtrKovalenko: It is even more invalid, somewhere I thought that this kind of import will be deprecated. We are primarily using `import * as React`, but in examples using this notation. 
I have no preference and can change

README.md
oliviertassinari: I don't have a clear idea of the best approach. There is a tradeoff between the consitency of the filenames and the npm name of the dependencies. Both seem OK.

README.md
oliviertassinari: Yeah, you are right, `date` is probaby too short. A third possible option: `dateAdapter`.

docs/pages/guides/DateAdapterProp.example.jsx
oliviertassinari: There is a related thread on this topic at https://github.com/mui-org/material-ui/pull/19802.
What approach have you seen the more frequently used by TypeScript users on one side? And by JavaScript users on the other?

There are no clear guidelines by React on the matter, but I could find these two examples:
- https://reactjs.org/docs/hooks-intro.html 
- https://www.typescriptlang.org/docs/handbook/react-&-webpack.html

dmtrKovalenko: For typescript codebase, it is much easier to use `import *  as React` because a lot of internal types are also exported, like `React.ReactNode`, `React.HtmlProps` and so on. 

> My own notice: It is easier to visual recognize default react hooks like `useEffect ` or `useState` when they are used like `React.useEffect`, thus all custom hooks are easily distinguishable

dmtrKovalenko:
Should we update Readme? Because it will affect new v3.x, because next is our default branch.

oliviertassinari:
Maybe the Readme should point to the documentation websit and be kept generic?

PR #1571, opened 2020-03-11
Remove undefined from MuiPickersOverrides interface

roikoren755:
## Description
`@material-ui`'s Overrides type has each override, and each key of that override, as optional. Making the MuiPickersOverrides keys possibly undefined does not allow extending Overrides without TS errors.

This PR simply makes all the fields of MuiPickersOverrides not-undefined, allowing users to extend Overrides properly.

### Current behaviour
Using the stable version of this package, a TS error is thrown when trying to override:
`TS2344: Type 'ComponentNameToClassKey[Name]' does not satisfy the constraint 'string'.
  Type '"button" | "hidden" | "fixed" | "dashed" | "inset" | "content" | "default" | "progress" | "text" | "rtl" | "row" | "expanded" | "dense" | "last" | "absolute" | "static" | "sticky" | ... 400 more ... | undefined' is not assignable to type 'string'.
    Type 'undefined' is not assignable to type 'string'.`


dmtrKovalenko:
Please show how you are using overrides

roikoren755:
in `styled.d.ts` I have this

```
import { MuiPickersOverrides } from '@material-ui/pickers/typings/overrides';
import '@material-ui/core/styles/overrides';

declare module '@material-ui/core/styles/overrides' {
  type overridesNameToClassKey = {
    [P in keyof MuiPickersOverrides]: keyof MuiPickersOverrides[P];
  };

  export interface ComponentNameToClassKey extends overridesNameToClassKey {}
}
```

as per the documentation in `docs/typings.d.ts`

And then in a different file

```
export const muiTheme = createMuiTheme({
  overrides: {
    MuiPickersToolbar: {
      toolbar: { backgroundColor: 'blue' }
    },
  },
});
```

Running my application I see the following error from `fork-ts-checker-webpack-plugin`

```
ERROR in /Users/roikoren/development/test-repo/node_modules/@material-ui/core/styles/overrides.d.ts(101,64):
TS2344: Type 'ComponentNameToClassKey[Name]' does not satisfy the constraint 'string'.
  Type '"button" | "hidden" | "fixed" | "dashed" | "inset" | "content" | "default" | "progress" | "text" | "rtl" | "row" | "expanded" | "dense" | "last" | "absolute" | "static" | "sticky" | ... 400 more ... | undefined' is not assignable to type 'string'.
    Type 'undefined' is not assignable to type 'string'.
```

(Webstorm is showing the same error when opening the file and hovering over `ComponentNameToClassKey`).

When I use Webstorm's inspection, `overridesNameToClassKey` resolve to an equivalent of `Partial<Record<keyof MuiPickersOverrides, undefined>>`.

After my changes, there are no errors running my app, and `overridesNameToClassKey` resolves to the same thing, but with the actual class keys of each key of MuiPickersOverrides, instead of undefined.

Before the change this PR is proposing:
image

After:
image

dmtrKovalenko:
I see, but I think that we should fix the reason of the problem in v4. 
We will not make a v3 release for such small change. For now you can use `Required<PickersOverrides>`

dmtrKovalenko:
Anyway thanks for your intention to help! 👍

roikoren755:
I understand, though know that Required doesn't work in this case, as the problem is that `MuiPickersOverrides[P]` is always undefined.

I will continue with what I currently have, a simple `@ts-ignore`, until v4 is released then.
Thanks

dmtrKovalenko:
It should work if you create own type based on MuiPickersOverrides

roikoren755:
I think I understand what you mean, I will try.
Thanks again

roikoren755:
Just in case someone else runs into this, this should work until a fix is released:
```
import { MuiPickersOverrides } from '@material-ui/pickers/typings/overrides';
import '@material-ui/core/styles/overrides';

declare module '@material-ui/core/styles/overrides' {
  type overridesNameToClassKey = {
    [P in keyof Required<MuiPickersOverrides>]: keyof Required<MuiPickersOverrides>[P];
  };

  // eslint-disable-next-line @typescript-eslint/no-empty-interface
  export interface ComponentNameToClassKey extends overridesNameToClassKey {}
}
```

PR #1590, opened 2020-03-18
(chore) Better material-ui core components import strategy

oliviertassinari:
- Fix Capture d’écran 2020-03-18 à 22 11 36
- Import one level deep to speed-up bundling


lib/src/_shared/PickerToolbar.tsx
```diff
@@ -1,12 +1,14 @@
 import * as React from 'react';
 import clsx from 'clsx';
 import Toolbar, { ToolbarProps } from '@material-ui/core/Toolbar';
+import { makeStyles } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
+import IconButton from '@material-ui/core/IconButton';
+import Grid from '@material-ui/core/Grid';
```
oliviertassinari, line 7: To investigate if we can't get rid of this module. However, without the bundle size tracking tool of the core repository, it might be a bit challenging to measure progress on the impact.

dmtrKovalenko:
Wouldn’t core imports be treeshaked even when mixed?

oliviertassinari:
> Wouldn’t core imports be treeshaked even when mixed

They are.

The motivation is not to force a choice on the users regarding https://material-ui.com/guides/minimizing-bundle-size/#development-environment.

oliviertassinari:
> Code style issues found in the above file(s). Forgot to run Prettier?

How can I run Prettier (my IDE doesn't support it)?

dmtrKovalenko:
Something broken with latest `husky` release. For some reason precommit hooks don’t work even for me 🤷‍♂️

oliviertassinari:
On the core repository, we have `yarn prettier`, but I'm not sure what to use here 😁. Plus, there is no visual error output I can use to patch it by hand.

oliviertassinari:
> For some reason precommit hooks don’t work even for me

It does run on my side, I wish it didn't run actually, it's frustrating me to have to wait a couple of seconds to be able to commit.

PR #1593, opened 2020-03-19
(bugfix) Implement forwardRef for all components

### LINKED ISSUE #1303: [core] Implement forwardRef
Hi,
First of all, thank for your work !

I would like to know if is it possible to pass ref to my DatePicker ?

In my case, I use your Date Picker in a form using react-final-form.
I have wrappers for each of my inputs to customize the error message. However, I am obliged to pass a ref to each of my components in order to recover the size of said component and display a popper of this size.
Currently, from what I have seen in the doc, it is impossible for me to pass a reference to my DatePicker.

Do you have a solution to offer me or can you add the possibility to pass some?

Thank you for your time and answer !

### LINKED ISSUE #1574: 4.0.0-alpha.3 breaks onBlur for DesktopDatePicker
## Environment

| @material-ui/pickers |   4.0.0-alpha.3      |
| material-ui          |    4.9.2     |
| React                |   16.12.0      |
| Browser              |   Chrome      |
| Peer library         |  moment@2.24.0  |
| Formik                |  formik@2.0.0

## Steps to reproduce

1. For formik to consider the input as `touched` an `onBlur` callback is required.
 
## Expected behavior

The datepicker should be considered `touched` on focus loss

## Actual behavior

Nothing Happens on Desktop. Functionality works as expected for `MobileWrapper`

## Live example
This example contains 2 DatePickers, one that does not work as expected, and another that contains a "hack" to fix the issue.

https://codesandbox.io/s/formik-material-ui-date-picker-with-yup-validate-5h2v6

## Workaround
Create a custom component that overrides the `TextFieldComponent`

```
const FocusFixField = ({...other}) => {
    const form = useFormikContext();
    return <TextField onBlur={form.handleBlur} {...other} />
};

const FormikDatePicker = ({ field, form, helperText, ...other }) => {
    const currentError = form.errors[field.name];
    const touched = !!form.touched[field.name];

    return (
        <DatePicker
            clearable
            autoOk
            variant="outlined"
            name={field.name}
            value={field.value}
            format="MM/DD/YYYY"
            mask="__/__/____"
            TextFieldComponent={FocusFixField}
            helperText={currentError || helperText}
            error={touched && !!currentError}
            onChange={date => form.setFieldValue(field.name, date)}
            {...other}
        />
    );
};
```

dmtrKovalenko:
This PR closes #1303 
Also accidentally fixes #1574 

## Description
Implement `React.forwardRef` for all components.

### DIFF
--- a/lib/src/__tests__/e2e/DatePickerProps.test.tsx
+++ b/lib/src/__tests__/e2e/DatePickerProps.test.tsx
@@ -65,6 +65,7 @@ describe('DatePicker - different props', () => {
       <MobileDatePicker
         autoOk
         showTodayButton
+        cancelLabel="stream"
         onClose={onCloseMock}
         onChange={onChangeMock}
         value={utilsToUse.date('2018-01-01T00:00:00.000Z')}
@@ -76,5 +77,30 @@ describe('DatePicker - different props', () => {
 
     expect(onCloseMock).toHaveBeenCalled();
     expect(onChangeMock).toHaveBeenCalled();
+  });
+
+  it('ref - should forwardRef to text field', () => {
+    const Component = () => {
+      const ref = React.useRef<HTMLInputElement>(null);
+      const focusPicker = () => {
+        if (ref.current) {
+          ref.current.focus();
+        } else {
+          throw new Error('Ref must be available');
+        }
+      };
+
+      return (
+        <>
+          <DatePicker ref={ref} id="focusing-picker" value={null} onChange={jest.fn()} />
+          <button id="focus-picker" onClick={focusPicker} />
+        </>
+      );
+    };
+
+    const component = mount(<Component />);
+    component.find('#focus-picker').simulate('click');
+
+    expect(document.activeElement?.id).toBe('focusing-picker');
   });
 });
--- a/lib/src/__tests__/e2e/DatePickerProps.test.tsx
+++ b/lib/src/__tests__/e2e/DatePickerProps.test.tsx
@@ -84,6 +84,7 @@ describe('DatePicker - different props', () => {
       const ref = React.useRef<HTMLInputElement>(null);
       const focusPicker = () => {
         if (ref.current) {
+          ref.current.tabIndex = 1; // fix jsdom behavior https://github.com/jsdom/jsdom/blob/5e39a4c60377c95ac09ea938ecbf5f1f91cb0360/lib/jsdom/living/helpers/focusing.js#L25
           ref.current.focus();
         } else {
           throw new Error('Ref must be available');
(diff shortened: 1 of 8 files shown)


lib/src/__tests__/e2e/DatePickerProps.test.tsx
```diff
@@ -77,4 +78,29 @@ describe('DatePicker - different props', () => {
     expect(onCloseMock).toHaveBeenCalled();
     expect(onChangeMock).toHaveBeenCalled();
   });
+
+  it('ref - should forwardRef to text field', () => {
```
oliviertassinari, line 83: We have started to normalize the common expectations from the components in https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/test-utils/describeConformance.js. It might help in the future.

oliviertassinari:
@dmtrKovalenko Do you plan to implement forwardRef for all the public components? From what I understand, we now have the following covered:

- StaticTimePicker
- MobileTimePicker
- DesktopTimePicker
- TimePicker
- DatePicker
- MobileDatePicker
- StaticDatePicker
- DekstopDatePicker
- DateTimePicker
- MobileDateTimePicker
- StaticDateTimePicker
- DekstopDateTimePicker

but we miss:

- Calendar
- ClockView

Is this accurate?

dmtrKovalenko:
@oliviertassinari  probably yes, but where should we forward ref for calendar? Into the root element? Is it really needed? 
Because we have no input for them, as well for new public `<Day />` component

lib/src/_shared/KeyboardDateInput.tsx
```diff
@@ -91,6 +93,8 @@ export const KeyboardDateInput: React.FC<DateInputProps> = ({
 
   const adornmentPosition = InputAdornmentProps?.position || 'end';
   const inputProps = {
+    ref: containerRef,
+    inputRef: forwardedRef,
```
oliviertassinari, line 97: Something is off with the forwardRef logic, the standard we went for is to apply it on the root element. Here, it seems to be forwarded on an arbitrary deeper element: the text box.

oliviertassinari:
> Into the root element? Is it really needed?

Yes, it's a convention all the components have, the aim is to behave like a built-in element. It will also very likely become the default behavior in a future version of React. Facebook has a RFC about treating the `ref` prop like any other props. So it would spread on the same element where the other props spread, by convention, it's the root element.

lib/src/_shared/KeyboardDateInput.tsx
oliviertassinari: This sounds like a good reason to start using https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/test-utils/describeConformance.js in the tests, it will increase the consistency of the components.

eduardoribeiro17:
hey guys! 
when will this be finished? I have the same problem in use...

oliviertassinari:
@eduardoribeiro17 Please open an issue in https://github.com/mui-org/material-ui with a reproduction on v5.

PR #1602, opened 2020-03-24
[DateRangePicker] Introduce new component  🎉

### LINKED ISSUE #364: Feature Request: Ability to select a date range
Thank you for this amazing component. Your work and contributions are very much appreciated.

It would be amazing if this component could also work to select a date range (similar to https://ipiz.herokuapp.com/md-date-range-picker-demo/index.html).

### LINKED ISSUE #1306: Using memo with day components
I skip questions since I don't find them relevant, the only thing I need is a way to memoize day components which can extremely improve performance since it seems all days are getting re-rendered on every date change

### LINKED ISSUE #1420: [question] can a InvalidDateMessage override helperText?
Hi,

Is there any way to override the "helperText" when there is an invalid date? Because when I don't have a "helperText" and the user inserts an invalid format the "invalidDateMessage" appears. But when I add the "helpText" the "invalidDateMessage" never appears.

```
<KeyboardTimePicker
        className="timepicker-new-event"
        name={field.name}
        value={field.value}
        ampm={false}
        variant="dialog"
        label="Hours"
        views={["hours", "minutes"]}
        keyboardIcon={null}
        onChange={date => form.setFieldValue(field.name, date, true)}
        invalidDateMessage="Invalid Format"
        helperText="XPTO"
/>
```

Hope you can help me. Thank you in advance :)

### LINKED ISSUE #1603: Crash without ThemeProvider
> TypeError: Cannot read property 'breakpoints' of null

The problems originate from this line

https://github.com/mui-org/material-ui-pickers/blob/0d41c48aaa8b9b6f67e72deee9583340cd218b89/lib/src/wrappers/ResponsiveWrapper.tsx#L27

For more details: https://github.com/mui-org/material-ui/issues/16859

## Environment

| @material-ui/pickers | v4.0.0-alpha.4 |

## Steps to reproduce

https://codesandbox.io/s/cranky-hamilton-130yq

dmtrKovalenko:
This PR closes #364  

Closes #1306
Closes #1603 
Closes #1420 (but needs more documentation) 

## Description
Introduce new `DateRangePicker` component. Link to the documentation of deploy preview (https://material-ui-pickers-git-feature-daterangepicker.mui-org.now.sh/demo/daterangepicker)

DateRangePickerPreview


oliviertassinari (commented):
One step closer to this major component. Two years in the making 06498fcba1e801895f684a15b051306b865e51bf :D.

Here is an early feedback around the `ClickAwayListener` concern you have raised. I have went a bit offroad, sorry 🙃:

- It might be better to work on a fork, I believe the creation of branches can confuse developers that look for older versions. Basically, master, next and v3-x only as branches would reduce noise.
- We might want to trap the focus when it lands inside the Popper, `<TrapFocus disableEnforceFocus disableAutoFocus />`.
- `dataset` I didn't know this property existed, I have learned something. Unfortunately, applying muiPickersInput to the input isn't enough. If you click on the edge of the input, you can witness a quick open (focus) and close (click away) behavior.
- I would suggest replacing the usage of ClickAwayListener with a blur based approach. Basically, if the focus leaves the widget, we can close it. This way, we get keyboard + click handled at once. Two birds with one stone. I have used this approach with the Autocomplete.
- For the API, what do you think of an [Autocomplete](https://material-ui.com/components/autocomplete/) like approach?

```jsx
    <DateRangePicker
        renderInput={(firstInput, secondInput) => (
            <>
                <TextField {...firstInput} />}
                {' to '}
                <TextField {...secondInput} />}
            </>
        )}
    />
```
  It would empower a simpler composition.
- Switching months feels a bit slow, I measure 300ms on my high-end laptop. This resonates with a past discussion we had regarding prioritizing an investigation in the runtime performance. It sounds to me like a confirmation that we will need to look into it (later) :).

dmtrKovalenko:
Hahaha september of 2018 XD 

Regarding to click away - data attribute is temporary solution, now working on different structure
```
<TrapFocus>
  <ClickAwayListener>  
    <div>
      <Input1 />
      <Input2 /> 

      <Popper disablePortal> 
           <Calendar />
      </Popper>
    </div> 
  <ClickAwayListener> 
<TrapFocus> 
```

This should fix all the issues

oliviertassinari:
> Regarding to click away - data attribute is temporary solution, now working on different structure

@dmtrKovalenko I fear the `disablePortal` will prevent the usage of the date range picker in fairly common use cases, for instance, inside a Dialog. Luckily, we might have a good solution for it in https://github.com/mui-org/material-ui/issues/18586#issuecomment-562183410.
But my main concern is about: what happens with the keyboard handling? How do we detect that we should close the popup? I imagine that we will need to listen to the blur event (as of now, the popup is open on focus, but not close when moving to a different field with <kbd>Tab</kbd>). Once the keyboard logic is in place, would we still need ClickAwayListener? I suspect we won't, at least from my past experience, but to explore.

dmtrKovalenko:
We need to make possible clicking on input without closing popup. 
In order to switch between start and end selection

dmtrKovalenko:
TODO fix displaying and interaction with range in safari

oliviertassinari:
@dmtrKovalenko Regarding asking for feedback on the API, a few thoughts :)

1. I wonder about adding the support for `defaultValue`, hence making `value` not required.
1. Given that we have already taught users how to use the `onChange` prop, I think that a non required onChange prop would be better (e.g. for disabled or read-only values).
1. You might want to add a dot at the end of the props' descriptions. It's a convention we enforce in the mono-repository.
1. What about we rename the localization strings to match the "text" convention: `cancelLabel -> cancelText`, `clearLabel -> clearText`, `todayLabel  -> todayText`, `invalidDateMessage -> invalidDateText`, `minDateMessage  -> invalidMinDateText `, `maxDateMessage  -> invalidMaxDateText `, and `okLabel -> okText`.
1. Mobile wrapper might miss a default value for `clearable`.
1. ~I would add `event` as the first argument to all the handlers, per convention with the core components: `onChange`, `onAccept`, `onClose`, `onError`, `onMonthChange`, and `onOpen`.~ After further internal discussion, this is the result of a tradeoff. We haven't found a consensus, we can wait for further users' feedback before going in a specific direction.
1. Mobile wrapper might miss a default value for `showTodayButton`.
1. I think that setting the default text values in lower case `CLEAR -> Clear` (and using CSS to uppercase them) could be interesting for consistency with the core components.
1. I wonder if `emptyInputText` couldn't be replaced with another prop: `inputFormat`, by accepting a function.
1. Is `PopoverProps` still relevant?
1. Did you advance the reflection on the alternative render APIs (to increase composition)?

```jsx
<DateRangePicker 
  renderFromInput={props => <TextField {...props} />}
  delimiter="to"
  renderToInput={props => <TextField {...props} />}
/> 
```

or 

```jsx
<DateRangePicker
    renderInput={(firstInput, secondInput) => (
        <>
            <TextField {...firstInput} />}
            <Box m={2}>{' to '}</Box>
            <TextField {...secondInput} />}
        </>
    )}
/>
```

lib/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }
+    if (!disableOpenPicker) {
+      openPicker();
+    }
+  };
+
+  return (
+    <div id={id} className={clsx(classes.rangeInputsContainer, className)} ref={containerRef}>
+      <KeyboardDateInput
+        {...other}
```
oliviertassinari:
```jsx
    <div id={id} className={clsx(classes.rangeInputsContainer, className)} ref={containerRef} {...other}>
      <KeyboardDateInput
```

docs/pages/demo/daterangepicker/BasicDateRangePicker.example.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
```
oliviertassinari, line 1:
```suggestion
import React from 'react';
```

docs/pages/demo/daterangepicker/CalendarsDateRangePicker.example.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { Grid, Typography } from '@material-ui/core';
```
oliviertassinari:
```suggestion
import Grid from '@material-ui/core/Grid';
import Typography from '@material-ui/core/Typography';
```

docs/pages/demo/daterangepicker/MinMaxDateRangePicker.example.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import addWeeks from 'date-fns/addWeeks';
+import { Dayjs } from 'dayjs';
+import { Moment } from 'moment';
+import { DateTime } from 'luxon';
+import { makeJSDateObject } from '../../../utils/helpers';
```
oliviertassinari, line 6: It won't work in people's codebases.

lib/src/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,149 @@
...
+export function makeRangePicker<TWrapper extends SomeWrapper>(Wrapper: TWrapper) {
+  const WrapperComponent = makeWrapperComponent<DateRangeInputProps, RangeInput, DateRange>(
+    Wrapper,
+    {
+      KeyboardDateInputComponent: DateRangePickerInput,
+      PureDateInputComponent: DateRangePickerInput,
+    }
+  );
+
+  function RangePickerWithStateAndWrapper({
```
oliviertassinari, line 54: Should it forward ref here?

lib/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+      display: 'flex',
+      alignItems: 'center',
+      // ? TBD
+      [theme.breakpoints.down('xs')]: {
+        flexDirection: 'column',
+      },
+    },
+    toLabelDelimiter: {
+      margin: '0 16px',
+      [theme.breakpoints.down('xs')]: {
```
oliviertassinari: We use `theme.breakpoints.up()` as much as possible in the codebase, up('sm') in our case.

lib/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+    <div id={id} className={clsx(classes.rangeInputsContainer, className)} ref={containerRef}>
+      <KeyboardDateInput
+        {...other}
+        open={open}
+        forwardedRef={startRef}
+        rawValue={start}
+        parsedDateValue={start}
+        onChange={handleStartChange}
+        disableOpenPicker
+        openPicker={() => {}}
```
oliviertassinari: Dead code?

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+    <PickerToolbar
+      toolbarTitle={toolbarTitle}
+      isMobileKeyboardViewOpen={isMobileKeyboardViewOpen}
+      toggleMobileKeyboardView={toggleMobileKeyboardView}
+      isLandscape={false}
+      penIconClassName={classes.penIcon}
+    >
+      <div className={classes.dateTextContainer}>
+        <ToolbarButton
+          variant={Boolean(start) ? 'h5' : 'h6'}
```
oliviertassinari, line 68: Do we need to cast the variable to a boolean?

```suggestion
          variant={start ? 'h5' : 'h6'}
```

lib/src/DateRangePicker/DateRangePickerView.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+    if (
+      (currentlySelectingRangeEnd === 'start' && start === null) ||
+      (currentlySelectingRangeEnd === 'end' && end === null)
+    ) {
+      return;
+    }
+
+    scrollToDayIfNeeded(currentlySelectingRangeEnd === 'start' ? start : end);
+  }, [currentlySelectingRangeEnd, date]); // eslint-disable-line
```
oliviertassinari, line 128:
```suggestion
  }, [currentlySelectingRangeEnd, date]);
```

lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -0,0 +1,176 @@
...
+    changeMonth(utils.getNextMonth(currentMonth));
+  }, [changeMonth, currentMonth, utils]);
+
+  const selectPreviousMonth = React.useCallback(() => {
+    changeMonth(utils.getPreviousMonth(currentMonth));
+  }, [changeMonth, currentMonth, utils]);
+
+  return (
+    <div className={classes.dateRangeContainer}>
+      {new Array(calendars).fill(0).map((_, index) => {
```
oliviertassinari: IE 11?

lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -0,0 +1,176 @@
...
+  rightArrowButtonProps,
+  rightArrowButtonText,
+  rightArrowIcon,
+  onChange,
+  disableFuture,
+  disablePast,
+  minDate,
+  maxDate,
+  currentlySelectingRangeEnd,
+  ...CalendarProps
```
oliviertassinari:
```suggestion
  ...other
```

lib/src/__tests__/setup.js
```diff
@@ -4,6 +4,17 @@ const EnzymeAdapter = require('enzyme-adapter-react-16');
 // Setup enzyme's react adapter
 Enzyme.configure({ adapter: new EnzymeAdapter() });
 
+// Fix popper mount https://github.com/mui-org/material-ui/issues/15726
+document.createRange = () => ({
```
oliviertassinari, line 9: It should no longer be required: https://github.com/mui-org/material-ui/issues/15726#issuecomment-610084883.

lib/src/_shared/icons/ArrowDropDownIcon.tsx
```diff
@@ -1,4 +1,4 @@
 import React from 'react';
-import createSvgIcon from '@material-ui/core/internal/svg-icons/createSvgIcon';
+import createSvgIcon from '@material-ui/core/utils/createSvgIcon';
```
oliviertassinari:
```suggestion
import { createSvgIcon } from '@material-ui/core/utils';
```

I wish it was solved in the different pull request!

lib/src/views/Calendar/CalendarView.tsx
```diff
@@ -146,6 +72,9 @@ export const useStyles = makeStyles(
   { name: 'MuiPickersCalendarView' }
 );
 
+export const defaultReduceAnimations =
+  typeof window !== 'undefined' && /(android)/i.test(window.navigator.userAgent);
```
oliviertassinari: For comparison in the core

```jsx
const disableSwipeToOpenDefault =
  typeof navigator !== 'undefined' && /iPad|iPhone|iPod/.test(navigator.userAgent);

```
```suggestion
  typeof navigator !== 'undefined' && /(android)/i.test(navigator.userAgent);
```

oliviertassinari (commented):
I have made a couple of suggestions. I couldn't deep dive into the code, they are too many changes. Also, the review experience on GitHub is horrible. I'm happy to provide more details on my comment where needed.

lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
dmtrKovalenko: this spreading rename makes code more readable and strict, because we know that there should be only `Calendar` props

lib/src/__tests__/setup.js
dmtrKovalenko: But it doesn't work without :)

docs/pages/demo/daterangepicker/MinMaxDateRangePicker.example.tsx
dmtrKovalenko: I know, but only one thing we can do without dramatically changes in examples - add warning comment

docs/pages/demo/daterangepicker/BasicDateRangePicker.example.tsx
dmtrKovalenko: `import React from 'react'` - actually is a weird workaround coming from interop between UMD and esm/cjs. We should not use default export from `React`

lib/src/DateRangePicker/DateRangePickerInput.tsx
dmtrKovalenko: No, it means that we are doing nothing on this step. Probably need to make a function `doNothing` to avoid creating a new function

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
dmtrKovalenko: My point: implicit types castings are the root of evil

lib/src/DateRangePicker/DateRangePickerView.tsx
dmtrKovalenko: No, because we need to react only on this 2 props change

lib/src/__tests__/setup.js
oliviertassinari: Meaning you want to upgrade jsdom in a different pull request? Great call. In this case, I would suggest a simple comment: 
```suggestion
// TODO, upgrade jsdom to 16.0.0 and remove this line.
document.createRange = () => ({
```

lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
oliviertassinari: In this case, I propose we introduce a new prop `CalendarProps`. So we can keep `other` spreading on the root element, per convention.

lib/src/DateRangePicker/DateRangePickerInput.tsx
oliviertassinari: What do you think of making `openPicker` optional? Why having it required? Do we have here an example that it's not always the expected behavior?

docs/pages/demo/daterangepicker/MinMaxDateRangePicker.example.tsx
oliviertassinari: Actually, it won't be the only issue. Doesn't `DateRangePicker` has a required context that we should problem?

I think that we will need to have a dedicated story for this. It's important all the demos can be copied and paste without any exceptions. It hasn't always been the case in the core repository, it has been painful for developers.

docs/pages/demo/daterangepicker/BasicDateRangePicker.example.tsx
oliviertassinari: We have a thread on this topic at https://github.com/mui-org/material-ui/pull/19802. Ok, it's not important, let's ignore it, it will be solved with the repository migration anyway.

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
oliviertassinari: It seems simple enough here 🤔, what would be the downside?

lib/src/DateRangePicker/DateRangePickerView.tsx
oliviertassinari: > we need to react only on this 2 props change

I used to believe the same. Experience has shown me that I was wrong. Anytime I tried such an approach, somebody came up with a bug including all the dependencies solves. I would respect if you have reservations about this point. I wouldn't mind actually, I'm confident you will change your mind with the bug reports that will come in :). So, happy to solve that later.

docs/pages/demo/daterangepicker/MinMaxDateRangePicker.example.tsx
dmtrKovalenko: I mean, this is require to interop with different libraries. 
Ideally we need separate demo for each lib (maybe somehow generated as we are doing for formats)

lib/src/DateRangePicker/DateRangePickerInput.tsx
dmtrKovalenko: Actually from the strong typed system perspective. 
This prop is required. But in this particular context — handled on the level higher (and it is expected, because we are doing wrapper above 2 inputs that changing their behavior).

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
dmtrKovalenko: Maybe it is simple here. But what if an empty string will be able to pass to the ‘start’?
This approach makes it easier to understand that you are doing unsafe in JavaScript operation. 

And again in strongly typed languages like Haskell, OCaml, Rust, C# and many other. It’s just a story about strictness. 

But only one thing I’m worrying about — bundlesize :(

lib/src/DateRangePicker/DateRangePickerView.tsx
dmtrKovalenko: This hook is running month switching animation. I suppose nobody will want to run month switching outside of switching month. 

But I agree everything possible. Thanks, let’s wait for reports. 😊

oliviertassinari: I could be wrong too 😁

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
oliviertassinari: What's the approach used in the main repository? If it already works well for us, I would stick to it.

dmtrKovalenko: Yeah, it is an unpopular way to handle things. But it doesn't mean that everybody can't be wrong 
image

lib/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+  };
+
+  return (
+    <div
+      id={id}
+      className={clsx(classes.rangeInputsContainer, className)}
+      ref={mergeRefs([containerRef, forwardedRef])}
+    >
+      <KeyboardDateInput
```
DDDDDanica, line 120: [Suggestion]: You might want to provide an option to disabled auto fill from browser.

oliviertassinari: Do you have more details on the potential problem and solution? Thanks.

DDDDDanica: So in the root input field component, just need to disable autoComplete :P
```
 <InputField autoComplete="off" ... />
```
:)

dmtrKovalenko: You can pass a prop `inputProps={{ autoComplete: "off" }}` right to the component. But this API will be changed to `renderInput={props => <TextField {...props} />}`

oliviertassinari: What should be the default behavior? Displaying the browser's suggestion on top of the popup could be an issue but we won't display the popup by default as in https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html. I was proposing two optional props `openOnFocus` and `openOnClick` to have the popup open by default. In such case, should we change the default autocomplete? I'm not sure. We do change is with the Autocomplete component, and had no people complaining, but still?

I'm in favor of keeping the current tradeoff.

DDDDDanica: The reason behind suggesting it being off is one scenario we encountered before:
When we have both suggested auto fill & calendar deployed(opened), it was not a very good user experience to keep both.
As mentioned, it is an optional change. I'm happy to see how the current API goes and maybe the autoOn is a more favoring solution :P

Issue #1648, opened 2020-04-12
[DatePicker] Update to match the specification (desktop)


oliviertassinari:
## Environment

| Tech                 | Version |
| -------------------- | ------- |
| @material-ui/pickers | v4.0.0-alpha.5  |

## Steps to reproduce

https://next.material-ui-pickers.dev/demo/datepicker#basic-usage

## Actual behavior

<img width="383" alt="Capture d’écran 2020-04-13 à 00 11 25" src="https://user-images.githubusercontent.com/3165635/79081186-557bff00-7d1b-11ea-8f0a-616324205d1a.png">

## Expected behavior

<img width="421" alt="Capture d’écran 2020-04-13 à 00 07 00" src="https://user-images.githubusercontent.com/3165635/79081168-29607e00-7d1b-11ea-8546-efea151e1566.png">

1. The component should have a higher density on desktop. It should be 258x281 instead of 320x358.
2. The popper should be positioned to start on the right side, not at the center.
3. The day chip should be 28x28 (maybe less, not sure) instead of 36x36 pixels.
4. The year chip should be 54x28 instead of 72x36 pixels.
5. The colors of the range don't match. Try to set the same primary color as in material.io `#7001f3` to get an idea of the issue. I propose this diff:

```diff
diff --git a/lib/src/DateRangePicker/DateRangePickerDay.tsx b/lib/src/DateRangePicker/DateRangePickerDay.tsx
index 9f54cb31..bff27496 100644
--- a/lib/src/DateRangePicker/DateRangePickerDay.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerDay.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import clsx from 'clsx';
 import { DAY_MARGIN } from '../constants/dimensions';
 import { useUtils } from '../_shared/hooks/useUtils';
-import { makeStyles, fade } from '@material-ui/core/styles';
+import { makeStyles, fade, lighten } from '@material-ui/core/styles';
 import { Day, DayProps, areDayPropsEqual } from '../views/Calendar/Day';

 interface DateRangeDayProps extends DayProps {
@@ -38,8 +38,8 @@ const useStyles = makeStyles(
     },
     rangeIntervalDayHighlight: {
       borderRadius: 0,
-      color: theme.palette.primary.contrastText,
-      backgroundColor: fade(theme.palette.primary.light, 0.6),
[...]

PR #1653, opened 2020-04-13
Better recognition of touch devices

### LINKED ISSUE #1649: The responsive logic triggers too early for mobile
## Environment

## Actual behavior

https://github.com/mui/material-ui-pickers/blob/7bed283699ef768936a3ec7c5dc89571492dddd4/lib/src/wrappers/ResponsiveWrapper.tsx#L24

## Expected behavior

The default value of the breakpoint should be `sm`. It's what we use in the mono-repository both for [CSS](https://github.com/mui-org/material-ui/blob/8e2f0cc9e5696952566ce97fe4ac5aa3a64ee2f4/packages/material-ui/src/Tooltip/Tooltip.js#L141) and for [JavaScript](https://material-ui.com/components/dialogs/#responsive-full-screen).

### LINKED ISSUE #1664: Strange black border in the calendar
## Steps to reproduce

- I open the link in question
- Click on input Mobile start
- The calendar opens as shown

## Live example

[v4-0-0-alpha-5](https://v4-0-0-alpha-5.material-ui-pickers.dev/demo/daterangepicker#responsiveness)

image

dmtrKovalenko:
This PR closes #1649 
Also closes #1664

## Description
Use `media (pointer: fine)` (https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer) which is the best way to determine touch devices. So use it to render mobile wrapper and hide focus outline. 

> P.S. It is not supported in IE but it is fine because we are falling back to desktop mode :)

### DIFF
--- a/lib/src/wrappers/ResponsiveWrapper.tsx
+++ b/lib/src/wrappers/ResponsiveWrapper.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import useMediaQuery from '@material-ui/core/useMediaQuery';
-import { useTheme } from '@material-ui/core/styles';
 import { MobileWrapperProps, MobileWrapper } from './MobileWrapper';
 import { DesktopWrapperProps, DesktopWrapper } from './DesktopWrapper';
 import { Breakpoint } from '@material-ui/core/styles/createBreakpoints';
@@ -10,10 +9,11 @@ export interface ResponsiveWrapperProps
   extends DesktopWrapperProps,
     DesktopPopperWrapperProps,
     MobileWrapperProps {
-  /** Breakpoint when `Mobile` mode will be changed to `Desktop`
-   * @default 'sm'
+  /** Css media query when `Mobile` mode will be changed to `Desktop`
+   * @default "(hover: hover)"
+   * @example "(min-width: 720px)" or theme.breakpoints.up("sm")
    */
-  desktopModeBreakpoint?: Breakpoint;
+  desktopModeMediaQuery?: Breakpoint;
 }
 
 export const makeResponsiveWrapper = (
@@ -21,7 +21,7 @@ export const makeResponsiveWrapper = (
   MobileWrapperComponent: React.FC<MobileWrapperProps>
 ) => {
   const ResponsiveWrapper: React.FC<ResponsiveWrapperProps> = ({
-    desktopModeBreakpoint = 'sm',
+    desktopModeMediaQuery = '(hover: hover)',
     okLabel,
     cancelLabel,
     clearLabel,
@@ -35,8 +35,7 @@ export const makeResponsiveWrapper = (
     displayStaticWrapperAs,
     ...other
   }) => {
-    const theme = useTheme();
-    const isDesktop = useMediaQuery(theme.breakpoints.up(desktopModeBreakpoint));
+    const isDesktop = useMediaQuery(desktopModeMediaQuery);
 
     return isDesktop ? (
       <DesktopWrapperComponent
--- a/lib/src/wrappers/ResponsiveWrapper.tsx
+++ b/lib/src/wrappers/ResponsiveWrapper.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import useMediaQuery from '@material-ui/core/useMediaQuery';
 import { MobileWrapperProps, MobileWrapper } from './MobileWrapper';
 import { DesktopWrapperProps, DesktopWrapper } from './DesktopWrapper';
-import { Breakpoint } from '@material-ui/core/styles/createBreakpoints';
 import { DesktopPopperWrapperProps, DesktopPopperWrapper } from './DesktopPopperWrapper';
 
 export interface ResponsiveWrapperProps
@@ -13,7 +12,7 @@ export interface ResponsiveWrapperProps
    * @default "(hover: hover)"
    * @example "(min-width: 720px)" or theme.breakpoints.up("sm")
    */
-  desktopModeMediaQuery?: Breakpoint;
+  desktopModeMediaQuery?: string;
 }
 
 export const makeResponsiveWrapper = (
--- a/lib/src/wrappers/ResponsiveWrapper.tsx
+++ b/lib/src/wrappers/ResponsiveWrapper.tsx
@@ -10,8 +10,8 @@ export interface ResponsiveWrapperProps
   extends DesktopWrapperProps,
     DesktopPopperWrapperProps,
     MobileWrapperProps {
-  /** Breakpoint when `Desktop` mode will be changed to `Mobile`
-   * @default 'md'
+  /** Breakpoint when `Mobile` mode will be changed to `Desktop`
+   * @default 'sm'
    */
   desktopModeBreakpoint?: Breakpoint;
 }
@@ -21,7 +21,7 @@ export const makeResponsiveWrapper = (
   MobileWrapperComponent: React.FC<MobileWrapperProps>
 ) => {
   const ResponsiveWrapper: React.FC<ResponsiveWrapperProps> = ({
-    desktopModeBreakpoint = 'md',
+    desktopModeBreakpoint = 'sm',
     okLabel,
     cancelLabel,
     clearLabel,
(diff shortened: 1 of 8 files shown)


lib/src/wrappers/ResponsiveWrapper.tsx
```diff
@@ -10,8 +10,8 @@ export interface ResponsiveWrapperProps
   extends DesktopWrapperProps,
     DesktopPopperWrapperProps,
     MobileWrapperProps {
-  /** Breakpoint when `Desktop` mode will be changed to `Mobile`
-   * @default 'md'
+  /** Breakpoint when `Mobile` mode will be changed to `Desktop`
+   * @default 'sm'
    */
```
oliviertassinari:
```suggestion
  /**
   *  The desktop version will be shown instead of the mobile one for any screen width above this breakpoint value.
   * @default 'sm'
   */
```

oliviertassinari (commented):
I wonder about a potential inconsistency this prop can bring. In the CSS we will use `sm`, with no way to have it configured. Do we need to expose a `desktopModeBreakpoint` prop? What if it was hard coded?

dmtrKovalenko:
What do you mean by? For example somebody can decide to show the mobile picker for tablets
> In the CSS we will use sm, with no way to have it configured

oliviertassinari:
Ok, so we have two different concerns:

- **Logic consistency**. Would it be weird to have a mobile version with small day cells or a desktop version with large day cells? Should we make sure that the CSS media query match the version of the component? 
- **Threshold**. What's the tablet's screen width? What version should we display for a tablet? 

It seems that it's should be the touch vs mouse pointer support that should determine what version to display. Maybe we are no looking at the right media query. What if we were using `@media (hover: none)` instead of the screen width? The more I think about it, the more I like this approach. However, I have seen cases (in Galaxy phones where it doesn't work, there is an issue in the core). So I think that we would also need to add a condition for small screens.

dmtrKovalenko:
@oliviertassinari really, it is the best way to recognize touch devices. It works really nice for our case. Thanks for this idea! 💪

dmtrKovalenko:
Todo figure out visual regression

oliviertassinari:
Regarding the limitation with Samsung phones, here is it: https://github.com/mui-org/material-ui/pull/15000. You can use BrowserStack to try that out.
I believe something like this `@media (hover: none), (max-width: 600px)` would allow to cover most of the cases.

dmtrKovalenko:
@oliviertassinari also I want to handle the case when the user is on desktop, but the window is small (e.g. iframes or just resized window) 
image

Maybe we can make a hook `isTouchDevice` and place it in `@materail-ui/core/utils` with all the required logic to recognize touch device?

oliviertassinari:
> Maybe we can make a hook isTouchDevice and place it in @materail-ui/core/utils with all the required logic to recognize touch device?

@dmtrKovalenko Given that we only have a use case, for now, in the date picker, I think that we should keep the solution scoped to the date picker. Once 1. we find a solution, 2. that we stress test with the users of the date picker, 3. that we see a need outside the date picker, then, yes, definitely, I think that it would make a great addition to the /core or /utils package.

oliviertassinari:
This approach seems to do it.

```diff
   const ResponsiveWrapper: React.FC<ResponsiveWrapperProps> = ({
-    desktopModeMediaQuery = '(hover: hover)',
+    desktopModeMediaQuery = '@media (pointer: fine)',
     okLabel,
     cancelLabel,
     clearLabel,
```
*I have included @media to ease pattern search in the repository (between JS and CSS)*

The global document listener approach would be challenging if we need to wait for a first pointer interaction: https://stackoverflow.com/questions/23885255/how-to-remove-ignore-hover-css-style-on-touch-devices

dmtrKovalenko:
Pointer coarse doesn’t work. You can check it with current outline styles for mobile. They appearing on real devices.

oliviertassinari:
*Sorry, outdated comment, please check again with "fine", it seemed to work on BrowserStack in all the configurations.*

dmtrKovalenko:
Confirmed for galaxy s* `(pointer: fine)` is working just fine 
image

lib/src/wrappers/ResponsiveWrapper.tsx
```diff
@@ -1,27 +1,27 @@
...
 
 export interface ResponsiveWrapperProps
   extends DesktopWrapperProps,
     DesktopPopperWrapperProps,
     MobileWrapperProps {
-  /** Breakpoint when `Desktop` mode will be changed to `Mobile`
-   * @default 'md'
+  /** Css media query when `Mobile` mode will be changed to `Desktop`
+   * @default "@media (pointer: fine)"
+   * @example "@media (min-width: 720px)" or theme.breakpoints.up("sm")
```
oliviertassinari, line 14: What do you think of this variation to better match the style of the mono-repository for the description of the props?

```suggestion
  /**
   * CSS media query when "mobile" mode will be changed to "desktop".
   * 
   * @default "@media (pointer: fine)"
   * @example "@media (min-width: 720px)" or theme.breakpoints.up("sm")
```

lib/src/wrappers/ResponsiveWrapper.tsx
```diff
@@ -1,27 +1,27 @@
...
    */
-  desktopModeBreakpoint?: Breakpoint;
+  desktopModeMediaQuery?: string;
```
oliviertassinari, line 16: Have you considered a shorter name for the prop?

```suggestion
  desktopMediaQuery?: string;
```

oliviertassinari (commented):
Awesome. I'm excited, I can't wait to try it out in real life (publish) and the implication for updating the styles for desktop #1648 :)

oliviertassinari:
> Also closes #1664

Are you sure about this one? Yes, it solves the issue on mobile, but what about desktop?

dmtrKovalenko:
Yeah, because it needs for desktop a11y. We need to render outline for the focused element. (This is a problem of trap focus actually😮)

And it is not a problem for desktop, because we are not doing missclicks thanks to mouse. On desktop you will probably never see it if not navigating through the Tab. But on mobile it is appearing all the time.

oliviertassinari:
I have added an extra description on #1664 to describe what I have observed.

lib/src/wrappers/ResponsiveWrapper.tsx
dmtrKovalenko: I think the shorter variant is less expressive. What "desktop" means?

lib/src/wrappers/ResponsiveWrapper.tsx
dmtrKovalenko: `Mobile` and `Desktop` are used as prefixes for specific components (e.g. `MobileDatePicker`) so it should be more clear for users. Isn't it?

dmtrKovalenko: `Mobile` and `Desktop` are used as prefixes for specific components (e.g. `MobileDatePicker`) so it should be more clear for users. Isn't it?

lib/src/wrappers/ResponsiveWrapper.tsx
oliviertassinari: It could be interpreted as "a media query for desktop", definitely not as good as the information we can provide in a description. Adding mode definitely provide extra context. I don't know. I was wondering about this tradeoff of information vs length for the props, e.g. `openPickerIcon`, vs `pickerIcon`.

oliviertassinari (approved):

lib/src/wrappers/ResponsiveWrapper.tsx
oliviertassinari: Using `Mobile` feels like it's a piece of code that can be used on its own. What about 1. `MobileDatePicker` or 2. "mobile", I think that we would lose value if we try to compromise between these two options.

spider9375:
@oliviertassinari sorry for bringing this up, but i'm confused. The documentation states: 

[fine](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer#fine)
The primary input mechanism includes an accurate pointing device, such as a **mouse**.

This means that this media query matches desktop, and not mobile devices.

And at the same time the name of the constant is
 IS_TOUCH_DEVICE_MEDIA which when true means that it is in fact not touch device and instead one with a mouse? Or am i understanding this wrong?

PR #1661, opened 2020-04-15
Improve input mask UX and a11y

### LINKED ISSUE #1144: Can no longer access `keepCharPositions` parameter in version 3.1.0
Hi there,

Thanks for your work on this great library!

I've been trying to set the `keepCharPositions` flag for the underlying text field. I followed the solution in [this issues](https://github.com/mui-org/material-ui-pickers/issues/505), but it doesn't seem to be working for `@material-ui/pickers` version `3.1.0`.

For example, in the following code keepCharPositions does not take effect:

```
    <KeyboardDatePicker
      autoOk={true}
      variant="inline"
      label={"Test"}
      format={"dd/MM/yyyy"}
      inputValue={"12/12/1900"}
      value={null}
      InputAdornmentProps={{ position: "end" }}
      placeholder={"__/__/____"}
      onChange={(_date, value) => {
        console.log(value);
      }}
      InputProps={{
        inputProps: {
          keepCharPositions: true
        }
      }}
    />
```

Any ideas how to use `keepCharPositions` with the latest version?

### LINKED ISSUE #1352: Keyboard picker with dayjs does not allow last minute digit to be typed in
## Environment

| @material-ui/pickers |   3.2.6      |
| material-ui          |   4.5.1      |
| React                |    16.8.6     |
| Browser              |    Chrome     |
| Peer library         |   dayjs 1.8.16      |  |

## Steps to reproduce

1. Enter a valid date and time up to the last minute digit (e.g. 10192019 080). Note the last 0 was not entered on the keyboard.
2. The last 0 is entered automatically and it is very difficult to change it to something else.

## Expected behavior

The date should not be valid or at least the component should allow the user to type in the last digit of the minutes without it being done for them.

## Actual behavior

The date is treated as valid without the user entering the last digit which causes onChange to fire with a valid dayjs object. This then re-renders the component with 0800. If you backspace the last 0 it puts it back in. You have to select the last 0 to change it to something else.

## Live example

https://codesandbox.io/s/mui-pickers-dayjs-keyboard-issue-kw74k

### LINKED ISSUE #1409: Change the display "__/__/____" when entering date for KeyboardDatePicker
I'm trying to find a way to keep the placeholder on display when user start typing the date. So instead of showing something like "1_/__/____" I want it to be like "1D / MM / YYYY"

### LINKED ISSUE #1660: DateRangePicker do not display result in inputs for luxon adapter
## Environment

| @material-ui/pickers | 4.0.0-alpha5        |
| material-ui          |         |
| React                |         |
| Peer library         | luxon@1.23.0         |  |

## Steps to reproduce

Open live [example](https://codesandbox.io/s/determined-star-yb0hq?file=/index.js) & try to change luxon date range
Result is applied but not displayed in inputs

## Expected behavior

Display selected range in inputs

## Actual behavior

No visible result for luxon only n

## Live example

https://codesandbox.io/s/determined-star-yb0hq?file=/index.js

dmtrKovalenko:
Linked issues:
* closes #1660 
* closes #1409
* closes #1352
* closes #1144 

## Breaking changes
* Remove `maskChar` prop, because it is not visible for users anymore and need only to generate and parse mask pattern.

## Features 
Change the logic of the input mask to be more accessible and less noisy. 

1. It is not rendering the whole mask from now (12/1_/____) for partial input it will show only typed part and the next delimiter (12/12/). And still does not allow to enter invalid symbols and type outside of date format. 
2. Generate user-readable placeholder and helper text thanks to https://github.com/dmtrKovalenko/date-io/pull/340. This should significantly increase UX for keyboard input

Screen Recording 2020-04-15 at 17 38 50


lib/src/_shared/KeyboardDateInput.tsx
```diff
@@ -83,7 +80,7 @@ export const KeyboardDateInput: React.FC<DateInputProps & DateInputRefs> = ({
     if ((rawValue === null || utils.isValid(rawValue)) && !isFocused) {
       setInnerInputValue(getInputValue());
     }
-  }, [rawValue]); // eslint-disable-line
+  }, [rawValue, utils, inputFormat]); // eslint-disable-line
```
oliviertassinari:
```suggestion
  }, [rawValue, utils, inputFormat]);
```

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -44,7 +44,7 @@ export function usePickerState<TInput, TDateValue>(
     if (!valueManager.areValuesEqual(pickerDate, date)) {
       setPickerDate(date);
     }
-  }, [value]); // eslint-disable-line
+  }, [value, utils]); // eslint-disable-line
```
oliviertassinari, line 52:
```suggestion
  }, [value, utils]);
```

oliviertassinari (commented):
What's the implication with `rifm`, does it mean we no longer need it?

dmtrKovalenko:
The logic of masked input is still working through the rifm, it is actually a tool for this (saving cursor position during input formats).

lib/src/_shared/KeyboardDateInput.tsx
dmtrKovalenko: There is a side-effect vakye `isFocused` which should not be used as a dependency. Do you think we need to move it to refs?

lib/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -81,15 +86,23 @@ export const DateRangePickerInput: React.FC<DateRangeInputProps> = ({
     }
   }, [currentlySelectingRangeEnd, open]);
 
+  // TODO: rethink this approach. We do not need to wait for calendar to be updated to rerender input (looks like freezing)
+  // TODO: so simply break 1 react's commit phase in 2 (first for input and second for calendars) by executing onChange in the next tick
+  const lazyHandleChangeCallback = React.useCallback(
```
dmtrKovalenko, line 91: Described more in this [twitter thread](https://twitter.com/dmtrKovalenko/status/1250777792930873349)

oliviertassinari: Is the underlying issue that the calendar (containing ~30 days instances) too slow to update?

lib/src/_shared/KeyboardDateInput.tsx
oliviertassinari: I don't follow. If we use `isFocused` in the logic, why not include it in the dependency? If it doesn't have a negative side effect, and if it allows us to removes `eslint-disable-line`. Then, it could help us catch issues after a future refactoring?

oliviertassinari (approved):

lib/src/DateRangePicker/DateRangePickerInput.tsx
dmtrKovalenko: Yeah it is slow, cause actually 70+ days for standard desktop calendar. 

Moreover it is working like throttle but on the react level. User doesn’t expect to see January of 0002 year when input is 01/01/2

lib/src/_shared/KeyboardDateInput.tsx
dmtrKovalenko: We don't need to run effect if the input is changed state from focus to blur. We only need to handle it differently by the fact of focused state. 

So this focused state does not need to be state actually. It can be a ref
```jsx
const isFocusedRef = React.useRef(false) 

onFocus={() => isFocusedRef.current = true} 
onBlur={() => isFocusedRef.current = false} 
```

And I am actually thinking about making it a ref...

oliviertassinari: +1 for using a ref if we don't need to rerender. 

I can see `emptyInputText` as a dependency that is also ignored. Is this a dead prop?

Issue #1657, opened 2020-04-14
Import /unstable_TrapFocus instead of /Modal/TrapFocus


oliviertassinari:
Material-UI forbid imports over two levels as it can lead to duplication in people bundles. Meaning, if somebody does:

```jsx
import { Modal } from '@material-ui/core'; // 1.
import TrapFocus from '@material-ui/core'; // 2.
```

It will bundle:

1. https://unpkg.com/browse/@material-ui/core@4.9.10/esm/Modal/TrapFocus.js
2. https://unpkg.com/browse/@material-ui/core@4.9.10/Modal/TrapFocus.js

Now, given we do:

https://github.com/mui-org/material-ui-pickers/blob/7bed283699ef768936a3ec7c5dc89571492dddd4/lib/src/wrappers/DesktopPopperWrapper.tsx#L6

in the codebase, we very likely cause this double bundling quite often.

dmtrKovalenko:
Hmm, I am not really sure about why? In theory, webpack should understand that we are importing the same file. 
> P.S. I`d likely not use `@material-ui/core/Button` imports because of typescript autoimport. I am not writing imports manually at all. 

By the way, is there any eslint plugin that will manually change the import. If no – I can write one

TrySound:
@dmtrKovalenko `@material-ui/core/Modal/TrapFocus.js` and `@material-ui/core/esm/Modal/TrapFocus.js` is not the same file

dmtrKovalenko:
ohh god, really... thanks @TrySound than its definitely a problem 
Moreover, I think we need to provide more clear esm build.

oliviertassinari:
It can be verified with this sandbox https://codesandbox.io/s/wonderful-yonath-5cted?file=/src/App.js I have deployed for production in https://csb-5cted.netlify.com/. 

Duplication 🔽
<img width="206" alt="Capture d’écran 2020-04-14 à 15 42 32" src="https://user-images.githubusercontent.com/3165635/79231591-9af91300-7e66-11ea-8839-8505eb04b417.png">

TrySound:
unstable_TrapFocus export is a good way to go for internal usage.

dmtrKovalenko:
I mean overall. Does import from the core for the current build is somehow worse than path import? 
I`d like to open another issue to clarify imports

oliviertassinari:
> By the way, is there any eslint plugin that will automatically change the import. If no – I can write one

[...]

PR #1671, opened 2020-04-17
renderInput API 

dmtrKovalenko:
## Description

Our new API to render the text field, that solves a lot of problems! Like spreading props down to text field, override them, override classes for the picker – and saves a lot of bundlesize.

## Breaking changes

From now it is required to pass the `renderInput` prop to any picker. For the compatibility (no user-facing changes) it will be 

```diff
<DatePicker
+  renderInput={props => <TextField {...props} />}
/>
```


docs/pages/demo/daterangepicker/CalendarsDateRangePicker.example.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import Grid from '@material-ui/core/Grid';
-import Typography from '@material-ui/core/Typography';
+import { Typography, TextField } from '@material-ui/core';
```
oliviertassinari, line 3: Did we settle on this approach?

```suggestion
import TextField from '@material-ui/core/TextField';
import Typography from '@material-ui/core/Typography';
```

oliviertassinari (commented):
Everything seems to go in the right direction: more freedom with the API, easier to understand, smaller bundle size 💯

```diff
  "build/dist/material-ui-pickers.esm.js": {
+   "gzipped": 26870,
-   "gzipped": 25950,
```

docs/pages/demo/daterangepicker/CalendarsDateRangePicker.example.tsx
dmtrKovalenko: Let's make all these changes separately? Like a part of #1657 
We will never fix it because of typescript autoimport – need automate solution

oliviertassinari: Sure 💯

oliviertassinari: My concern was with https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html, but addressing it, later on, is fine too.

oliviertassinari:
Regarding the props that do no longer spread, did you use our helper to warn?

```jsx
import { exactProp } from '@material-ui/utils';
```

dmtrKovalenko:
@oliviertassinari do you mean to pass all `TextField` prop-types there?

dmtrKovalenko:
I am not sure that this module will work for us properly... I mean that there will be more problems than positive effects, as we have a lot of props that are spread for different components (and for this design we passing all the props everywhere XD)

oliviertassinari:
@dmtrKovalenko This plugin works as long as 1. the props aren't spread (which we are introducing) 2. all the prop-types are defined (which is a standard for our components). So I don't see your concern, do you have more details? For instance, in this case:

```jsx
<DatePicker foo="bar" />
```

should it warn? saying that the `foo` prop isn't supported?

dmtrKovalenko:
We don't have all the prop-types defined in 1 place. We defining prop-types in the components that are using that prop-types.

oliviertassinari:
@dmtrKovalenko Ok, so we can't yet do it. I assume this problem will be solved in the mono-repository. We use the Types to automatically generate the prop-types. We would have them all defined, as in https://next.material-ui-pickers.dev/api/DatePicker.

dmtrKovalenko:
yes, it should be a nice addition for *Pickers

oliviertassinari (approved):

PR #1672, opened 2020-04-17
[DatePicker] Fix inputFormat prop update logic

### LINKED ISSUE #1568: DesktopDatePicker change format issue - not updating date

## Environment

| @material-ui/pickers | 4.0.0-alpha.3 |
| material-ui          |  4.6.1 |
| Peer library         |   date-fns      |

## Steps to reproduce
1.Select date
2. Select date in format "dd/MM/yyyy" 
3. Change format in App.jsx to "MMM/dd/yyyy" (Added comments)
4. (Date does not change)

## Expected behavior
When I change the format dinamically, datepicker should rerender the selected date and display in the expected format.

## Actual behavior

For example, if format is "dd/MM/yyyy" and I switch to "MMM/dd/yyyy", value (date) does not update its format automatically, when it should.

## Live example

https://codesandbox.io/s/vigorous-khorana-zvfo8?file=/src/App.js

vdyachenko:
This PR closes #1568 

## Description

Add a missing inputFormat prop to update the value when format is changed

### DIFF
--- a/lib/src/_shared/KeyboardDateInput.tsx
+++ b/lib/src/_shared/KeyboardDateInput.tsx
@@ -83,7 +83,7 @@ export const KeyboardDateInput: React.FC<DateInputProps & DateInputRefs> = ({
     if ((rawValue === null || utils.isValid(rawValue)) && !isFocused) {
       setInnerInputValue(getInputValue());
     }
-  }, [rawValue]); // eslint-disable-line
+  }, [inputFormat, rawValue]); // eslint-disable-line
 
   const handleChange = (text: string) => {
     const finalString = text === '' || text === mask ? null : text;


dmtrKovalenko:
Thank you for your work, I am really sorry but this one is already fixed by #1661

vdyachenko:
> Thank you for your work, I am really sorry but this one is already fixed by #1661

No problem, glad to help :) Mb you have some other bug/feature to contribute?

oliviertassinari:
@dmtrKovalenko I think that we should favor solving individual issues when we can, so we can deliver an independent chunk of values.

Unless #1661 makes the fix proposed here outdated, I think that we could add a test case and accept it :).

PR #1689, opened 2020-04-19
Align DateRangePicker inputs by baseline

Philipp91:
In particular, this affects the positioning of the "to" in between the two boxes. It doesn't make a difference when the pickers use outlined styling (like in the demo), but it does fix the positioning with the standard styling.

### DIFF
--- a/lib/src/DateRangePicker/DateRangePickerInput.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerInput.tsx
@@ -14,7 +14,7 @@ export const useStyles = makeStyles(
   theme => ({
     rangeInputsContainer: {
       display: 'flex',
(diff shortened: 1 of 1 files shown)


dmtrKovalenko (approved):

dmtrKovalenko:
@Philipp91  Could you please give more context? I don't like the new appearance. XD 

image

Philipp91:
Whoops. Maybe I adjusted the wrong field? Here is what I was trying to fix:
image

This PR was untested, also it doesn't reproduce in the demo because the demo uses outlined text fields.

dmtrKovalenko:
The problem is for mobile displaying. Check from any mobile device.

PR #1692, opened 2020-04-20
[core] Align internal icons with mono-repository

oliviertassinari:
- Remove redundant `Icon` suffix
- Use import *
- Label as internal, a deterrent for developers to import these icons
- Remove useless paths


dmtrKovalenko (changes requested):
I'd like not to move to default exports of `React` at least for now. 
And also interesting why visual regression didn't start? Probably because this is a forked branch. Need to investigate – I'd like to see what's changed after deleting paths from SVG.

oliviertassinari:
What's the concern with the import, it seems we do it most of the time already? 
https://github.com/mui-org/material-ui-pickers/blob/0d41c48aaa8b9b6f67e72deee9583340cd218b89/lib/src/views/Clock/ClockNumber.tsx#L1

dmtrKovalenko:
image

But it is there ...
image

dmtrKovalenko:
@oliviertassinari  Sorry I am blind. For some reason I thought you moved `*as` to default. It is ok – thanks... my bad XD 

> Its left to figure out what's wrong with visual regression. It's for me probably

dmtrKovalenko (approved):

Issue #1693, opened 2020-04-20
Customizing KeyboardIcon and PenIcon for keyboard view toggle


Philipp91:
The icons are currently [hard-coded](https://github.com/mui-org/material-ui-pickers/blob/c413b02149ecad16c928fd043967509b1f5f42f3/lib/src/_shared/PickerToolbar.tsx#L104-L106).

There is already a `keyboardIcon` prop, but that one controls the icon in the input adornment. I don't really understand why that prop has "keyboard" in its name. It does not show a keyboard on the icon. And one would click that button precisely when _not_ using the keyboard but rather the mouse to input a date. I'd call that one `openIcon` (by function) or `calendarIcon` (by icon content).

dmtrKovalenko:
Yeah, it is reaaly legacy name. Should be nice to change in scope of `v4-alpha` breaking changes

dmtrKovalenko:
Would you like to open PR?

Philipp91:
Well it's not just about renaming the existing prop, but also about exposing new props for those two hard-coded icons.

dmtrKovalenko:
Yeah so I propose these changes:

1. Rename `keyboardIcon` =>  `calendarIcon` 
2. Use this icon also for rendering mobile keyboard input view 
3. Add new prop `editIcon` for the mobile keyboard input view

Philipp91:
Sure, I can take a stab at this. I can't assign myself in GitHub, but stay tuned.

Philipp91:
> Yeah so I propose these changes:
> 
> 1. Rename `keyboardIcon` =>  `calendarIcon`

Naming the prop after the icon content doesn't play well with [this usage](https://github.com/mui-org/material-ui-pickers/blob/341171f5ae4e600f8d8efcfebf3837adbbb293cc/lib/src/TimePicker/TimePicker.tsx#L41), where we'd end up with `calendarIcon: <ClockIcon/>` even those two are sort-of opposites and shouldn't be equated like that.

So perhaps `openIcon` or something else in the direction of the icon's functionality (adornmentIcon, buttonIcon) would be better?

dmtrKovalenko:
Hmm really, `openPickerIcon` should be nice?

Philipp91:
> 2. Use this icon also for rendering mobile keyboard input view
> 3. Add new prop editIcon for the mobile keyboard input view

[...]

PR #1696, opened 2020-04-20
Rename keyboardIcon => openPickerIcon

Philipp91:
See #1693 [step 1](https://github.com/mui-org/material-ui-pickers/issues/1693#issuecomment-616758267).

### DIFF
--- a/lib/src/_shared/icons/CalendarIcon.tsx
+++ b/lib/src/_shared/icons/CalendarIcon.tsx
@@ -6,5 +6,5 @@ export const CalendarIcon = createSvgIcon(
     <path d="M17 12h-5v5h5v-5zM16 1v2H8V1H6v2H5c-1.11 0-1.99.9-1.99 2L3 19c0 1.1.89 2 2 2h14c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2h-1V1h-2zm3 18H5V8h14v11z" />
     <path fill="none" d="M0 0h24v24H0z" />
   </>,
-  'CalendarIcon'
+  'Calendar'
 );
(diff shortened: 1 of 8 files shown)


Philipp91:
Is there an ongoing list of breaking changes where I should be adding this?

dmtrKovalenko (approved):

lib/src/_shared/icons/CalendarIcon.tsx
```diff
@@ -1,10 +1,10 @@
...
 import { createSvgIcon } from '@material-ui/core/utils';
 
-export const KeyboardIcon = createSvgIcon(
+export const CalendarIcon = createSvgIcon(
   <>
     <path d="M17 12h-5v5h5v-5zM16 1v2H8V1H6v2H5c-1.11 0-1.99.9-1.99 2L3 19c0 1.1.89 2 2 2h14c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2h-1V1h-2zm3 18H5V8h14v11z" />
     <path fill="none" d="M0 0h24v24H0z" />
   </>,
-  'KeyboardIcon'
+  'CalendarIcon'
```
oliviertassinari:
```suggestion
  'Calendar'
```

oliviertassinari: https://github.com/mui-org/material-ui/blob/0298ee013da2d8e7e8944cb18471da08e02f87b3/packages/material-ui/src/utils/createSvgIcon.js#L17

Philipp91:
Do you generally prefer additional commits pushed behind the existing ones, or commit amendments?

dmtrKovalenko:
Coverage mismatch is actually caused by missing cypress coverage #1507. So we can ignore it

PR #1706, opened 2020-04-24
Expose mobile{Keyboard,Default}ViewIcon props from PickerToolbar

### LINKED ISSUE #1693: Customizing KeyboardIcon and PenIcon for keyboard view toggle

Philipp91:
This PR closes #1693.

### DIFF
--- a/docs/pages/demo/datetime-picker/CustomDateTimePicker.example.jsx
+++ b/docs/pages/demo/datetime-picker/CustomDateTimePicker.example.jsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import AlarmIcon from '@material-ui/icons/Alarm';
 import SnoozeIcon from '@material-ui/icons/Snooze';
 import ClockIcon from '@material-ui/icons/AccessTime';
+import PenIcon from '@material-ui/icons/Edit';
 import { TextField } from '@material-ui/core';
 import { DateTimePicker, MobileDateTimePicker } from '@material-ui/pickers';
 
@@ -26,6 +27,8 @@ function CustomDateTimePicker(props) {
         leftArrowButtonText="Open previous month"
         rightArrowButtonText="Open next month"
         openPickerIcon={<ClockIcon />}
+        mobileKeyboardViewIcon={<PenIcon color="black" />}
+        mobileDefaultViewIcon={<ClockIcon />}
         minTime={new Date(0, 0, 0, 9)}
         maxTime={new Date(0, 0, 0, 20)}
         renderInput={props => (
--- a/lib/src/Picker/Picker.tsx
+++ b/lib/src/Picker/Picker.tsx
@@ -35,6 +35,8 @@ export type ToolbarComponentProps<
   timeIcon?: React.ReactNode;
   isLandscape: boolean;
   ampmInClock?: boolean;
+  mobileKeyboardViewIcon?: React.ReactNode;
+  mobileDefaultViewIcon?: React.ReactNode;
   isMobileKeyboardViewOpen: boolean;
   toggleMobileKeyboardView: () => void;
   getMobileKeyboardInputViewButtonText?: () => string;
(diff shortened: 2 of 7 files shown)


lib/src/_shared/PickerToolbar.tsx
```diff
@@ -95,16 +99,12 @@ const PickerToolbar: React.SFC<PickerToolbarProps> = ({
         {children}
         <IconButton
```
dmtrKovalenko, line 100: Probably we need to also expose props that passing directly to this component.

`MobileKeyboardButtonProps`

docs/pages/demo/datetime-picker/CustomDateTimePicker.example.jsx
```diff
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import AlarmIcon from '@material-ui/icons/Alarm';
 import SnoozeIcon from '@material-ui/icons/Snooze';
 import ClockIcon from '@material-ui/icons/AccessTime';
+import PenIcon from '@material-ui/icons/Edit';
```
oliviertassinari, line 5: For consistency with the two imports above?

```suggestion
import AccessTimeIcon from '@material-ui/icons/AccessTime';
import EditIcon from '@material-ui/icons/Edit';
```

lib/src/_shared/PickerToolbar.tsx
Philipp91: Okay. At that point, maybe we should remove `mobileKeyboardViewButtonClassName`? To me it seems like it's just yet another prop on that component -- or are `className`s worth treating explicitly in general?

The `onClick={toggleMobileKeyboardView}` is probably worth keeping as a separate prop, as it must be implemented by the user of the component.

Philipp91: Ah well, `mobileKeyboardViewButtonClassName` is used by all the Toolbar implementations, so let's keep it.

Philipp91: So you proposed "**MobileKeyboardButton**Props". I called it "**mobileKeyboardViewButton**ClassName" so far. Now I realised that there's already "get**MobileKeyboardInputViewButton**Text = defaultGet**KeyboardInputSwitchingButton**Text" in the code. Probably we shouldn't have 4 different terms for the same button.

Currently, the [DateTimePicker](https://next.material-ui-pickers.dev/demo/datetime-picker) also shows this button on desktop (because it uses the toolbar in its popover, because it has so many different views). Is that desired and will stay like that (then I'd propose to not have "mobile" in the name of this button), or should the desktop DateTimePicker not have the button (then I can send a separate PR to hide it there).

Philipp91: @dmtrKovalenko @oliviertassinari Any opinions on whether the view switching button should be shown on desktop at all?

And what's the proper name for this button?

lib/src/Picker/Picker.tsx
```diff
@@ -35,6 +35,8 @@ export type ToolbarComponentProps<
   timeIcon?: React.ReactNode;
   isLandscape: boolean;
   ampmInClock?: boolean;
+  mobileKeyboardViewIcon?: React.ReactNode;
+  mobileDefaultViewIcon?: React.ReactNode;
```
dmtrKovalenko, line 39: Probably better name will be 
```suggestion
  mobilePickerViewIcon?: React.ReactNode;
```

dmtrKovalenko: And also if it is a public API it should be documented in jsdoc. I am not sure where we are documenting props for `Toolbar` but please try just adding here a comment with description

dmtrKovalenko (changes requested):

oliviertassinari:
@Philipp91 Thanks for the pull request. Unfortunately, we can't accept it here. On-going work was moved to v5 in the main repository: https://next.material-ui.com/components/date-picker/.

I have transferred the linked issue to https://github.com/mui-org/material-ui/issues/23604.

Issue #1604, opened 2020-03-26
Implement forwardRef


oliviertassinari:
While https://github.com/mui-org/material-ui-pickers/pull/1593 is a great step forward, I believe there is still work needed to be done in order to match the expectations developers have built using the core components:

### 1. Public components missing forwardRef:

- Calendar
https://github.com/mui-org/material-ui-pickers/blob/5560ade36a42a038bfef0cae25055ebc50577982/lib/src/views/Calendar/Calendar.tsx#L96
- ClockView
https://github.com/mui-org/material-ui-pickers/blob/5560ade36a42a038bfef0cae25055ebc50577982/lib/src/views/Clock/ClockView.tsx#L84

### 2. Ref not forwarded on the root element:

The ref is forwarded to the underlying `<input>` element, which is not the root element: 

- StaticTimePicker
- MobileTimePicker
- DesktopTimePicker
- TimePicker
- DatePicker
- MobileDatePicker
- StaticDatePicker
- DekstopDatePicker
- DateTimePicker
- MobileDateTimePicker
- StaticDateTimePicker
- DekstopDateTimePicker 

## Environment

| Tech                 | Version |
| -------------------- | ------- |
| @material-ui/pickers | v4.0.0-alpha.4  |

## Steps to reproduce

https://codesandbox.io/s/gracious-haibt-0pnhc

stunaz:
Hello @oliviertassinari , why isnt this part of V4... I think it was fine with material-ui@v4..... but with material-ui@v5 it breaks all over the place.

oliviertassinari:
@stunaz ?

PR #1707, opened 2020-04-24
Export the default Toolbar components

### LINKED ISSUE #1694: Consider exposing the default toolbar implementations
I would like to extend the DateTimePickerToolbar through composition, so it would be useful to be able to import the [default implementation](https://github.com/mui-org/material-ui-pickers/blob/next/lib/src/DateTimePicker/DateTimePickerToolbar.tsx).

Philipp91:
This PR closes #1694

### DIFF
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -2,11 +2,19 @@ import { DayProps as DayPropsType } from './views/Calendar/Day';
 
 export * from './DatePicker';
 
+export { DatePickerToolbar } from './DatePicker/DatePickerToolbar';
+
 export * from './TimePicker';
+
+export { TimePickerToolbar } from './TimePicker/TimePickerToolbar';
 
 export * from './DateTimePicker';
 
+export { DateTimePickerToolbar } from './DateTimePicker/DateTimePickerToolbar';
+
 export * from './DateRangePicker/DateRangePicker';
+
+export { DateRangePickerToolbar } from './DateRangePicker/DateRangePickerToolbar';
 
 export { Calendar } from './views/Calendar/Calendar';
 


dmtrKovalenko (approved):

oliviertassinari (commented):
The change seems to make 4 new components to include in item 1 of #1604, is this accurate? Should we add the forwardRef now, while we are at it?

oliviertassinari:
This one is also interesting, the name of the stylesheet doesn't match the name of the component:

https://github.com/mui-org/material-ui-pickers/blob/7bed283699ef768936a3ec7c5dc89571492dddd4/lib/src/DateTimePicker/DateTimePickerToolbar.tsx#L37

I think that we should update it.

Actually, given that the style sheet is public (`Mui` prefix, not `Private`) making the component public too makes sense.

Philipp91:
> The change seems to make 4 new components to include in item 1 of #1604, is this accurate? Should we add the forwardRef now, while we are at it?

I'd do that in a separate commit/PR anyway, as it's a separate concern. Personally I have 0 knowedge about refs, I don't use them in my application.

> the name of the stylesheet doesn't match the name of the component. I think that we should update it.

Good point. Actually, they're currently called: `MuiPickersDatePickerRoot`, `MuiPickersTimePickerToolbar`, `MuiPickerDTToolbar` and `MuiPickersDatePickerRoot` (for the range picker). The latter obviously lacks a "Range" in the middle. But also, is the correct prefix "MuiPicker" or "MuiPickers"? And is the correct suffix "Root" or "Toolbar"?

oliviertassinari:
> I'd do that in a separate commit/PR anyway, as it's a separate concern. Personally I have 0 knowedge about refs, I don't use them in my application.

@Philipp91 I think that we would want to handle it here, if we want to make the components public, we might as well have it following the standard.

Regarding the name of the style sheets, they should be identical to the name of the exposed components. Does it answer the problem?

dmtrKovalenko (changes requested):
I think that we should 
1. handle the public ref
2. handle the public class names

Philipp91:
> I think that we should
> 
> 1. handle the public ref

I tried, but like I said I have too little experience with refs (and I guess material-ui's TypeScript types in this case) to figure it out. Here's my current attempt: https://github.com/Philipp91/material-ui-pickers/commit/29f16c2029063d79b9accfdd51f38b9acd42f5d8

It throws a long error message on the `<Toolbar` line, which I believe boils down to
```
Type 'HTMLDivElement' is not assignable to type 'CommonProps<ToolbarTypeMap<{}, "div">>'. 
  The types of 'style.backfaceVisibility' are incompatible between these types. 
  Type 'string' is not assignable to type '"-moz-initial" | "inherit" | "initial" | "revert" | "unset" | "hidden" | "visible" | undefined'.
```

> 2. handle the public class names

By that, did you mean the styles overrides? (See https://github.com/mui-org/material-ui-pickers/pull/1736) If not, what else do you mean by class names?

oliviertassinari:
> By that, did you mean the styles overrides? (See #1736) If not, what else do you mean by class names?

Great, thanks for initiating this effort

dmtrKovalenko (approved):

Issue #1674, opened 2020-04-17
Improve runtime


oliviertassinari:
Opening the date picker on desktop and mobile have a perceptible delay, it's even more noticeable with the range picker.

For instance, using the desktop version as a baseline: https://next.material-ui-pickers.dev/demo/daterangepicker#responsiveness. On my high-end $3,000 MacBook, it takes about 190ms to get visual feedback:

<img width="592" alt="Capture d’écran 2020-04-18 à 00 19 38" src="https://user-images.githubusercontent.com/3165635/79618508-850b7c80-810a-11ea-83c1-578089d92832.png">

To be compared with:

- the 180ms of [Kendo-UI](https://www.telerik.com/kendo-react-ui/components/dateinputs/daterangepicker/) 
- the 140ms of [react-dates](https://airbnb.io/react-dates/?path=/story/daterangepicker-drp--default)
- the 80ms of [react-datepicker](https://reactdatepicker.com/)
- the 40ms of [react-day-picker](http://react-day-picker.js.org/)

I think that anything below 80ms would be great 🔥(react-datepicker, relative to my machine). For comparison, in the [Autocomplete](https://material-ui.com/components/autocomplete/#combo-box), it takes 70ms to open the popup with 100 items. 40ms is mind-blowing 🤯 (react-day-picker, relative to my machine).

---

Regarding delivering on better runtime performance

1. The first thing I would look at is the usage of the `ButtonBase` for the days, I suspect an easy win by replacing them with a simple `<button>`, it seems to be an obvious win, we don't need the ripple there.

For instance, in dev mode, we get from 400ms to 250ms with this diff:

```diff
diff --git a/lib/src/DateRangePicker/DateRangePickerDay.tsx b/lib/src/DateRangePicker/DateRangePickerDay.tsx
index 9f54cb31..05ea0182 100644
--- a/lib/src/DateRangePicker/DateRangePickerDay.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerDay.tsx
@@ -135,13 +135,7 @@ export const PureDateRangeDay = ({
           [classes.rangeIntervalDayPreviewStart]: isStartOfPreviewing || isStartOfMonth,
         })}
       >
-        <Day
-          allowKeyboardControl={false}
-          disableMargin
-          {...other}
-          day={day}
-          selected={selected}
-          inCurrentMonth={inCurrentMonth}
+        <button
           data-mui-test="DateRangeDay"
           className={clsx(
[...]

PR #1730, opened 2020-04-30
Improve onError validation 

### LINKED ISSUE #1675: Remove the built-in validation logic
As raised by @dmtrKovalenko we will be better off leaving the resolution of this problem to the form libraries. We can simplify the component and save bundle size by moving in this direction. The only aspect we need to be cautious about is maintaining a way for the developers to know that the component has failed to format the input of the user. 

We haven't looked too much into the resolution of this issue. We would need to benchmark first. Some options that came to mind:

1. Add a new onError callback prop. But does it play well with the form libraries
2. Add `onInputChange` callback prop to be able to double monitor the value of the textbox and the picker, exactly like the Autocomplete do. http://react-day-picker.js.org/examples/input-state
3. Implement something similar to the native date picker input. If the value is invalid, it has: `event.target.value === ''` and `event.target.validity.valid = false`.

No matter what we pick, I think that we should decide based on formik, final-form, react-simple-form perspective.

dmtrKovalenko:
This PR closes #1675 

## Description
Reworked our internal validation to be more useful and easier in form integration. 

## Breaking changes 

* Removed props: 
1. `minDateMessage`
2. `maxDateMessage`
3. `invalidDateMessage` 
4. `strictCompareDates` 
5. `emptyText` 

* Change the signature of `onError` callback to return a reason why the error must be displayed. We are not displaying english error messages by default, If the input is invalid by default picker will only show red colored `TextField`. So from now form validation with sharing internal pickers logic will looks like this (`reason` is well-typed string union): 

```jsx
<DatePicker
  onError={(reason, value) => {
    switch (reason) {
      case 'invalidDate':
        setFieldError(name, 'Invalid date format');
        break;

      case 'disablePast':
        setFieldError(name, 'Values in the past are not allowed');
        break;

      case 'maxDate':
        setFieldError(name, `Date should not be after ${format(maxDate, 'P')}`);
        break;

      case 'minDate':
        setFieldError(name, `Date should not be after ${format(maxDate, 'P')}`);
        break;

      case 'shouldDisableDate':
        // shouldDisableDate returned true, render custom message according to the `shouldDisableDate` logic
        setFieldError(name, getCustomShouldDisableDateErrorMessage(value));
        break;

      default:
        clearError(name);
    }
  }}
/>;

```


docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
...
+              [name]: undefined,
+            });
+        }
+      }}
+      // Make sure that your 3d param is set to `true` on order to not clear errors
+      onChange={date => form.setFieldValue(name, date, false)}
+      renderInput={props => (
+        <TextField
+          name={name}
+          {...props}
```
oliviertassinari:
```suggestion
          {...props}
          name={name}
```

docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
...
+          // Make sure that your 3d param is set to `true` on order to not clear errors
+          onBlur={() => form.setFieldTouched(name, true, false)}
+        />
+      )}
+      {...other}
+    />
+  );
+};
+
+const FormikExample = () => {
```
oliviertassinari:
```suggestion
export default FormikOurValidation() {
```

docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <Formik onSubmit={console.log} initialValues={{ date: new Date() }}>
```
oliviertassinari, line 91: A debug utils forgotten?

```suggestion
    <Formik initialValues={{ date: new Date() }}>
```

lib/src/__tests__/Validation.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+    if (process.env.UTILS === 'luxon') {
+      return;
+    }
+
+    const onErrorMock = jest.fn();
+    const component = mountPickerWithState(utilsToUse.date(), stateProps => (
+      <DesktopDatePicker {...stateProps} disablePast onError={onErrorMock} />
+    ));
+
+    component.find('input').simulate('change', {
```
oliviertassinari, line 55: Is this a new test? What about we force ourselves not to write a single new test with enzyme? Preferring testing-library as much as possible, cypress when testing-library isn't enough (10% of the cases?).

docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import Code from '../../_shared/Code';
```
oliviertassinari: Using `Code` breaks the isolation of the demo, please use something else.

```suggestion
```

I think that not displaying the error at all would provide the best tradeoff, maybe relying on the TextField feedback only?

docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { Grid } from '@material-ui/core';
```
oliviertassinari:
```suggestion
import Grid from '@material-ui/core/Grid';
```

docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { TextField } from '@material-ui/core';
```
oliviertassinari:
```suggestion
import TextField from '@material-ui/core/TextField';
```

docs/pages/guides/FormikOurValidation.example.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const DatePickerField: React.FC<DatePickerFieldProps> = ({
+  form,
+  field: { value, name },
+  maxDate = new Date('2100-01-01'),
+  minDate = new Date('1900-01-01'),
+  getShouldDisableDateError,
+  ...other
+}) => {
```
oliviertassinari: For consistency

```suggestion
function DatePickerField(props: DatePickerFieldProps) {
  const {
    form,
    field: { value, name },
    maxDate = new Date('2100-01-01'),
    minDate = new Date('1900-01-01'),
    getShouldDisableDateError,
    ...other
  } = props;
```

lib/src/DateRangePicker/DateRangePicker.tsx
```diff
@@ -56,79 +56,70 @@ export function makeRangePicker<TWrapper extends SomeWrapper>(Wrapper: TWrapper)
...
-    disableHighlightToday,
-    reduceAnimations,
     value,
     onChange,
     mask = '__/__/____',
     startText = 'Start',
     endText = 'End',
     inputFormat: passedInputFormat,
-    ...restPropsForTextField
+    minDate: __minDate = defaultMinDate,
```
oliviertassinari, line 68: For consistency

```suggestion
    minDate: minDateProp = defaultMinDate,
```

lib/src/__tests__/Validation.test.tsx
dmtrKovalenko: I am not using testing lib in pickers yet

docs/pages/guides/FormikOurValidation.example.tsx
dmtrKovalenko: It’s about showing form values on submit for user. Moreover this prop is required by ts

docs/pages/guides/FormikOurValidation.example.tsx
dmtrKovalenko: What about form state understanding? Now it is really easy to understand how this example works.

lib/src/__tests__/Validation.test.tsx
oliviertassinari: What about introducing it then? So we stop investing time into a direction we know we want to go back from?

dmtrKovalenko: Probably you are right. I thought that will be able to move 💯 in cypress, but it will not be possible.

oliviertassinari: I had a look at the current state of the tests on the date picker components. It seems that, we have:

- 63 with enzyme mount/shallow (shallow much less frequent) (count with `*.test.tsx` + ` it('`)
- 27 with Cypress (count with `*.spec.ts` + ` it('`)
- 16 with date management (count with `*.test.ts` + ` it('`)

*What's the difference between the usage of "test(" and "it("? It seems to be a simple alias https://jestjs.io/docs/en/api#testname-fn-timeout, what about we update the codebase to a single approach to ease the static analysis of the codebase? The usage of "it" is more frequent, so in favor of using "it" everywhere.* 

For comparison, the Autocomplete component has 82 tests, and the core components have 1956 tests. So it seems that migrating the date picker components in any direction will be a much smaller effort (1/10) than the one we have currently in progress from enzyme to testing-library for the core (there are 600 tests+ still to be migrated).

oliviertassinari: @eps1lon Assuming the sources of the date picker will be hosted on the mono-repository. Do you have any preference on the strategy we use for testing it?

dmtrKovalenko: I am not sure how you calculate test count. But we have 77 cypress tests (count by cypress dashboard comment)

docs/pages/guides/FormikOurValidation.example.tsx
dmtrKovalenko: I prefer export default to be always the last export. It is more readable.

oliviertassinari: We systematically use the other way around to 1. increase the code density of the demos, 2. reduce the duplicate of the name of the demo. I think that it's important we still to a single convention.

lib/src/__tests__/Validation.test.tsx
oliviertassinari: > I am not sure how you calculate test count. But we have 77 cypress tests (count by cypress dashboard comment)

@dmtrKovalenko Interesting, it's the first time I see the Cypress dashboard 👌. I have ignored (1.) `App.spec.ts` and (2.) `VisualRegression.spec.ts` in the previous count. I believe we could remove 1. per the structure of the docs in the mono-repository and need to replace 2. with Argos-CI fixtures screenshot tests. After these two subtractions, we get the correct amount :) 27-28.

Actually, it's interesting to look at the Cypress dashboard. Any idea why it takes 26s to run 28 tests? If we were to write all our tests with Cypress, would it mean 1956 / 28 * 26 = 1816s (30min) of test duration **per** browser target?
In the mono-repository test stack, we run them in about 15-60s per browser (we run them in 4 different browsers). I suspect that fixing #1674 can speed up the tests by a factor x3-5. But will it be enough? It seems that we are still talking about **one order of magnitude** of slowdown if we make Cypress a systematic choice.

This raises a second question, what's our time budget (the time it takes to get feedback and $/month)? Would we be below or above?

docs/pages/guides/FormikOurValidation.example.tsx
oliviertassinari: I think that it depends on the objective. Should we 1. expect the developer to use this demo or 2. is it only for illustration?

oliviertassinari: if it's 2. we will prevent developers to see the source, for instance like in https://material-ui.com/components/popover/#anchor-playground. If it's 1. we absolutely need to make it work in isolation.

lib/src/__tests__/Validation.test.tsx
eps1lon: We don't want to move all component test to cypress. That is for 99% of our tests overkill (we're talking about 1s per test instead of 50ms). Especially considering they polyfill a lot of browser behavior (we want to control). 

There are probably some tests that might be more suited for e2e tests but I'd rather use a lower level library for that (like puppeteer or playwright). Cypress really shines once you have complete user stories e.g. login or data fetching.

oliviertassinari: In this case, I would be in favor of migrating **all** the tests to testing-library, until we hit a roadblock, something that is compelling enough to introduce a new tool in the stack, that the engineers working on Material-UI need to learn and maintain.

We had one case like this in the past, to test the UMD release. We had multiple regressions, we couldn't use Karma, we went with Puppeteer (the test is still present).

As far as I can envision the limits of testing-library + karma, we have:

1. No way to test drag and drop, it's going to be important for the DataGrid and TreeView. For the ClockView of the date picker, it's might also be a concern but I think a less important one. There are no underlying elements that could catch the events and mess around. 
1. No way to test the tab behavior. I have tested this diff, using https://github.com/WordPress/gutenberg/pull/5618 as inspiration, it seems to work with Puppeteer.

```diff
diff --git a/packages/material-ui/test/umd/run.js b/packages/material-ui/test/umd/run.js
index 367466663..8c34ecb2b 100644
--- a/packages/material-ui/test/umd/run.js
+++ b/packages/material-ui/test/umd/run.js
@@ -71,6 +71,8 @@ function App() {

   return (
     <React.Fragment>
+      <input type="text" id="foo" />
+      <input type="text" id="bar" />
       <Button onClick={() => setOpen(true)}>Super Secret Password</Button>
       <Dialog open={open}>
         1-2-3-4-5
@@ -141,8 +143,12 @@ async function run() {

     const page = await startBrowser();
     await page.goto(`http://${host}:${port}`);
+    await page.click( '#foo' );
+    await page.keyboard.press( 'Tab' );
+    const activeElement = await page.evaluate(() => {
+      return document.activeElement.id;
+    });
+    console.log('activeElement', activeElement); // output bar
    await expect(page).toClick('button', { text: 'Super Secret Password' });
    await expect(page).toMatch('1-2-3-4-5');
   } catch (err) {
     log.fatal({
       name: 'test',
```
3. No reliable way to test the performance of the components (do we?). It's going to be important for the DataGrid. We might also need to track this metric too for #1674.
4. No built-in parallelism. We can probably still write x2 more tests without it harming our DX. After that, we might have an issue with the time it takes for the tests to run. We could split the tests into two chunks, one half going to BrowserStack, the other to SauceLabs.
5. else? did I forget something? The debugging experience isn't great but it doesn't seem to have been something we have struggled a lot with (so far).

eps1lon: > No reliable way to test the performance of the components (do we?)

1. I experimented a bit with React.Profiler in our test suite which means that we wouldn't need to write tests and benchmark but reuse the testcases for benchmarking. That's a viable option if I remember correctly. Just needs some infra so that we get a good sample size and don't run it on any PR.

> No way to test drag and drop, it's going to be important for the DataGrid and TreeView.

2. That is something I wouldn't let automated testing touch anyway. Fixtures and manual testing would be way more reliable and also ensures that Drag'n'Drop is actually usable. I think this should be tested manually on different devices.

> We can probably still write x2 more tests without it harming our DX.

3. Please consider that your workflow is the issue. You absolutely do not need to run the full test suite in watchmode when working on the codebase for virtually every use case. `--grep` and `.only` give you all the tools you need.

> No way to test the tab behavior.

4. This is something I never understood. The only meaningful behavior you're missing out is :focus-visible and that is only visual anyway. For all the rest `.focus()` should be enough, no? If you're interested in tab-order then you need to test this manually with an actual browsers/assistive technology.

oliviertassinari: @eps1lon 

1. Great. I have no knowledge of how this should be done correctly. My understanding of the problem is limited to: reproducibility is paramount. Without a stable, noise-free measure, we can't do anything.
2. Agree, manual testing should probably be the main tool used when iterating on the logic. Maybe one automated test, as a rough safe belt could be valuable.
3. My concern is with the CI, the **delay** of the feedback loop on pull requests.
4. Considering the TrapFocus component, Do we have enough tools to test it correctly?

lib/src/DateRangePicker/DateRangePicker.tsx
dmtrKovalenko: Doesn't it look less clear? Because accidentally type `minDateProp` when you are typing `minDate` is much much easier than `__minDate`. And `__` is some kind of convention about naming private variables?

docs/pages/guides/FormikOurValidation.example.tsx
dmtrKovalenko: What do you mean under consistency? `function` vs `const`?

docs/pages/guides/FormikOurValidation.example.tsx
dmtrKovalenko: Yeah, but for now pickers are standalone project. And if we are not planning to migrate docs before v5 we could stick to the our internal conventions for now. (before we merge docs)

oliviertassinari: I think that we can break down the dicussion in 3 different topics:

1. **Standaloneness**. The picker components are part of Material-UI, like any other official component. I think that we should treat the component here as they were inside the mono repository. Where the code is a hosted shouldn't influence the developer experience. I think that it should be to the own discretion of how the core team wants to work.
Why? The consistency of the components is one of the important value propositions of the library, I think that steps in this direction improve the DX.
2. **Progressiveness**. I think that we should leverage all the opportunities to reduce the gap in the "standard way". I think that any step in the long term direction is good to take. What the advantage of delaying it? I find it interesting to handle it as we go, as it will be one less change to handle in the future.
3. **Convention evolution**. I think that we should evolve the convention as we learn from our users and the environment change. RFC sounds like the best framework to make changes.
For history, we used to follow the approach of the pull request, it was optimized for reducing the switching cost between class and functional components. With hooks, this requirement is no longer relevant.

docs/pages/guides/FormikOurValidation.example.tsx
oliviertassinari: I mean that we systematically write private components following this pattern:

```jsx
function DatePickerField(props: DatePickerFieldProps) {
  const {
    form,
    field: { value, name },
    maxDate = new Date('2100-01-01'),
    minDate = new Date('1900-01-01'),
    getShouldDisableDateError,
    ...other
  } = props;
```

and for public ones:

```jsx
const DatePickerField = React.forwardRef(function DatePickerField(
  props: DatePickerFieldProps,
  ref: React.Ref<HTMLElement>,
) {
  const {
    form,
    field: { value, name },
    maxDate = new Date('2100-01-01'),
    minDate = new Date('1900-01-01'),
    getShouldDisableDateError,
    ...other
  } = props;
```

lib/src/DateRangePicker/DateRangePicker.tsx
oliviertassinari: No specific preference, we could make it the new convention, if we do, could you update all the other components (core, lab, etc.)?

dmtrKovalenko: Probably yes :) 
I`ll give you more context: Here we had exactly the same naming `minDateProp` but it was problematic because appeared in the autocomplete for `minDate` so I decided to use another convention

PR #1736, opened 2020-05-03
Clean up overrides.ts

Philipp91:
- Use inline imports to make the file less verbose and easier to change in the future.
- Sort alphabetically instead of arbitrarily.
- Name styles consistently like `MuiPickers<component_name>`, except for `MuiPickersBasePicker`.

### DIFF
--- a/lib/src/DatePicker/DatePickerToolbar.tsx
+++ b/lib/src/DatePicker/DatePickerToolbar.tsx
@@ -18,7 +18,7 @@ export const useStyles = makeStyles(
       top: 4,
     },
   },
-  { name: 'MuiPickersDatePickerToolbar' }
+  { name: 'PrivateDatePickerToolbar' }
 );
 
 export const DatePickerToolbar: React.FC<ToolbarComponentProps> = ({
--- a/lib/src/DatePicker/DatePickerToolbar.tsx
+++ b/lib/src/DatePicker/DatePickerToolbar.tsx
@@ -18,7 +18,7 @@ export const useStyles = makeStyles(
       top: 4,
     },
   },
-  { name: 'MuiPickersDatePickerRoot' }
+  { name: 'MuiPickersDatePickerToolbar' }
 );
 
 export const DatePickerToolbar: React.FC<ToolbarComponentProps> = ({
--- a/lib/src/DateRangePicker/DateRangePickerToolbar.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerToolbar.tsx
@@ -17,7 +17,7 @@ export const useStyles = makeStyles(
       display: 'flex',
     },
   },
-  { name: 'MuiPickersDateRangePickerToolbar' }
+  { name: 'PrivateDateRangePickerToolbar' }
 );
 
 interface DateRangePickerToolbarProps
--- a/lib/src/DateRangePicker/DateRangePickerToolbar.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerToolbar.tsx
@@ -17,7 +17,7 @@ export const useStyles = makeStyles(
       display: 'flex',
     },
   },
-  { name: 'MuiPickersDatePickerRoot' }
+  { name: 'MuiPickersDateRangePickerToolbar' }
 );
 
 interface DateRangePickerToolbarProps
--- a/lib/src/DateTimePicker/DateTimePickerTabs.tsx
+++ b/lib/src/DateTimePicker/DateTimePickerTabs.tsx
@@ -44,7 +44,7 @@ export const useStyles = makeStyles(
       },
     };
   },
-  { name: 'MuiPickerDTTabs' }
+  { name: 'MuiPickersDateTimePickerTabs' }
 );
 
 export const DateTimePickerTabs: React.SFC<DateTimePickerTabsProps> = ({
--- a/lib/src/DateTimePicker/DateTimePickerTabs.tsx
+++ b/lib/src/DateTimePicker/DateTimePickerTabs.tsx
@@ -44,7 +44,7 @@ export const useStyles = makeStyles(
       },
     };
   },
-  { name: 'MuiPickersDateTimePickerTabs' }
+  { name: 'PrivateDateTimePickerTabs' }
 );
 
 export const DateTimePickerTabs: React.SFC<DateTimePickerTabsProps> = ({
--- a/lib/src/DateTimePicker/DateTimePickerToolbar.tsx
+++ b/lib/src/DateTimePicker/DateTimePickerToolbar.tsx
@@ -34,7 +34,7 @@ export const useStyles = makeStyles(
       right: 8,
     },
   }),
-  { name: 'MuiPickerDTToolbar' }
+  { name: 'MuiPickersDateTimePickerToolbar' }
 );
 
 export const DateTimePickerToolbar: React.FC<ToolbarComponentProps> = ({
--- a/lib/src/DateTimePicker/DateTimePickerToolbar.tsx
+++ b/lib/src/DateTimePicker/DateTimePickerToolbar.tsx
@@ -34,7 +34,7 @@ export const useStyles = makeStyles(
       right: 8,
     },
   }),
-  { name: 'MuiPickersDateTimePickerToolbar' }
+  { name: 'PrivateDateTimePickerToolbar' }
 );
 
 export const DateTimePickerToolbar: React.FC<ToolbarComponentProps> = ({
--- a/lib/src/typings/overrides.ts
+++ b/lib/src/typings/overrides.ts
@@ -17,37 +17,10 @@ type Classes<T> = Partial<
 >;
 
 export interface MuiPickersOverrides {
-  MuiArrowSwitcher?: Classes<typeof import('../_shared/ArrowSwitcher').useStyles>;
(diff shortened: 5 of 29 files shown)


lib/src/DatePicker/DatePickerToolbar.tsx
```diff
@@ -18,7 +18,7 @@ export const useStyles = makeStyles(
       top: 4,
     },
   },
-  { name: 'MuiPickersDatePickerRoot' }
+  { name: 'MuiPickersDatePickerToolbar' }
```
dmtrKovalenko: Maybe we need some reusable function that will build a standardized name. Something like
```ts
const buildStyleName = (component: React.ComponentType<any>) => {
   return `MuiPickers${component.displayName || component.name}`
}
```

@oliviertassinari what do your think?

Philipp91: How about making the second argument to `makeStyles()` a component?

oliviertassinari: I think that we should drop the "Pickers" prefix to match the exported name. What would be wrong with?

```suggestion
  { name: 'MuiDatePickerToolbar' }
```

Actually, this would be even better as the component is private:

```suggestion
  { name: 'PrivateDatePickerToolbar' }
```

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
```diff
@@ -17,7 +17,7 @@ export const useStyles = makeStyles(
       display: 'flex',
     },
   },
-  { name: 'MuiPickersDatePickerRoot' }
+  { name: 'MuiPickersDateRangePickerToolbar' }
```
oliviertassinari:
```suggestion
  { name: 'MuiDateRangePickerToolbar' }
```

lib/src/DateTimePicker/DateTimePickerTabs.tsx
```diff
@@ -44,7 +44,7 @@ export const useStyles = makeStyles(
       },
     };
   },
-  { name: 'MuiPickerDTTabs' }
+  { name: 'MuiPickersDateTimePickerTabs' }
```
oliviertassinari:
```suggestion
  { name: 'PrivateDateTimePickerTabs' }
```

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
```diff
@@ -34,7 +34,7 @@ export const useStyles = makeStyles(
       right: 8,
     },
   }),
-  { name: 'MuiPickerDTToolbar' }
+  { name: 'MuiPickersDateTimePickerToolbar' }
```
oliviertassinari:
```suggestion
  { name: 'MuiDateTimePickerToolbar' }
```

lib/src/typings/overrides.ts
```diff
@@ -39,34 +17,29 @@ type Classes<T> = Partial<
...
-  MuiBasePickerStyles?: Classes<typeof MuiBasePickerStyles>;
-  // consider using inline import type notation
-  MuiPickersDesktopDateRangeCalendar?: Classes<
-    typeof import('../DateRangePicker/DateRangePickerViewDesktop').useStyles
-  >;
   MuiPickersArrowSwitcher?: Classes<typeof import('../_shared/ArrowSwitcher').useStyles>;
-  MuiPickersDateRangePickerInput?: Classes<
-    typeof import('../DateRangePicker/DateRangePickerInput').useStyles
-  >;
+  MuiPickersBasePickerStyles?: Classes<typeof import ('../Picker/Picker').useStyles>;
```
oliviertassinari: Most of the exported override keys here seem to be from private components, so to be removed.
Also, in https://github.com/mui-org/material-ui/issues/19427, it seems that we make the DX simpler for TypeScript users.

lib/src/DatePicker/DatePickerToolbar.tsx
Philipp91: What do you mean by "even better"?
#1707 makes this public

Philipp91: Also regarding the `buildStyleName()` function, especially if the prefix is now just `Mui` and nost `MuiPickers`, I think this library should just do the same as the main material-ui library. Does that one have such function?

oliviertassinari: > #1707 makes this public

I think that the change should be atomic. If we make a new component public, it should be done all at once (same pull request}. This would avoid the cases were we forget/revert/etc.

oliviertassinari: > Does that one have such function?

Great point, we don't. I think that a string duplication is good enough.

lib/src/typings/overrides.ts
Philipp91: So the idea there is to stop having this central "registry" in overrides.ts and instead only declaring the export in the respective component's file?
(Something like that would make it easier to keep privateness in sync.)

Philipp91: What makes a component public (or private)? For instance, if I don't see the `ArrowSwitcher` component exported from `src/index.ts`, does that mean that it's private? And also, assuming that was right, should I remove it from `overrides.ts` and call its styles `PrivateArrowSwitcher` instead?

Will we also have a `buildStyleName()` variant or parameter for private components?

oliviertassinari: > What makes a component public (or private)?

Great question. So far on https://github.com/mui-org/material-ui it has been as simple as: Is the module exported from the barrel index? If exported public, if not, private.

Philipp91: Side question: Is it intended that things like `defaultReduceAnimations` are exported from the top-level `index.ts`?

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
Philipp91: If the prefix is really just `Mui` and not `MuiPickers`, that might cause clashes on things like the `ModalDialog`. Currently, that one is not exported, but I'm not sure if even `PrivateModalDialog` could clash with another material-ui-based library that has its own ModalDialog?

Same for `Toolbar`, `ToolbarButton` or `StaticWrapper`.

lib/src/typings/overrides.ts
Philipp91: Do the `overrides.ts` exports and the `index.ts` exports necessarily have to be in sync? Is that a best practice? What are the advantages?

Anyway I pushed a commit that hopefully does that.

oliviertassinari: So far, we have been keeping them in sync. So unless there is a good reason to deviate from this rule, I would be in favor of sticking to it.

Philipp91:
There are a number of questions that haven't been answered:
> Side question: Is it intended that things like defaultReduceAnimations are exported from the top-level index.ts?
> If the prefix is really just Mui and not MuiPickers, that might cause clashes on things like the ModalDialog.
> Maybe we need some reusable function that will build a standardized name. Something like ... `buildStyleName` ...

Though it's quite possible that you think it's fine to leave as is. Otherwise please let me know if you want me to change something in this PR.

oliviertassinari:
> Side question: Is it intended that things like defaultReduceAnimations are exported from the top-level index.ts?

@Philipp91 I think that it was exposed by chance.

> If the prefix is really just Mui and not MuiPickers, that might cause clashes on things like the ModalDialog.

The picker components are part of the official set of Material-UI components. I think that the prefix should be `Mui`.

> Maybe we need some reusable function that will build a standardized name. Something like ... buildStyleName ...

What's wrong with a duplication of the name of the component?

Philipp91:
Okay, then this PR should be fine as is.

dmtrKovalenko:
@oliviertassinari we cannot ship this PR. 
We are not providing any other styling solution different from global overrides. So we **must** make all of our components style-able by global overrides.

I am so sorry for being too late @Philipp91. You are doing awesome contribution, thank you!

dmtrKovalenko:
Even if we will provide `classes` prop. We need to make all the components like `Calendar` accept global overrides, just because it is useful as feature

oliviertassinari:
> So we must make all of our components style-able by global overrides.

@dmtrKovalenko The concern is interesting. The way we have solved that problem with the other components was by either 1. exposing the lower-level building blocks, allowing people to compose them or 2. to keep all the logic under a single public style sheet and hide a bit into private style sheets with the option to provide custom implementations.
However, this component is different. It's the first time I encounter a component with this set of challenges.

Any objection to make all the components that expose their styles public?

I have noticed a couple of styles that are private, is it expected?

Capture d’écran 2020-05-08 à 00 32 22

dmtrKovalenko:
1. I don’t really know about all the components. But it looks useful from the custom design system perspective. If we can design anything in 1 place it is easier to reuse custom theme. 

2. Private styles are unexpected.

oliviertassinari:
How can we move forward? Maybe we should be less ambitious with the changes here. @Philipp91 Was was the original issue you were going after?

Philipp91:
My ultimate goal was exporting the toolbar components (#1707). It doesn't really matter (for my use case) whether their styles are exported too or not.

oliviertassinari:
Ok cool, in this case, maybe we should close and leave this pull request to the core team?

Philipp91:
Sure, I'll let you take it over.

PR #1760, opened 2020-05-09
Improve `renderInput` spreading experience

### LINKED ISSUE #1751: [docs] Document how to use a custom input with renderInput
## Environment

| @material-ui/pickers |   4.0.0-alpha.6     |
| material-ui                  |   4.9.13      |
| React                          |    16.12     |
| Browser                      |    Chrome     |
| Peer library                |  moment-io|

I have this DatePicker
```
<DatePicker
   renderInput={props => <MyText {...props} />}
   value={date}
   onChange={newDate => onChange(newDate)}
/>
```

And this custom TextField:
```
const MyText = props => {
  return (
    <input
      {...props.inputProps}
      onChange={props.onChange}
      onBlur={props.onBlur}
      onClick={props.onClick}
      value={props.value}
      className="input"
      type="text"
    />
  );
};
```
But there are several problems:
* When I click on the input I am expecting to see popup with the days. However can't see it. I see that props.onClick === openPicker. Its correct. What the problem could be? (it works ok with <TextField /> from the core);
* Also I can see warning: `Failed prop type: You provided a "value" prop to a form field without an "onChange" handler.` Not clear why is that onChange does not propagated?

Could you guys maybe share small example how to use renderInput with custom html input?

### LINKED ISSUE #1758: Improve renderInput props structure
Having a closer look at #1751, I think that we have an opportunity to improve the API. Basically, I have been wondering about how to best structure the `props` object for the Autocomplete in the past: https://github.com/mui-org/material-ui/blob/05480c173f6a491760fdbf51552ff8bcc4beb7d2/packages/material-ui-lab/src/Autocomplete/Autocomplete.js#L408-L456.

What about?

- Remove `data-mui-test` from the spread options. This should be development only prop, and even then, we should aim to replace most, if not all of these attributes with a11y features. In our case, it seems that `keyboard-date-input` isn't used. I believe we can remove it without much concerns :).
- Move `onChange`, `onFocus`, `onBlur`, `type`, `value`, `placeholder ` into `inputProps`.
- Remove `label`, leave it to the user
- In Autocomplete, move ref from `InputProps` to the root.

This way, we would get a similar DX and a simpler customization API:

```jsx
<DatePicker
  open={isForcePickerOpen}
  onClose={() => setIsOpen(false)}
  value={selectedDate}
  onChange={handleDateChange}
  renderInput={({ ref, inputProps, InputProps }) => (
    <div ref={ref}>
      <input {...inputProps} />
      {InputProps.endAdornment}
    </div>
  )}
/>
```

dmtrKovalenko:
This PR closes #1751 and closes #1758


docs/pages/demo/datepicker/CustomInput.example.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { useState } from 'react';
+import { styled } from '@material-ui/core';
+import { DesktopDatePicker } from '@material-ui/pickers';
+
+const InputContainer = styled('div')({
+  display: 'flex',
+  alignItems: 'center',
+});
+
+function CustomInput() {
```
oliviertassinari, line 10: ?
`export default function CustomInput()`

docs/pages/demo/datepicker/CustomInput.example.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { useState } from 'react';
```
oliviertassinari, line 1: ?
`import * as React from 'react';`

docs/pages/regression/Regression.tsx
```diff
@@ -100,9 +100,15 @@ function Regression() {
         onChange={changeRange}
         renderInput={(startProps, endProps) => (
           <>
-            <TextField {...startProps} inputProps={{ 'data-mui-test': 'desktop-range-picker' }} />
+            <TextField
+              {...startProps}
+              inputProps={{ ...startProps.inputProps, 'data-mui-test': 'desktop-range-picker' }}
```
oliviertassinari, line 105: Side question, should we use a label to differentiate the two inputs instead of `data-mui-test`? Like `label="Start date"` and `label="End date"`?

docs/pages/demo/datepicker/index.mdx
```diff
@@ -53,11 +53,11 @@ For that use `StaticDatePicker`.
   <Example paddingBottom source={StaticDatePicker} />
 </Hidden>
 
-#### Advanced keyboard input
+#### Custom input component
 
-We are providing default localized formats, but you can change this behaviour with `format` and `mask` props.
+You can customize rendering input by `renderInput` prop. But make sure you will spread `ref` and `inputProps` correctly to the custom input component.
```
oliviertassinari, line 58: Maybe?

```
The `renderInput` prop allows you to customize the rendered input.
The first argument of this render prop contains props that you need to forward.
Pay specific attention to the `ref` and `inputProps` keys.
```

cc @mbrookes

docs/pages/demo/datepicker/CustomInput.example.jsx
```diff
@@ -0,0 +1,28 @@
...
+import { styled } from '@material-ui/core';
```
oliviertassinari, line 2: ?
`import { styled } from '@material-ui/core/styles';`

docs/pages/demo/datepicker/index.mdx
dmtrKovalenko: We must specify that props are https://material-ui.com/api/text-field/#textfield-api. Will refer this comment at #1761

dmtrKovalenko: By the way the type of `renderInput` looks to awful

```tsx
(props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>
```

Because of core TextField typings (https://github.com/mui-org/material-ui/issues/20716)

oliviertassinari: Would a similar approach to Autocomplete work?

```tsx
export interface AutocompleteRenderInputParams {
  id: string;
  disabled: boolean;
  InputLabelProps: object;
  InputProps: {
    ref: React.Ref<any>;
    className: string;
    startAdornment: React.ReactNode;
    endAdornment: React.ReactNode;
  };
  inputProps: object;
}

  /**
   * Render the input.
   *
   * @param {object} params
   * @returns {ReactNode}
   */
  renderInput: (params: AutocompleteRenderInputParams) => React.ReactNode;
```

PR #1761, opened 2020-05-09
Add missing prop types

### LINKED ISSUE #1750: Missing `renderInput` prop-type
## Environment

| @material-ui/pickers |   4.0.0-alpha.6     |
| material-ui                  |   4.9.13      |
| React                          |    16.12     |
| Browser                      |    Chrome     |
| Peer library                |  moment-io|

I have this DatePicker with latest alpha6
```
          <DatePicker
            variant="outlined"
            autoOk
            //renderInput={TextField}
            value={date}
            disabled={isLoading}
            onChange={newDate => onChange(newDate.format(apiDateFormat))}
            maxDate={maxDate}
            disablePast
          />
```

When I comment out the `renderInput` props I get error "renderInput is not a function";
image

Am I doing something wrong? Or `renderInput` required now?

dmtrKovalenko:
This PR closes #1750


lib/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -183,3 +184,5 @@ export const DateRangePickerInput: React.FC<DateRangeInputProps> = ({
     </div>
   );
 };
+
+DateRangePickerInput.propTypes = inputPropTypes;
```
oliviertassinari: Is the `inputPropTypes` import pruned from the bundle?
What do you think of duplicating the information in the propTypes (no imports)? This could be one step closer to how we automatically generate the prop types from the typescript definitions in the main repository.

lib/src/constants/prop-types.ts
```diff
@@ -17,3 +17,13 @@ export const DomainPropTypes = { date, datePickerView };
 export const defaultMinDate = new Date('1900-01-01');
 
 export const defaultMaxDate = new Date('2100-01-01');
+
+export const inputPropTypes = {
+  renderInput: PropTypes.func.isRequired,
```
oliviertassinari: sort asc?

lib/src/DateRangePicker/DateRangePickerInput.tsx
dmtrKovalenko:
```
process.env.NODE_ENV !== "production" ? PureDateInput.propTypes = inputPropTypes : void 0;
```

From build code. Do you think we should duplicate it?

oliviertassinari: I'm asking because, in the core, we are from time to time using prop-types helpers:

- exactProp
- chainPropTypes
- elementAcceptingRef
- elementTypeAcceptingRef
- refType
- deprecatedPropType
- requiredPropFactory
- unsupportedProp

It seems that none of the imports are pruned in people bundle (removing X.propTypes = Y not being enough). But to be confirmed.

PR #1770, opened 2020-05-13
Support "empty" value in pickers' UX

### LINKED ISSUE #1613: DatePicker - leave value blank/empty until user takes action?
Hello,

First, my apology if I'm posting in the wrong place or if this has been answered before. I did search, but didn't find any similar questions.

With that out of the way, my question is whether it's possible to use `DatePicker` (or `KeyboardDatePicker`, etc.) and have it initialize without automatically filling in the current date when the value passed in is `null`?

For my use case, the valid values of my DatePicker input are `null` as well as a user-specified date. The behavior seems to be that `null` values are turned into `nowRef.current` (via `useValueToDate` method (see: https://github.com/mui-org/material-ui-pickers/blob/master/lib/src/_shared/hooks/usePickerState.ts ) during initialization. 

For my case, I need the value to stay `null`/empty unless my user explicitly picks a date so I am looking for a way to prevent this automatic initialization from occurring.

Is there a prop or setting that I should look into? Or, perhaps a workaround that anyone can suggest?

Thanks.

dmtrKovalenko:
This PR closes #1613  

# User facing changes

## Breaking changes
* [UI] DatePicker and TimePicker will not render anything if `value={null}` and the user didn't take any action. (previously it has been rendering `defaultHighlight` or now)
* Remove `defaultHighlight` prop 

## Features
* DatePicker and TimePicker will not render anything if `value={null}` and the user didn't take any action. (previously it has been rendering `defaultHighlight` or now)
* Add new `toolbarPlaceholder` prop that will be displayed instead of day text when `value` is `null`. By default `"––"`


lib/src/TimePicker/TimePickerToolbar.tsx
```diff
@@ -122,7 +124,7 @@ export const TimePickerToolbar: React.FC<ToolbarComponentProps> = ({
             variant={clockTypographyVariant}
             onClick={() => setOpenView('hours')}
             selected={openView === 'hours'}
-            value={ampm ? utils.format(date, 'hours12h') : utils.format(date, 'hours24h')}
+            value={!date ? '--' : formatHours(date)}
```
oliviertassinari: Reverse the logic to avoid the negation?


```js
value={date ? formatHours(date) : '--'}
```

oliviertassinari (commented):
Nice improvement

Issue #1654, opened 2020-04-13
Implement string localization


oliviertassinari:
Developers should be able to follow the instructions in https://material-ui.com/guides/localization/ and get the wordings of the calendar localized.

In order to make it happen, we would likely need:

1. Move the sources to the mono-repository https://github.com/mui-org/material-ui/issues/19706 (easier to keep them in sync).
2. Rename the wording props to match the convention. For instance: `todayLabel` -> `todayText`.
3. Solve #1340 to enable injecting default wording props.

Issue #1340, opened 2019-10-03
Default props support missing


jasonkylefrank:
## Environment

| Tech                 | Version |
| -------------------- | ------- |
| @material-ui/pickers |  3.2.6       |
| material-ui          |  4.3.3       |
| React                |    16     |
| Browser              |         |
| Peer library         |   date-fns      |  |

## Question / issue
How can we supply default props to these controls?

One of the reasons we need to do this is that we have a customized MUI TextField component (we create a wrapper component around MUI's TextField) that we need every instance of these pickers to use.  This will ensure our theming and behavior customizations on the TextField will be consistent for all TextFields across our app.

So far I have been able to pass-in our custom TextField component via the `TextFieldComponent` prop.    But obviously that's not very DRY to have to do that on every single instance.   And we need to set it up so that other developers don't have to remember to pass that in every time.  

We accomplish default props with our other MUI controls by using their built-in theming system.  They have a `props` key (on the theme object) that can be utilized to set such default prop values.  See [this section](https://v4-3-3.material-ui.com/customization/globals/#default-props) of their docs.

So far I have tried to accomplish setting some default props for these picker controls by specifying a proper `key` in the theme object.  I can't seem to tell which `key` to use to affect the component which renders the TextField, so for now I'm just trying several possible MUI Pickers keys (none of which are working):

![image](https://user-images.githubusercontent.com/7502365/66158024-0712d880-e5f3-11e9-8ed6-40e939bcb8ef.png)

![image](https://user-images.githubusercontent.com/7502365/66158120-42150c00-e5f3-11e9-88dc-29ed0e70606c.png)

## Summary
So how do we set default props on these controls?

oliviertassinari:
@jasonkylefrank Thanks for the report, you are right, we miss the support for default props. We should 1. use `withStyles` over `makeStyles`, 2. provide a name to the stylesheets.

For instance: it won't for this public component:
https://github.com/mui-org/material-ui-pickers/blob/0d41c48aaa8b9b6f67e72deee9583340cd218b89/lib/src/views/Calendar/Calendar.tsx#L57

On the TypeScript side of the problem, we will need to implement: https://github.com/mui-org/material-ui/issues/19427.

PR #1773, opened 2020-05-14
Implement text props name convention

### LINKED ISSUE #1429: Add built-in localization support
We should support new localization feature. 

- [ ] Rename props to look conventional to material ui text props :
If we should rename the prop to `titleText` to something that starts to look like a convention:
- backIconButtonText
- labelRowsPerPage
- labelDisplayedRows
- nextIconButtonText
- getLabelText
- clearText
- closeText
- loadingText
- noOptionsText
- openText

_Originally posted by @oliviertassinari in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MzM1MDA1Njky/pull_request_reviews/more_threads_

- [ ] Make it possible to localize all formats from the localization feature. It should be a killer feature

dmtrKovalenko:
This PR resolves the first part of #1654 and probably closes #1429   

One question as for me. For now, all of our text props are spread through the different components, like `PickersModalDialog`, `ArrowSwitcher`, and `KeyboardDateInput`. It looks like that will be too awful to localize. 

Maybe we should add one prop like `texts` with all the text props combined in 1 object? But it is still possible to pass any prop to the just `DatePicker`, so support localization only for root components could be a workaround as well


oliviertassinari:
Regarding #1429, I would propose we close it once we have the counterpart of the localization strings in the mono repository, probably with a working demo in https://material-ui.com/guides/localization/#example. I think that we need to make sure it works, end-to-end. It would also be a great constrain to avoid regressions. What do you think?

> Maybe we should add one prop like texts with all the text props combined in 1 object?

Agree, I think that after 5 localization strings for a single component, we should start to wonder. With 10, a `texts` likrbprop sounds required. Should we call it `texts` or something else? 

There is one aspect I wonder about. How should we handle the strings duplication and fragmentation? Do you have an example of what the change to `locale.ruRu` look like to have to have the date picker components localized in Russian?

dmtrKovalenko:
I propose the following strategy:
1. Merge this PR with only renaming prop
2. Work for another PR with default props support (#1340)
3. Work on Localization PR with examples and support (and there combine props into new object)

oliviertassinari:
@dmtrKovalenko This plan sounds awesome!

lib/src/__tests__/DatePickerProps.test.tsx
```diff
@@ -63,15 +63,15 @@ describe('DatePicker - different props', () => {
...
+  it('showTodayButton – accept current date when "today" button is clicked', () => {
     const onCloseMock = jest.fn();
     const onChangeMock = jest.fn();
     const component = mount(
       <MobileDatePicker
         renderInput={props => <TextField {...props} />}
         autoOk
         showTodayButton
-        cancelLabel="stream"
+        cancelText="stream"
```
oliviertassinari, line 74: Any reason for providing a string in this test?

lib/src/_shared/PickerModalDialog.tsx
```diff
@@ -7,21 +7,48 @@ import Dialog, { DialogProps } from '@material-ui/core/Dialog';
...
-  onDismiss: () => void;
-  onClear: () => void;
-  onSetToday: () => void;
-  okLabel?: React.ReactNode;
-  cancelLabel?: React.ReactNode;
-  clearLabel?: React.ReactNode;
-  todayLabel?: React.ReactNode;
+export interface ExportedPickerModalProps {
+  /**
+   * "OK" button text
```
oliviertassinari: Maybe should make a sentence.

```suggestion
   * "OK" button text.
```

lib/src/_shared/PickerModalDialog.tsx
```diff
@@ -7,21 +7,48 @@ import Dialog, { DialogProps } from '@material-ui/core/Dialog';
...
+   * @default "TODAY"
+   */
+  todayText?: React.ReactNode;
+  /**
+   * Show clear action in picker dialog
+   * @default false
+   */
   clearable?: boolean;
+  /**
+   * If `true` today button will be displayed. **Note** that `showClearButton` has higher priority
```
oliviertassinari:
```suggestion
   * If `true`, the today button will be displayed. **Note** that `showClearButton` has a higher priority.
```

lib/src/_shared/PickerModalDialog.tsx
```diff
@@ -7,21 +7,48 @@ import Dialog, { DialogProps } from '@material-ui/core/Dialog';
...
+   * @default "CLEAR"
+   */
+  clearText?: React.ReactNode;
+  /**
+   * "TODAY" Text message
+   * @default "TODAY"
+   */
+  todayText?: React.ReactNode;
+  /**
+   * Show clear action in picker dialog
```
oliviertassinari:
```suggestion
   * If `true`, it shows the clear action in the picker dialog.
```

lib/src/__tests__/DatePickerProps.test.tsx
dmtrKovalenko: It is probably not the best way, but I am testing typescript bindings for props such way. 
Probably need to add some special kind of typescript definition tests

oliviertassinari: So the purpose is to have a test equivalent to, say, https://github.com/mui-org/material-ui/blob/a75c19f267edfdf71487764961de5c425096f47b/packages/material-ui/src/Typography/typography.spec.tsx?

oliviertassinari:
@dmtrKovalenko We have an ongoing thread on the naming convention for these props in https://github.com/mui-org/material-ui/issues/21018#issuecomment-629701596.

PR #1796, opened 2020-05-20
Global theme default props support

### LINKED ISSUE #1340: Default props support missing

dmtrKovalenko:
This PR closes #1340  
Also closes #1806 

## Description

- [x] Support global props patching through the theme.
- [x] Provide typescript typings for overriding default `createMuiThemeBehavior`.
- [x] Properly reexport all the component prop interfaces from the root.


docs/pages/guides/CssOverrides.example.tsx
```diff
@@ -1,9 +1,8 @@
-// @ts-nocheck
 import React, { useState } from 'react';
 import lightBlue from '@material-ui/core/colors/lightBlue';
 import { TextField } from '@material-ui/core';
-import { DatePicker } from '@material-ui/pickers';
 import { createMuiTheme, ThemeProvider } from '@material-ui/core';
+import { DatePicker, MaterialUiPickersDate } from '@material-ui/pickers';
```
oliviertassinari: I believe we encourage the following in the mono repository cc @eps1lon. Would it work? 

```suggestion
import { DatePickerProps } from '@material-ui/pickers';

DatePickerProps['value']
```

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
```diff
@@ -33,51 +35,54 @@ interface DateRangePickerToolbarProps
...
-  toggleMobileKeyboardView,
-  currentlySelectingRangeEnd,
-  setCurrentlySelectingRangeEnd,
-  startText,
-  endText,
-  toolbarTitle = 'SELECT DATE RANGE',
-}) => {
-  const utils = useUtils();
-  const classes = useStyles();
+export const DateRangePickerToolbar: React.FC<DateRangePickerToolbarProps> = withDefaultProps(
```
oliviertassinari, line 38: What's the advantage of makeStyles+withDefaultProps over withStyles? I think that the codebase should be as close as possible to the mono repository. It will be less migration work to do when merging the git repositories. It's also less room for mistake, less confusion for new contributors etc.

dmtrKovalenko: It is easier to migrate from the existed codebase and much easier to solve typing issues from injected `classes` prop.

I am not tent to rewrite everything to look the same as in the core repository. For example:  sure today everybody using `makeStyles` instead of `withStyles` because it is better for all cases. Just do the same

docs/pages/guides/CssOverrides.example.tsx
dmtrKovalenko: Yes it will

lib/src/DateRangePicker/DateRangePickerToolbar.tsx
oliviertassinari: Ok so the issue is with the TypeScript definitions. Can the typing we use in the mono-repository be used to solve the problem?

dmtrKovalenko: No really, I am sure that `default props` should not be a part of the theme. Because it is not a theme and we have already started discussion of moving them to separate provider, like `Configuration`. 

So decoupling styles and theme already should help for potential refactorings. And overall I am not seeing any advantages in `withStyles` over `withDefaultProps`, it is just more friendly to an existing codebase, faster and easier for props type inference

oliviertassinari: A couple of thoughts:

- 1. We have an opportunity to empower a simpler migration once we change the component structure strategy, later on, by using a single approach now.
- 2. Assuming we have a ConfigProvider, we could consider moving the default prop logic to this new object, however, it would hurt an issue with theming. it's relatively frequent to change default props that are only style-related: `disableRipple`, `color="secondary"`, `headlineMapping`, etc.
- 3. If we decouple, we should design the new API to minimize mistakes & inconsistencies. Right now, it's helped by bundling all the features under a single high-order component. For instance: https://github.com/mui-org/material-ui-pickers/pull/1796#discussion_r429960818.
- 4. In https://github.com/mui-org/material-ui/pull/15023, we explore a potential migration path to a hook API. I think that the conclusion was that it didn't worth the effort at that point in time. 
- 5. Looking at how the code look like now and how it would look like in v5, I see a high level of chance that it will go in the **opposite direction**: default props would move from a hoc to a hook (because it has a smaller footprint), styles will move from a hook to a hoc (because of styled-components).

@dmtrKovalenko Would you agree with the above points?

oliviertassinari: > So decoupling styles and theme already should help for potential refactorings. And overall I am not seeing any advantages in withStyles over withDefaultProps, it is just more friendly to an existing codebase, faster and easier for props type inference

Going back to your questions
- prop type inference is definitely a concern. Higher-order components are hard to type. However, we had to solve this problem in the past for the developers. I don't know this part very well, I believe/hope we did a great job in here, without any significant downside :)
- the advantage of withStyles over withDefaultProps is that it will enable the current API documentation generation stack to output the correct information, without the need for custom work. It will also enable the `getClasses` testing tool to take the component, without custom work.
- regarding the performance, I wouldn't expect any difference. I think that the perf win in mui-org/material-ui#15023 was related to the removal of a React component, nothing else, which we are introducing and paying now with `withDefaultProps`.

lib/src/DateRangePicker/DateRangeDelimiter.tsx
```diff
@@ -1,6 +1,12 @@
...
+import { withDefaultProps } from '../_shared/withDefaultProps';
 
-export const DateRangeDelimiter = styled(Typography)({
-  margin: '0 16px',
-});
+export const DateRangeDelimiter = withDefaultProps(
+  { name: 'MuiPickersDateRangeDelimiter' },
+  styled(Typography)({
+    margin: '0 16px',
+  })
```
oliviertassinari, line 9: missing style sheet name, cannot be customized globally (example of what I meant in [iii.](https://github.com/mui-org/material-ui-pickers/pull/1796#discussion_r429953553), it requires extra attention)

oliviertassinari:
It seems that the following components have a style sheet global override support but not a default prop global override support:

- PureArrowSwitcher
- PickerModalDialog
- PickerToolbar
- ToolbarText
- PureDateRangeDay
- DateRangePickerInput
- DateRangePickerViewDesktop
- DateTimePickerTabs
- CalendarHeader
- Month
- MonthSelection
- SlideTransition
- Year
- YearSelection
- ClockNumber
- StaticWrapper

What should be our strategy with them?

dmtrKovalenko:
We have already discussed it, not? 
Make all the component stylable, but provide default props only for exported components to prevent monkey patching internal components.

oliviertassinari:
@dmtrKovalenko Ok, I wasn't aware :)

dmtrKovalenko:
Just to clarify: if we will make all our components available for default props – we will not be able to change literally anything without breaking changes. For now, we cannot change class names and public API – and it seems fine.

PR #1797, opened 2020-05-21
Fix ternary condition for default mask values 

### LINKED ISSUE #1776: Mas warning on console  The mask "__/__/____ __:__ _M" you passed is not valid for the format used P p. Falling down to uncontrolled not-masked input.
## Environment

| @material-ui/pickers |   4.0.0-alpha.7      |
| material-ui          |    4.9.14     |
| React                |   16.12.0      |
| Browser              |     Chrome    |
| Peer library         |    "date-fns": "2.13.0"     |  |

## Steps to reproduce
Code sandbox example:  https://codesandbox.io/s/hungry-bohr-ugb2e?file=/src/App.js
Use basic example from the website.

## Expected behavior
Not expecting any warnings from the example
## Actual behavior
Mask warning on console "The mask "__/__/____ __:__ _M" you passed is not valid for the format used P p. Falling down to uncontrolled not-masked input. "
image

## Live example
Code sandbox example:  https://codesandbox.io/s/hungry-bohr-ugb2e?file=/src/App.js

Philipp91:


oliviertassinari (commented):
Could you add a regression test case with the enzyme mount API?

Philipp91:
I'm not familiar with that.

dmtrKovalenko:
Holey moley, that javascript

oliviertassinari (approved):
~I couldn't find how to reproduce this issue in the CI. ~

oliviertassinari (dismissed):
Actually, I can still reproduce the issue with the above change. It seems that it doesn't solve the problem.

Philipp91:
You're right. I attempted to fix the original issue by adding `mask="__.__.____ __:__"` to my picker, which matches my locale. Due to a bug that this PR hopefully fixes (though I haven't even tested that), the `mask` prop that I supplied manually was ignored, so the error message still appears.

However, this doesn't fix the underlying issue tracked in #1776, which is that the mask should be correct automatically. My application happens to be single-locale, so I can just hard-code the correct mask, but not every application can do so, and it shouldn't be necessary. So to _really_ fix #1776, the hard-coded masks that match the American date format need to be replaced with something more dynamic.

I have removed the "fixes #1776" claim from this PR. I still think it fixes a bug though.

oliviertassinari:
@Philipp91 Awesome, thanks for the extra context! I have tried to add a test case but without any luck. It seems that the mask is ignored. I don't understand why.

lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -9,6 +9,23 @@ import { DateTimePicker, DateTimePickerProps } from '../DateTimePicker/DateTimeP
 
 const format = process.env.UTILS === 'moment' ? 'MM/DD/YYYY HH:mm' : 'MM/dd/yyyy hh:mm';
 
+describe('DateTimePicker', () => {
+  describe('prop: mask', () => {
+    it('should handle the mark prop', () => {
```
Philipp91: typo
```suggestion
    it('should handle the mask prop', () => {
```

lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -9,6 +9,23 @@ import { DateTimePicker, DateTimePickerProps } from '../DateTimePicker/DateTimeP
...
+      const component = mount(
+        <DateTimePicker
+          renderInput={props => <TextField {...props} />}
+          label="DateTimePicker"
+          onChange={() => {}}
+          mask="__.__.____ _:__ _M"
+          value={utilsToUse.date('2018-01-01T00:00:00.000Z')}
+        />
+      );
+      expect(component.find('input').prop('value')).toBe('01.01.2018 1:00 AM');
```
Philipp91: This won't work unless you also somehow change the locale used in the utils. In fact, I'm not sure if the mask has anything to do with how the component maps Date values to strings for the TextField.

Perhaps you could change to `value={null}` above and change the assertion to output the correct mask?

lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -9,6 +9,23 @@ import { DateTimePicker, DateTimePickerProps } from '../DateTimePicker/DateTimeP
 
 const format = process.env.UTILS === 'moment' ? 'MM/DD/YYYY HH:mm' : 'MM/dd/yyyy hh:mm';
 
+describe('DateTimePicker', () => {
+  describe('prop: mask', () => {
+    it('should handle the mask prop', () => {
+      const component = mount(
+        <DateTimePicker
```
dmtrKovalenko: I think we need to write bug-pinning test here. Something like "should render custom mask even if `ampm` prop is provided"

```suggestion
        <DateTimePicker 
          ampm
```

dmtrKovalenko (dismissed):

oliviertassinari:
@Philipp91 I'm sorry. I'm making no progress on this one. I'm giving up. A few observations:

1. /demo/datetime-picker raises the warning mentioned in #1776.
2. I had forgotten how testing with enzyme was frustrating compared to testing the DOM. I couldn't figure out how to fire a change event on the input and assert. This fails, god 😆. @dmtrKovalenko I can't want for moving away from it 🙏

```jsx
      component.find('input').simulate('change', {
        target: {
          value: '12',
        },
      });
      expect((component.find('input').getDOMNode() as HTMLInputElement).value).toBe('12');
```

3. I couldn't observe the fixed behavior by the pull request. @Philipp91 Do you have a reproduction?

Philipp91:
To reproduce and verify that this PR works, even if it doesn't fully resolve the issue:

1. Patch https://github.com/Philipp91/material-ui-pickers/commit/f42bd3c92a501c55281e6516cee29f8aa391a721. (It deletes a bunch of example code to ensure that there's only one picker on the page to worry about.)
2. Open http://localhost:3001/demo/datetime-picker
3. Note how the patch changes the locale to `de-DE`. The error message appears saying that the `en-US` mask hard-coded in the library (namely `"__/__/____ __:__ _M"`) is invalid. Note how the reproduction code attempts to fix that by setting `mask="__.__.____ __:__"`, which would match the `de-DE` format. (Even if you remove that fix attempt, you'll see the warning.)
4. Patch this PR, then the warnings goes away. It still doesn't work without the explicit `mask="__.__.____ __:__"` (which it should), but the PR makes it a bit less broken.

I've [posted](https://github.com/mui-org/material-ui-pickers/issues/1776#issuecomment-632905414) a proposal for the real fix of the underlying issue. That real fix likely won't make the broken ternary expression go away, so it's worth fixing it anyway.

oliviertassinari:
@Philipp91 Awesome, thanks for the diff. I could figure a regression test out :).

oliviertassinari (approved):

dmtrKovalenko (approved):

oliviertassinari:
Rebased to fix the conflicts

oliviertassinari:
I'm excited about the new test 🎉  :) @Philipp91 Thanks so much for the help. We wouldn't have made it without you.

PR #1798, opened 2020-05-22
[test] CI should fail for new warnings

oliviertassinari:
This objective of this pull request is to surface an issue we will face with https://github.com/mui-org/material-ui/issues/19706. The equivalent behavior can be found in https://github.com/mui-org/material-ui/blob/14834adf8c07cd043a169cb486097f2363bbe420/test/utils/consoleError.js.

Waiting for #1776 to be resolved.

### DIFF
--- a/lib/src/__tests__/setup.js
+++ b/lib/src/__tests__/setup.js
@@ -36,3 +36,24 @@ console.error = (...args) => {
     throw new Error(message);
   }
 };
+
+// Convert any console warning into a thrown error
+const warn = console.warn;
+console.warn = (...args) => {
+  warn.apply(console, args);
+  if (args[0] instanceof Error) {
+    throw args[0];
+  } else {
+    // combine multi args into a string
+    const message = args
+      .map(value => {
+        if (typeof value === 'object') {
+          return JSON.stringify(value);
+        } else {
+          return value;
+        }
+      })
+      .join(' ');
+    throw new Error(message);
+  }
+};


lib/src/__tests__/setup.js
```diff
@@ -36,3 +36,24 @@ console.error = (...args) => {
     throw new Error(message);
   }
 };
+
+// Convert any console warning into a thrown error
+const warn = console.warn;
+console.warn = (...args) => {
+  warn.apply(console, args);
+  if (args[0] instanceof Error) {
```
dmtrKovalenko, line 44: I think it will not work for warn, right?

lib/src/__tests__/setup.js
```diff
@@ -36,3 +36,24 @@ console.error = (...args) => {
     throw new Error(message);
   }
 };
+
+// Convert any console warning into a thrown error
+const warn = console.warn;
+console.warn = (...args) => {
+  warn.apply(console, args);
```
dmtrKovalenko, line 43: Probably we can move this logic up to the reusable function, and use it for both console.error and console.warn

dmtrKovalenko (changes requested):

lib/src/__tests__/setup.js
oliviertassinari: I don't know. In the version of the module in the main repository, we assume the arguments is an array of strings.

```jsx
  console.error = (...args) => {
    // Can't use log as karma is not displaying them.
    console.info(...args);
    throw new Error(...args);
  };

  console.warn = (...args) => {
    // Can't use log as karma is not displaying them.
    console.info(...args);
    throw new Error(...args);
  };
```

We never had any issue with it yet. But it seems that the console methods can accept any object as input. What do you want to do?

lib/src/__tests__/setup.js
oliviertassinari: As you see fit, the module has a limited life expectancy.

oliviertassinari (commented):
Thanks for the review. I would propose we wait for the resolution of #1776 before rebasing this one and merging.

PR #1802, opened 2020-05-22
[core] Small changes

oliviertassinari:
- remove the null default value on the `className` prop, it's not required.
- use `it` over `test` as more frequently used in the codebase (they are aliases).
- remove the name of the root package to avoid confusion, it can't be published.
- sort ASC a couple of props.
- update the version on the root package to v4, the current working version.
- avoid alias 2212ccd: save a bit of bundle size and reduce cognitive overhead.
- normalize types comment 89ddefa: use the same code comment syntax as in the main repository
- sort prop asc d6fcaf6: make is easier to navigate
- remove empty function d60c239
- use shorthand notation 5d90550
- fix wrong @default API usage b21de7d

### DIFF
--- a/docs/prop-types.json
+++ b/docs/prop-types.json
@@ -294,7 +294,7 @@
   "DatePicker": {
     "views": {
       "defaultValue": null,
-      "description": "Array of views to show",
+      "description": "Array of views to show.",
       "name": "views",
       "parent": {
         "fileName": "material-ui-pickers/lib/src/Picker/SharedPickerProps.tsx",
@@ -307,7 +307,7 @@
     },
     "openTo": {
       "defaultValue": null,
-      "description": "First view to show",
+      "description": "First view to show.",
       "name": "openTo",
       "parent": {
         "fileName": "material-ui-pickers/lib/src/Picker/SharedPickerProps.tsx",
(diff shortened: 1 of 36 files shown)


dmtrKovalenko:
About `@type` 
This type needs to properly render type generated by typescript. Some time ago it was completely unreadable, now its more or less readable, but still not ideal ` "name": "Partial<OverrideProps<IconButtonTypeMap<{}, \"button\">, \"button\">>`

oliviertassinari:
@dmtrKovalenko I have reverted the change related to the types. It's would have had a too broad impact for the aim of the pull request: small details.

dmtrKovalenko (approved):

dmtrKovalenko:
Just wondering: do you have any automate solution for checking rules like dots (.) at the end of jsdoc comment and to only multi line comments? 

I suppose that it’s possible to create custom eslint rule.

oliviertassinari:
> Just wondering: do you have any automate solution for checking rules like dots (.) at the end of jsdoc comment and to only multi line comments?

@dmtrKovalenko Not I'm aware of. The closest thing we have is this eslint rule: https://github.com/mui-org/material-ui/blob/master/packages/eslint-plugin-material-ui/src/rules/docgen-ignore-before-comment.test.js. I think that the next step will be to rework the descriptions to match the style of the other components. We will need @mbrookes for double-checking the English side too :).

PR #1804, opened 2020-05-22
[test] Make sure the components can render with a default theme

oliviertassinari:
The objective is to prepare the migration into the main repository as much as to avoid theme requirement regressions until we do.


dmtrKovalenko (approved):
I am wondering. Do we have a test that checks opposite? That component can render with custom theme? And do we need it at all?

oliviertassinari:
There is one test that check if it work with the dark theme. I was hesitating to remove it because I have never seen a bug in this direction, but I figure you might have, so kept it.

PR #1817, opened 2020-05-25
[docs] make DateTimePickerValidation example work out of the box

dandv:
That example generates warnings [in the CSB](https://codesandbox.io/s/material-keyboarddatetimepicker-with-rhf-jqv1c?file=/index.tsx) console though (#1776).

> The mask "__/__/____ __:__ _M" you passed is not valid for the format used P p. Falling down to uncontrolled not-masked input. 

image

### DIFF
--- a/docs/pages/demo/datetime-picker/DateTimeValidation.example.jsx
+++ b/docs/pages/demo/datetime-picker/DateTimeValidation.example.jsx
@@ -1,12 +1,14 @@
 import React, { useState } from 'react';
 import { TextField } from '@material-ui/core';
 import { DateTimePicker } from '@material-ui/pickers';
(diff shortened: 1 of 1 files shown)


dandv:
Also, I can't select any time - #1818.

dmtrKovalenko:
We are not going to add the `<LocalizationProvider />` in each demo. At least for now, I think it is only adding white noise to the examples.

oliviertassinari:
> make DateTimePickerValidation example

@dandv We have decided to remove the built-in validation to leave more freedom to users #1730.

> That example generates warnings

See #1776.

> Also, I can't select any time - #1818.

Perfect, if you could open a new issue for each problem instead of "bundling" them, it would be much better.

PR #1829, opened 2020-05-25
[Calendar] Remove promise based loading in favor of `loading` prop

### LINKED ISSUE #1454: Remove Promise based api in favor of loading prop
Also use [Snackbar](https://material-ui.com/ru/components/snackbars/) instead of spinner
cc @oliviertassinari

dmtrKovalenko:
This PR closes #1454  

# User facing changes

## Breaking changes
* Removed ability to return `Promise` from the `onMonthChangeProp` to render spinner in the component. From now on use `loading` prop to render `LoadingComponent` in the calendar.
* Rename `loadingIndicator` to `LoadingComponent` 

## Features
* Improve the experience of server-side data loading by exposing `loading` prop that will render loading placeholder while data is loading.
* Provide new `CalendarSkeleton` component that can be used to achieve first-class UX experience for preloading data on month change: 
image


docs/pages/demo/datepicker/ServerRequest.example.jsx
```diff
@@ -4,23 +4,29 @@ import { TextField } from '@material-ui/core';
...
+
+    setSelectedDays(null)
+
+    const controller = new AbortController();
+    fetch(`/fakeApi/randomDate?month=${date.toString()}`, {
+      signal: controller.signal,
+    })
+      .then(res => res.json())
+      .then(({ daysToHighlight }) => {
+        setSelectedDays(daysToHighlight);
```
oliviertassinari: What if the component unmounts before the request resolves? I believe the set state call will warn/throw. The solution proposed in could be applied here too https://github.com/mui-org/material-ui-x/pull/5#discussion_r429676864

dmtrKovalenko: We are saving abort controller. So needs to make an effect that will abort request on unmount. thanks

oliviertassinari: It's one way to solve this. I have no idea what's the best approach. I guess it's good enough, no need to look further 👍

dmtrKovalenko: Interesting, but I think that [this](https://github.com/mui-org/material-ui/blob/90babbd13d8dc624ff23453a4e47a311e2758aa6/docs/src/pages/components/autocomplete/Asynchronous.tsx#L35) solution has one giant pitfall – it doesn't abort the request. 

Probably if we want to support IE we must not use fetch and promises at all.

oliviertassinari: Can requests be aborted? I mean, does it change something at the network layer?

dmtrKovalenko: Actually request/response model cannot be aborted, but browsers can prevent some calculations on the responses if they are aborted (like processing low-level body and CORS) 

Here network log when fast switching between months with request aborts: 
image

Without: 
image

oliviertassinari: Nice :)

PR #1834, opened 2020-05-28
Fix not working minDateTime and maxDateTime props

### LINKED ISSUE #1818: v4: minDateTime disables selecting any time
With `minDateTime` set to `new Date()`, I can't select any time, even if I select a date a few days in the future. All times are disabled:

image

## Environment

| @material-ui/pickers |  4.0.0-alpha7       |

## Live example

https://codesandbox.io/s/material-datetimepicker-mindatetime-doesnt-work-fymt5 but this can be seen on the `next` site, in [the demo](https://material-ui-pickers.mui-org.now.sh/demo/datetime-picker#date--time-validation).

dmtrKovalenko:
This PR closes #1818

### DIFF
--- a/lib/src/__tests__/DateTimePicker.test.tsx
+++ b/lib/src/__tests__/DateTimePicker.test.tsx
@@ -1,23 +1,14 @@
 import * as React from 'react';
-import DateFnsUtils from '@date-io/date-fns';
-import DateFnsLocaleDe from 'date-fns/locale/de';
-import LocalizationProvider from '../LocalizationProvider';
 import { ReactWrapper } from 'enzyme';
 import { TextField } from '@material-ui/core';
 import { mount as enzymeDefaultMount } from 'enzyme';
 import { MaterialUiPickersDate } from '../typings/date';
-import { createClientRender, fireEvent } from './createClientRender';
 import { mount, utilsToUse, mountPickerWithState } from './test-utils';
-import {
-  DesktopDateTimePicker,
-  DateTimePicker,
-  DateTimePickerProps,
-} from '../DateTimePicker/DateTimePicker';
+import { DateTimePicker, DateTimePickerProps } from '../DateTimePicker/DateTimePicker';
 
 const format = process.env.UTILS === 'moment' ? 'MM/DD/YYYY HH:mm' : 'MM/dd/yyyy hh:mm';
 
 describe('DateTimePicker', () => {
-  const render = createClientRender();
   let component: ReactWrapper<DateTimePickerProps>;
 
   const onCloseMock = jest.fn();
@@ -45,28 +36,6 @@ describe('DateTimePicker', () => {
     expect(component.find('WithStyles(ForwardRef(Dialog))').prop('open')).toBeTruthy();
   });
 
-  it('prop: mask – should take the mask prop into account', () => {
-    const { getByRole } = render(
-      <LocalizationProvider dateAdapter={DateFnsUtils} locale={DateFnsLocaleDe}>
-        <DesktopDateTimePicker
-          renderInput={props => <TextField autoFocus {...props} />}
-          mask="__.__.____ __:__"
-          onChange={() => {}}
-          value={null}
-        />
-      </LocalizationProvider>
-    );
-
-    const textbox = getByRole('textbox') as HTMLInputElement;
-    fireEvent.change(textbox, {
-      target: {
-        value: '12',
-      },
-    });
-
-    expect(textbox.value).toBe('12.');
-  });
-
   it('Should change internal state on update', () => {
     component.find('input').simulate('click');
     component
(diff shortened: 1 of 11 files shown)


lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -18,31 +18,6 @@ const format = process.env.UTILS === 'moment' ? 'MM/DD/YYYY HH:mm' : 'MM/dd/yyyy
...
-        target: {
-          value: '12',
-        },
-      });
-      expect(textbox.value).toBe('12.');
-    });
-  });
-});
-
-describe('e2e - DateTimePicker', () => {
```
oliviertassinari, line 45: nice, +1 for a single describe per test file

lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -83,7 +80,44 @@ describe('e2e - DateTimePicker', () => {
...
+        open
+        openTo="hours"
+        dateAdapter={new DateFnsUtils()}
+        renderInput={props => <TextField autoFocus {...props} />}
+        mask="__.__.____ __:__"
+        onChange={() => {}}
+        value={utilsToUse.date('2018-01-01T00:00:00.000Z')}
+        minDateTime={utilsToUse.date('2018-01-01T12:30:00.000Z')}
+      />,
+      { strict: false }
```
oliviertassinari: Instead of setting strict here, defining it at the createClientRender level could reduce duplication and give a chance to document why it's required. For instance:

```jsx
// StrictModeViolation: uses Grow
```

https://github.com/mui-org/material-ui/blob/d34d0eef762705e9230ea218221c87f7d763a5fe/packages/material-ui/src/Snackbar/Snackbar.test.js#L11-L14

lib/src/__tests__/createClientRender.tsx
```diff
@@ -59,24 +59,7 @@ function clientRender(
     baseElement,
     queries: { ...queries, ...customQueries },
     wrapper: Wrapper,
-  }) as any;
-
-  /**
-   * convenience helper. Better than repeating all props.
-   */
-  result.setProps = function setProps(props: any) {
```
oliviertassinari, line 67: Any reason for removing these helpers? You can look at the mono-repository codebase to see when it can be useful. We try to avoid using them, but I believe it's not always possible to avoid.

lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -83,7 +80,44 @@ describe('e2e - DateTimePicker', () => {
...
+    );
+
+    expect(getByLabelText('11 hours').getAttribute('aria-disabled')).toBe("true");
```
oliviertassinari: You might prefer (but I think that it's fine either way):

```suggestion
    expect(getByLabelText('11 hours)).to.have.attribute('aria-disabled', 'true');
```

At least, I believe @eps1lon would recommend it to improve the failing message if false

lib/src/__tests__/DateTimePicker.test.tsx
```diff
@@ -83,7 +80,44 @@ describe('e2e - DateTimePicker', () => {
...
     ).toBe('2018');
-    // expect(component.find('ToolbarButton').at(1).text()).toBe('Jan 3');
+  });
+
+  it('prop: minDateTime – hours is disabled by date part', () => {
+    const { getByLabelText } = render(
+      <DesktopDateTimePicker
+        open
+        openTo="hours"
+        dateAdapter={new DateFnsUtils()}
```
oliviertassinari: Haha smart

lib/src/__tests__/DateTimePicker.test.tsx
dmtrKovalenko: We do not have chai installed, I was thinking about adding https://github.com/testing-library/jest-dom

lib/src/__tests__/DateTimePicker.test.tsx
dmtrKovalenko: Probably need to add some kind of `renderWithAdapter` or `renderWithContex`, so it will not be needed.

lib/src/__tests__/createClientRender.tsx
dmtrKovalenko: It is about keep typescript typings. It is not possible to add something to the result, because it is `readonly` just quick feed

lib/src/__tests__/DateTimePicker.test.tsx
oliviertassinari: Right, because we need to test the other date libraries?

lib/src/__tests__/DateTimePickerTestingLib.test.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import { utilsToUse } from './test-utils';
+import { TextField } from '@material-ui/core';
+import { DesktopDateTimePicker } from '../DateTimePicker';
+import { createClientRender } from './createClientRender';
+import { fireEvent, screen } from '@testing-library/react';
+
+describe('DateTimePicker testing lib', () => {
```
oliviertassinari, line 8: What do you think of naming it

```jsx
describe('<DateTimePicker />', () => {
```

while renaming the older tests with "legacy" inside them? :)

lib/src/__tests__/DateTimePickerTestingLib.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        minDateTime={utilsToUse.date('2018-01-01T12:30:00.000Z')}
+      />
+    );
+
+    expect(screen.getByLabelText('25 minutes').getAttribute('aria-disabled')).toBe('true');
+    expect(screen.getByLabelText('35 minutes').getAttribute('aria-disabled')).toBe('false');
+  });
+
+  it('prop: minDateTime – hours is disabled by date part', () => {
+    screen.debug();
```
oliviertassinari, line 49: left over?

Issue #1778, opened 2020-05-16
[RFC] Component namespace


oliviertassinari:
During the last meeting on the date picker components between me and @dmtrKovalenko, we have started a discussion around the component namespace, at the Material-UI scale. 

The discussion has started from the requirement to use `withStyles` over `makeStyles` in order to benefit from https://material-ui.com/customization/globals/#default-props, which the components here don't use yet (IMHO we should be using the HOC and only to match the other components).

**How should we name the picker components?**

As far as I understand the situation, we have the following constraints:

- `theme.props` is a global namespace, so the name of the style sheet should account for it: no collision.
- the generate CSS has global class names, so the name of the style sheet should account for it (`theme.overrides`): no collision.
- We want to use the same component name for the default props overrides (`theme.props`) that the style sheet name for the CSS overrides (`theme.overrides`). It makes it easier to understand.

Now, this leaves us with this following tradeoff. 

- option 1 (current): we prefix all the style sheets of the pickers package with, say `Pickers`. But we don't prefix the name of the exported components to keep them shorter, people would have to remember that if the component comes from the pickers package, they need to add a prefix.

```jsx
import Day from `@material-ui/pickers/Day`; // MuiPickersDay
import DatePicker from `@material-ui/pickers/DatePicker`; // MuiPickersDatePicker
```

- option 2: we go full steam on the global namespace. We keep the exported name of the component identical to the name of the style sheet. We prefix the components with `Pickers` that might conflict in the future, e.g. `Day`.

```diff
-import Day from `@material-ui/pickers/CalendarDay`; // MuiPickersDay
+import PickersDay from `@material-ui/pickers/PickersDay`; // MuiPickersDay
-import DatePicker from `@material-ui/pickers/DatePicker`; // MuiPickersDatePicker
+import DatePicker from `@material-ui/pickers/DatePicker`; // MuiDatePicker
```

Today, I would be all 💯 in favor of option 2 because it's simpler for the developers to have a direct mapping between the stylesheet name and exported name. However, I have a few concerns with how it grows with the number of components we support, say, all the enterprise components and their lower-level components. Any preference?

cc @mui-org/core

ryancogswell:
I strongly prefer option 2. I think even as the number of components grows, it would be worthwhile to avoid duplication of names. I think it would be confusing (for logging issues, StackOverflow questions, finding documentation, etc.) to have an enterprise component with the exact same name as a non-enterprise component.

[...]

Issue #1768, opened 2020-05-12
Improve popup placement


oliviertassinari:
This is a subset of #1648 focused on the placement of the popup on desktop. I believe we should apply the following changes:

1. Similar to https://github.com/mui-org/material-ui/pull/18289, it would make sense to move the `ref` from the root of the renderInput params to the `InputProps` key. This would avoid showing the helper message when the popup is open.
1. The popper should be positioned to start on the right side, not at the center.

<img src="https://user-images.githubusercontent.com/3165635/79081168-29607e00-7d1b-11ea-8546-efea151e1566.png" width="400" />

PR #1850, opened 2020-06-02
Migrate to Popper from Popover

### LINKED ISSUE #1426: Use Popper over Popover for desktop?
Just noticed this but when date picker is opened in dropdown-like form the scroll of the page is hidden. Causes the design to shift slightly.

### LINKED ISSUE #1835: PopoverProps are not passed down to Popover component

## Environment

| @material-ui/pickers |     v4.0.0-alpha.7    |
| material-ui          |    4.9.13     |
| React                |         |
| Peer library         |         |  |

## Steps to reproduce

1. Try to pass PopoverProps to DatePicker component (like anchorOrigin or transformOrigin). For example, like this:
```js
      <LocalizationProvider dateAdapter={DateFnsUtils}>
        <DatePicker
          label="Basic example"
          value={selectedDate}
          onChange={(date) => handleDateChange(date)}
          renderInput={(props) => <TextField {...props} />}
          PopoverProps={{
            anchorOrigin: {
              vertical: "top",
              horizontal: "right",
            },
            transformOrigin: {
              vertical: "top",
              horizontal: "right",
            }
          }}
        />
      </LocalizationProvider>
```
## Expected behavior

Popover (ie DatePicker) should reposition based on passed anchorOrigin/transformOrigin props (vertical/bottom).

## Actual behavior

Popover position does not change, On versions 3.x PopoverProps worked nicely with variant="inline". Am I missing something here?

dmtrKovalenko:
This PR closes #1426 and also closes #1835


dmtrKovalenko:
One difference I am seeing – Popover doesn't return the focus back to the open button when closed. Any ideas @oliviertassinari

lib/rollup.config.js
```diff
@@ -39,6 +39,7 @@ const globals = {
   '@material-ui/core/Modal/TrapFocus': 'material-ui.TrapFocus',
```
oliviertassinari:
```suggestion
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,121 @@
...
+      zIndex: theme.zIndex.modal,
+    },
+    paper: {
+      transformOrigin: 'top center',
+      '&:focus': {
+        outline: 'auto',
+        [IS_TOUCH_DEVICE_MEDIA]: {
+          outline: 0,
+        },
+      },
```
oliviertassinari: We can remove redundant default for simplicity until we solve the root issue.

```suggestion
      [IS_TOUCH_DEVICE_MEDIA]: {
        outline: 0,
      },
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,121 @@
...
+    },
+    topTransition: {
+      transformOrigin: 'bottom center',
+    },
+  }),
+  { name: 'MuiPickersPopper' }
```
oliviertassinari, line 51: `s` or not? It doesn't match the name of the file. Which one should be fix?

```suggestion
  { name: 'MuiPickerPopper' }
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,121 @@
...
+    >
+      {({ TransitionProps, placement }) => (
+        <TrapFocus
+          open={open}
+          disableAutoFocus={role === 'tooltip'}
+          disableEnforceFocus={role === 'tooltip'}
+          isEnabled={() => true}
+          getDoc={() => paperRef.current?.ownerDocument ?? document}
+        >
+          <TransitionComponent {...TransitionProps} timeout={350}>
```
oliviertassinari: Where is 350 coming from, why overriding the default?

An order switch to allow customizations:

```suggestion
          <TransitionComponent timeout={350} {...TransitionProps}>
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,121 @@
...
+export const PickerPopper: React.FC<PickerPopperProps> = ({
+  role,
+  PopperProps,
+  TrapFocusProps,
+  TransitionComponent = Grow,
+  open,
+  innerRef = null,
+  anchorEl,
+  children,
+  onClose,
```
oliviertassinari: Sort asc

```suggestion
  anchorEl,
  children,
  innerRef = null,
  onClose,
  open,
  PopperProps,
  role,
  TransitionComponent = Grow,
  TrapFocusProps,
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,121 @@
...
+  ...other
+}) => {
+  const classes = useStyles();
+  const paperRef = React.useRef<HTMLElement>(null);
+  const handlePopperRef = useForkRef(paperRef, innerRef);
+
+  useGlobalKeyDown(open, {
+    [keycode.Esc]: onClose,
```
oliviertassinari, line 73: Do we need to listen globally to the <kbd>Escape</kbd> key? What prevent to scope the listener to the container?

lib/src/wrappers/DesktopTooltipWrapper.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+      onDismiss();
+    });
+  };
+
+  return (
+    <WrapperVariantContext.Provider value="desktop">
+      <KeyboardDateInputComponent {...DateInputProps} containerRef={inputRef} onBlur={handleBlur} />
+
+      <PickerPopper
```
oliviertassinari: `PickerPopper ` seems to be only used in one place. What about inlining it here, removing abstraction?

lib/src/wrappers/DesktopWrapper.tsx
```diff
@@ -1,77 +1,54 @@
...
-  const ref = React.useRef<HTMLDivElement>(null);
-  const classes = useStyles();
+  const isFocusedRef = React.useRef(false);
+  const dialogRef = React.useRef<HTMLDivElement>(null);
+  const inputRef = React.useRef<HTMLDivElement>(null);
 
   return (
     <WrapperVariantContext.Provider value="desktop">
-      <KeyboardDateInputComponent {...DateInputProps} containerRef={ref} />
+      <KeyboardDateInputComponent {...DateInputProps} containerRef={inputRef} />
```
oliviertassinari: Shouldn't listen to the "Input" rather than the "container"? The popup open bellow the helper text, I believe it should open below the `<input />` element.

lib/src/_shared/PickerPopper.tsx
oliviertassinari: Maybe it's a good opportunity to solve #1768? I don't know.

lib/src/wrappers/DesktopTooltipWrapper.tsx
oliviertassinari: I think that if we were to add an abstraction, it would be interesting to include the date range picker in it.

oliviertassinari (commented):
I did a quick review. I'm happy that we move forward in this direction :).

> One difference I am seeing – Popover doesn't return the focus back to the open button when closed. Any ideas @oliviertassinari?

I have also noticed that it breaks the scroll position.

lib/src/wrappers/DesktopWrapper.tsx
oliviertassinari: Maybe it's a good opportunity to solve #1768? I don't know.

lib/src/wrappers/DesktopTooltipWrapper.tsx
dmtrKovalenko: There are 2 places

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: It is not always focused on popper (DateRangePicker)

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: I think we will have some kind of name consolidation PR, so it will be easier to find all names by "MuiPickers"

oliviertassinari: Sounds great, I love fighting entropy with consistency 😁

lib/src/_shared/PickerPopper.tsx
oliviertassinari: The date range picker case makes me think of https://github.com/mui-org/material-ui/issues/17279.
For the date picker, I was under the assumption that the component will always close on blur. Hence, we can listen to the key event on the root.

lib/rollup.config.js
dmtrKovalenko: ???

oliviertassinari: #1857

BrentFarese:
@dmtrKovalenko any idea on the next release time line?  Just curious as we're checking this PR and waiting for it to hit.  Thanks!  Great library!

dmtrKovalenko:
Working on the issues caused by full migration to popper.

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: Will be suppressed by #1928

lib/rollup.config.js
```diff
@@ -39,6 +39,7 @@ const globals = {
   '@material-ui/core/Unstable_TrapFocus': 'material-ui.Unstable_TrapFocus',
   '@material-ui/core/Grow': 'material-ui.Grow',
   '@material-ui/core/Popper': 'material-ui.Popper',
+  '@material-ui/core/Unstable_TrapFocus': 'material-ui.Unstable_TrapFocus',
```
oliviertassinari:
```suggestion
```

lib/src/__tests__/DateRangePicker.test.tsx
```diff
@@ -1,76 +1,22 @@
...
 import { TextField, TextFieldProps } from '@material-ui/core';
-import { DesktopDateRangePicker } from '../DateRangePicker/DateRangePicker';
+import { createClientRender, fireEvent } from './createClientRender';
 
 const defaultRangeRenderInput = (startProps: TextFieldProps, endProps: TextFieldProps) => (
   <>
-    <TextField {...startProps} />
-    <TextField {...endProps} />
+    <TextField data-mui-test="start-input" {...startProps} />
+    <TextField data-mui-test="end-input" {...endProps} />
```
oliviertassinari: What about using `getAllByRole('textbox')[0]` / `getAllByRole('textbox')[1]` instead adding a `data-mui-test` attribute?

lib/src/__tests__/DateRangePicker.test.tsx
```diff
@@ -80,11 +26,9 @@ describe('DateRangePicker', () => {
...
 
-    component
-      .find('input')
-      .at(1)
-      .simulate('focus');
-
-    expect(component.find('div[role="tooltip"]').length).toBe(1);
+    fireEvent.focus(getByMuiTest('start-input'));
+    const el = await screen.findAllByRole('tooltip')
+    screen.debug(el[1]);
```
oliviertassinari: @eps1lon It could be interesting to have an eslint rule that fails if a debug is left in the tests.

lib/src/_shared/PickerPopper.tsx
oliviertassinari: Ok so if we refer to #1778, I think that we can use `MuiPickersPopper`, but rename the file to 
`lib/src/_shared/PickersPopper.tsx`

lib/src/_shared/PickerPopper.tsx
oliviertassinari: It seems that we have new lines of code, I don't see the immediate link with #1928. Could you expand? :).

lib/src/views/Clock/ClockNumbers.tsx
```diff
@@ -7,7 +7,7 @@ interface GetHourNumbersOptions {
   ampm: boolean;
   utils: IUtils<MaterialUiPickersDate>;
   date: MaterialUiPickersDate;
-  onChange: (value: number, isFinish?: boolean) => void;
+  onChange: (value: number, isFinish?: boolean | symbol) => void;
```
oliviertassinari: What's `symbol` is for?

```suggestion
  onChange: (value: number, isFinish?: boolean) => void;
```

lib/src/__tests__/DatePicker.test.tsx
```diff
@@ -24,7 +22,7 @@ describe('e2e - DatePicker default year format', () => {
       <DatePicker
         renderInput={props => <TextField {...props} />}
         DialogProps={{}}
-        PopoverProps={{}}
+        PopperProps={{}}
```
oliviertassinari: It seems we don't need these two props.

lib/src/__tests__/DateRangePicker.test.tsx
```diff
@@ -1,76 +1,22 @@
...
-        renderInput={defaultRangeRenderInput}
-        calendars={3}
-        onChange={jest.fn()}
-        value={[
-          utilsToUse.date('2018-01-01T00:00:00.000'),
-          utilsToUse.date('2018-01-31T00:00:00.000'),
-        ]}
-      />
-    );
+describe('<DatePicker />', () => {
```
oliviertassinari:
```suggestion
describe('<DateRangePicker />', () => {
```

lib/src/__tests__/TimePicker.test.tsx
```diff
@@ -139,7 +139,7 @@ describe('e2e - TimePicker with seconds', () => {
   });
 });
 
-describe('e2e - Timepicker view navigation', () => {
+describe.skip('e2e - Timepicker view navigation', () => {
```
oliviertassinari, line 143:
```suggestion
describe('e2e - Timepicker view navigation', () => {
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,143 @@
...
+  TrapFocusProps?: Partial<TrapFocusProps>;
+  anchorEl: PopperProps['anchorEl'];
+  open: PopperProps['open'];
+  onClose: () => void;
+  onOpen: () => void;
+}
+
+export const useStyles = makeStyles(
+  theme => ({
+    popper: {
```
oliviertassinari:
```suggestion
    root: {
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,143 @@
...
+          <TransitionComponent {...TransitionProps} onEntered={onOpen}>
+            <Paper
+              tabIndex={-1}
+              elevation={8}
+              ref={handlePopperRef}
+              className={clsx(classes.paper, {
+                [classes.topTransition]: placement === 'top',
+              })}
+              onBlur={handleBlur}
+              {...other}
```
oliviertassinari: `other` should be spread on the root element: Popper.

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,143 @@
...
+    });
+  };
+
+  return (
+    <Popper
+      transition
+      role={role}
+      open={open}
+      anchorEl={anchorEl}
+      placement="bottom"
```
oliviertassinari:
```suggestion
      placement="bottom-start"
```

per #1648

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,143 @@
...
+      className={clsx(classes.popper, PopperProps?.className)}
+      {...PopperProps}
+    >
+      {({ TransitionProps, placement }) => (
+        <TrapFocus
+          open={open}
+          disableAutoFocus
+          disableEnforceFocus={role === 'tooltip'}
```
oliviertassinari, line 120: Looking at the WAI-ARIA example, it seems we can disable enforce focus

```suggestion
          disableEnforceFocus
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,143 @@
...
+          isEnabled={() => true}
+          getDoc={() => paperRef.current?.ownerDocument ?? document}
```
oliviertassinari, line 122: I would expect it to the always defined. If not, it might hide something else

```suggestion
          getDoc={() => paperRef.current.ownerDocument}
```

lib/src/views/Calendar/YearSelection.tsx
```diff
@@ -58,26 +58,14 @@ export const YearSelection: React.FC<YearSelectionProps> = ({
...
-      try {
-        selectedYearRef.current.scrollIntoView({
-          block: wrapperVariant === 'static' ? 'nearest' : 'center',
-        });
-      } catch (e) {
-        // call without arguments in case when scrollIntoView works improperly (e.g. Firefox 52-57)
-        selectedYearRef.current.scrollIntoView();
-      }
-    }
-  }, []); // eslint-disable-line
```
oliviertassinari, line 79: nice!

oliviertassinari (changes requested):

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: https://github.com/mui-org/material-ui/blob/e0bd7326d160ac7757597786717eba77f986f2e4/packages/material-ui/src/utils/ownerDocument.js#L4

dmtrKovalenko: Moreover document if optional field in typings, so there are some cases when it is undefined. And we must handle them

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: It will make possible to tab out of <TrapFocus> which makes it completely useless

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: Looks awful as for me, maybe we need to provide shorthand prop for this case? To prevent using `PopperProps={{ placement: "bottom-start" }}`

image

lib/src/views/Clock/ClockNumbers.tsx
dmtrKovalenko: Will be not needed by #1923

lib/src/_shared/PickerPopper.tsx
oliviertassinari: `placement="bottom-start"` should be the default per the convention in the community. However, I think that developers should also be able to customize the default placement. For instance, we have seen people asking for `placement="bottom-end"`

lib/src/__tests__/DatePicker.test.tsx
dmtrKovalenko: For now, it is only checked that this prop is available. Should be moved to typescript.

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: Where says that it should be default? It looks askew on the page
image

lib/src/_shared/PickerPopper.tsx
oliviertassinari: Private component or not, I think that they should all follow this rule. I believe we already enforce it.

lib/src/__tests__/TimePicker.test.tsx
dmtrKovalenko: Haha interesting, after installing some package it started failing with heap overhead error. Needs some investigation – unrelated to this PR

lib/src/_shared/PickerPopper.tsx
oliviertassinari: By "tab out", are you referring to <kbd>Tab</kbd>? This sounds like an unrelated issue. The problem I see with the current logic is that if I click on another element of the page, it closes the popup but restores back the focus to calendar icon. Should we expect the focus to be set on the interacted element instead?

lib/src/_shared/PickerPopper.tsx
oliviertassinari: The defensive logic for ownerDocument is meant to handle the case where the argument is already the document. In our case, it's a nested element in the page, it serves no purpose.

Regarding handling the optional field, `// @ts-ignore` sounds like the appropriate solution here, however, it looks like `getDoc` is call in an unsafe manner. It seems that we have to account for it here anyway. I'm resolving the comment;

lib/src/__tests__/DatePicker.test.tsx
oliviertassinari: Agree, let's move it to a different place. We should keep the tests as focused and isolated as possible. Testing one thing at the time.

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: I cannot agree on such unsafe way. If typescript says that there is a possible case where `ownerDocument`, `ref.current` can be undefined – we **must** handle this case. 

This is, I suppose, the eternal problem of the javascript ecosystem. But this is how sound systems are built, and we as a core thing for a lot of projects cannot crash because of `cannot read property X of undefined`. We just cannot

lib/src/_shared/PickerPopper.tsx
dmtrKovalenko: I am playing locally with only `disableEnforceFocus` and it allows to tab out of picker and breaks TrapFocus at all

lib/src/_shared/PickerPopper.tsx
oliviertassinari: Yes, exactly, now, it also lacks 3 other important changes:

1. The reference point is wrong: #1768 (wrong ref).
2. The spacing of the desktop is wrong #1648 (the days take too much space).
3. The leading space at the bottom of the popup seems x3 too much.

Once done, it should be 👌 😍

lib/src/_shared/PickerPopper.tsx
```diff
@@ -49,7 +49,7 @@ export const useStyles = makeStyles(
       transformOrigin: 'bottom center',
     },
   }),
-  { name: 'MuiPickersPopper' }
+  { name: 'MuiPickerPopper' }
```
oliviertassinari:
```suggestion
  { name: 'MuiPickersPopper' }
```

Per the conclusion of #1778

lib/src/_shared/PickerPopper.tsx
oliviertassinari: Could it be an issue with the blur logic?

oliviertassinari: I could reproduce the issue on https://codesandbox.io/s/lucid-davinci-om21n?file=/src/App.js. It does sound like a bug with TrapFocus.

oliviertassinari: https://github.com/mui-org/material-ui/issues/21569 in case you want to work on it 👌

lib/src/__tests__/DateRangePicker.test.tsx
dmtrKovalenko: data-mui-test here set 1 time, and makes test a little bit easier to read, no? 
But this will work as weell

oliviertassinari: I think that the rule of thumb is that if we can use an accessibility feature, we should. It has more priority. Otherwise, we fallback to a data attribute.

oliviertassinari: Basically, following the same priority rule as [https://testing-playground.com/](https://testing-playground.com/?markup=DwGwhgRgpiAEBmB7ATgXgEQFcDOVkDswBbKdAPgFVcDipgB6caEMgKGAEt8AHTAF1YcAJhhx5CJdKwlRR1GVO7gAxlAAWiEELwYNAd1jYNmLbACSsZWBBwAnokwB%2BKULB8wAWj5RsfYaIVWPz4QWXQofG9kWHtMaLEaSVZrPgwIqJiHePlaKQA3a0wwgCtENXwPIURSINtuMO8ADz4pejZgCH4%2BRHxYGQxsDgBzfExudFg%2BOrDsTAgiDhayQZGxhk6%2Bbvx2oQ482Fd3Lx8-EXQ1GBtECd9bUIw9YT41AC5YAEYABm5GgG5YC7DNR8N5fH7-CBgZQAayGyAc%2BCEb2QUCEvwmbSAA&query=M4YwTgphB2B0DmEAuAhAngFQsJBJAJgBQDkAFhADZUD2xAlEA).

lib/src/_shared/PickerPopper.tsx
oliviertassinari: TypeScript can be as smart as the information he gets. He has no knowledge of the runtime guarantees. In this case. `ref.current` being defined is a constraint React enforce if accessed from an effect or layout effect. I think that the correct approach is `// @ts-ignore`, cc @eps1lon to get a broader point of view.

This is no different to what we explored in [2017 with flow](https://github.com/mui-org/material-ui/pull/9453). I think that we learned that it was better not to use defensive logic when we know the guarantee. If done otherwise, at best we add dead logic, at worse, we hide root issues.

Regarding `.ownerDocument` being defined is a constraint, if we look at the definitions, it's [null for a document](https://github.com/microsoft/TypeScript/blob/cc7cb9eadb10a56687513d4fc3a67d5210b387bb/src/lib/dom.generated.d.ts#L4546) and [defined for an element](https://github.com/microsoft/TypeScript/blob/cc7cb9eadb10a56687513d4fc3a67d5210b387bb/src/lib/dom.generated.d.ts#L5108). In our case, we always deal with an element, if will always be defined (if the element exists in the first place).

dmtrKovalenko: I think it is a problem of developers who never worked with languages with a really strong type system (Rust for example). 

React guarantee that refs are filled, but it is not always true. In case when document is mounted by timeout, or unmounted from Dom manually. 

If you are consciously do not guard from such cases — the final product can crash in the most unexpected moment.

lib/src/_shared/PickerPopper.tsx
oliviertassinari: Bump

lib/src/__tests__/DatePicker.test.tsx
oliviertassinari:
```suggestion
```

lib/src/__tests__/DatePicker.test.tsx
```diff
@@ -24,7 +22,7 @@ describe('e2e - DatePicker default year format', () => {
       <DatePicker
         renderInput={props => <TextField {...props} />}
         DialogProps={{}}
```
oliviertassinari:
```suggestion
```

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { useForkRef } from '@material-ui/core/utils';
+import { makeStyles } from '@material-ui/core/styles';
+import { useGlobalKeyDown, keycode } from './hooks/useKeyDown';
+import { IS_TOUCH_DEVICE_MEDIA } from '../constants/dimensions';
+import { executeInTheNextEventLoopTick } from '../_helpers/utils';
+import { TransitionProps } from '@material-ui/core/transitions/transition';
+
+export interface ExportedPickerPopperProps {
+  /**
+   * Popper props passed down to material-ui [Popper](https://material-ui.com/api/popper/#popper-api) component.
```
oliviertassinari: It's implicit

```suggestion
   * Popper props passed down to [Popper](https://material-ui.com/api/popper/) component.
```

lib/src/_shared/PickerPopper.tsx
oliviertassinari: bump

lib/src/_shared/PickerPopper.tsx
oliviertassinari: bump

lib/src/_shared/PickerPopper.tsx
oliviertassinari: #1928 was merged, should we update here?

lib/src/_shared/PickerPopper.tsx
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  return (
+    <Popper
+      transition
+      role={role}
+      open={open}
+      anchorEl={anchorEl}
+      className={clsx(classes.root, PopperProps?.className)}
+      {...PopperProps}
```
oliviertassinari:
```suggestion
      {...PopperProps}
      className={clsx(classes.root, PopperProps?.className)}
```

e2e/component/KeyboardNavigation.spec.tsx
```diff
@@ -5,6 +5,7 @@ import { DesktopDatePicker, MobileDatePicker, TimePicker } from '@material-ui/pi
 describe('Keyboard navigation', () => {
   context('DatePicker', () => {
     function testCalendarKeyboardNavigation() {
+      console.log(document.activeElement);
```
oliviertassinari:
```suggestion
```

oliviertassinari (approved):

diegovillacis10:
Since which version this change will be available?

dmtrKovalenko:
It is already available with the latest v4 **alpha**

PR #1857, opened 2020-06-04
Fix TrapFocus import with @material-ui/core v4.10.1

### LINKED ISSUE #1849: incompatible with @material-ui/core": "4.10.1
see https://github.com/mui-org/material-ui/pull/21201

```
./node_modules/@material-ui/pickers/DateTimePicker.js
Module not found: Can't resolve '@material-ui/core/Modal/TrapFocus' in '/Users/michaelli/Workspaces/jtd-admin/node_modules/@material-ui/pickers'
```

dmtrKovalenko:
This PR closes #1849

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1835,14 +1835,14 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
-"@babel/runtime@^7.1.2", "@babel/runtime@^7.3.1", "@babel/runtime@^7.4.4", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.3", "@babel/runtime@^7.7.4", "@babel/runtime@^7.8.3", "@babel/runtime@^7.8.4", "@babel/runtime@^7.8.7", "@babel/runtime@^7.9.6":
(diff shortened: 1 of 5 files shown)


oliviertassinari (commented):
We should increase the peer-dependency scope too, so that developers get a warning:

https://github.com/mui-org/material-ui-pickers/blob/fb9ca87d0072527ff71ecc9b995c90ea9d633f08/lib/package.json#L34-L35

yarn.lock
```diff
@@ -9224,6 +9237,16 @@ jss@*, jss@10.1.1, jss@^10.0.0, jss@^10.0.3:
     is-in-browser "^1.1.3"
     tiny-warning "^1.0.2"
 
+jss@10.2.0, jss@^10.0.3:
```
oliviertassinari: There is a nonnegligible amount of dependencies duplication in this yarn.lock. While we have built-in support for solving this problem in the CI of the main repository, I think that for this time, https://github.com/atlassian/yarn-deduplicate#usage will be good enough.
You should see a diff like this: `1 file changed, 99 insertions(+), 1172 deletions(-)`.

PR #1870, opened 2020-06-08
Move InputProps to last for override default input adornment

casamia918:
## Description

At now, InputProps.endAdornment is not working in KeyboardDatePicker component. Even though open picker button is desired behavior of KeyboardDateInput component, it needs to be customized.


dmtrKovalenko:
This is by design. We don’t provide a way to override the adornment. 

Moreover for v4 we have `renderInput` API that makes possible to do everything.

casamia918:
> This is by design. We don’t provide a way to override the adornment.
> 
> Moreover for v4 we have `renderInput` API that makes possible to do everything.

Ok thanks. Then... I'd rather use my own branch in my project, at now.

PR #1873, opened 2020-06-09
[typescript] Fix theme overrides type failing

### LINKED ISSUE #1858: TypeScript overrides are not working with v4.0.0-alpha.8
## Environment

| @material-ui/pickers |     v4.0.0-alpha.8    |
| material-ui          |     4.10.0    |
| TypeScript           |    3.9.3     |
| React                |     16.13.1    |
| Browser              |    All     |
| Peer library         |    date-fns     |

## Steps to reproduce

Following the [TypeScript example](https://next.material-ui-pickers.dev/guides/css-overrides#typescript) for overriding CSS causes a type error in `@material-ui/core`.

I have the following in my TypeScript declaration file:

```typescript
import { MuiPickersComponentsToClassName } from '@material-ui/pickers/typings/overrides';

declare module '@material-ui/core/styles/overrides' {
  export interface ComponentNameToClassKey extends MuiPickersComponentsToClassName {}
}
```

> Side note: The example also references the `src` path, which is also doesn't work.   Removing `src` seems to make that import work.

## Expected behavior

`tsc` should pass.

## Actual behavior

When I run `tsc` on my project, I get the following error:

```
$ tsc -p tsconfig.json
../../../node_modules/@material-ui/core/styles/overrides.d.ts:100:64 - error TS2344: Type 'ComponentNameToClassKey[Name]' does not satisfy the constraint 'string'.
  Type '"fixed" | "required" | "default" | "square" | "active" | "disabled" | "track" | "error" | "progress" | "message" | "middle" | "media" | "hidden" | "body" | "button" | "caption" | ... 486 more ... | undefined' is not assignable to type 'string'.
    Type 'undefined' is not assignable to type 'string'.

100   [Name in keyof ComponentNameToClassKey]?: Partial<StyleRules<ComponentNameToClassKey[Name]>>;
                                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

If I remove the picker overrides, `tsc` passes.

bopfer:
Closes #1858

### DIFF
--- a/lib/src/typings/overrides.ts
+++ b/lib/src/typings/overrides.ts
@@ -34,35 +34,35 @@ type Classes<T> = T extends string
   : never;
 
 export interface MuiPickersComponentsToClassName {
-  MuiPickersDay?: Classes<typeof DayStyles>;
-  MuiPickerDTHeader?: Classes<typeof DTHeaderStyles>;
-  MuiPickerDTTabs?: Classes<typeof DTTabsStyles>;
-  MuiPickersCalendar?: Classes<typeof CalendarStyles>;
-  MuiPickersCalendarView?: Classes<typeof CalendarViewStyles>;
-  MuiPickersCalendarHeader?: Classes<typeof CalendarHeaderStyles>;
-  MuiPickersSlideTransition?: Classes<typeof SlideTransitionStyles>;
-  MuiPickersYearSelection?: Classes<typeof MuiPickersYearSelectionStyles>;
-  MuiPickersYear?: Classes<typeof MuiPickersYearStyles>;
-  MuiPickersMonthSelection?: Classes<typeof MuiPickersMonthSelectionStyles>;
-  MuiPickersMonth?: Classes<typeof MuiPickersMonthStyles>;
-  MuiPickersTimePickerToolbar?: Classes<typeof TimePickerToolbarStyles>;
-  MuiPickersClock?: Classes<typeof ClockStyles>;
-  MuiPickersClockNumber?: Classes<typeof ClockNumberStyles>;
-  MuiPickersClockPointer?: Classes<typeof ClockPointerStyles>;
-  MuiPickersToolbar?: Classes<typeof PickerToolbarStyles>;
-  MuiPickersToolbarButton?: Classes<typeof ToolbarButtonStyles>;
-  MuiPickersToolbarText?: Classes<typeof MuiPickersToolbarTextStyles>;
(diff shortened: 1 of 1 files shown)


oliviertassinari:
@bopfer Awesome, thanks for opening a standalone pull request.

It appears that something is wrong with the CI on the pickers repository. If I apply this diff in the main repository:

```diff
diff --git a/packages/material-ui/src/styles/overrides.d.ts b/packages/material-ui/src/styles/overrides.d.ts
index fe4dab79e..707323f0e 100644
--- a/packages/material-ui/src/styles/overrides.d.ts
+++ b/packages/material-ui/src/styles/overrides.d.ts
@@ -144,7 +144,7 @@ export interface ComponentNameToClassKey {
   MuiExpansionPanelDetails: ExpansionPanelDetailsClassKey;
   MuiExpansionPanelSummary: ExpansionPanelSummaryClassKey;
   MuiFab: FabClassKey;
-  MuiFilledInput: FilledInputClassKey;
+  MuiFilledInput?: FilledInputClassKey;
   MuiFormControl: FormControlClassKey;
   MuiFormControlLabel: FormControlLabelClassKey;
   MuiFormGroup: FormGroupClassKey;
```

and run `yarn typescript`, I get the following error:

```sh
' is not assignable to type 'string'.
    Type 'undefined' is not assignable to type 'string'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

the same one that was reported on the linked issue: #1858.
This sounds like a new argument in favor of increasing the priority of https://github.com/mui-org/material-ui/issues/19706.

Do we have any idea on how we could update the CI in the pickers to catch the problem in the first place? It could be interesting to figure out in order to further align pickers with the main repo, I would fear we either introduce regression until them migration is completed and that we have other similar cases like this on in the pickers repository.

oliviertassinari (approved):

dmtrKovalenko:
I am not sure why the problem appears – our tests are telling us that everything works fine. Just checked it with up-to-date `next` https://github.com/mui-org/material-ui-pickers/blob/6cde0eaf11531d10667d9f4d78bbcb9b3d51763f/lib/src/__tests__/typescript/Overrides.tsx#L4

oliviertassinari:
Yeah, I think that there is an issue with how the TypeScript runs in the pickers repository. I came to this conclusion after looking at the behavior on the main repository and the reproduction example. 

Maybe it has something to do with the options we give to TypeScript here, and how strict we are with the soundness?

dmtrKovalenko:
our version is v3.9.3 and everything is as strict as possible. 
Needs to investigate

PR #1893, opened 2020-06-14
[docs] Prepare Material-UI X

oliviertassinari:
Preview:

Capture d’écran 2020-06-14 à 12 23 5

@dmtrKovalenko Regarding the ad, it's not right, do you want to contact CodeFund to block Kendo and Syncfusion? Alternatively, it might be simpler to switch the property to Material-UI, a bit ahead of time, we already have this blacklist in place.

### DIFF
--- a/docs/pages/demo/daterangepicker/index.mdx
+++ b/docs/pages/demo/daterangepicker/index.mdx
@@ -18,7 +18,7 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
 > ⚠️⚠️ The date range picker is unstable, it's **not suitable** for usage in production. ⚠️⚠️
 
 > The date range picker will be made available in the coming months for production use as part of a paid extension to the community edition (MIT license) of Material-UI.
-> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.). Pricing plans will start at \$129/developer for early access.
+> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.). Pricing for early access will start with an affordable plan.
 
 The [date range pickers](https://material.io/components/pickers/) let users select a range of dates. 
 
--- a/docs/pages/demo/daterangepicker/index.mdx
+++ b/docs/pages/demo/daterangepicker/index.mdx
@@ -13,9 +13,14 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
 
 <PageMeta component="DateRangePicker" />
 
-## Date picker
+## Date range picker
 
-[Date pickers](https://material.io/components/pickers/) let users select a date, or a range of dates. They should be suitable for the context in which they appear.
+> ⚠️⚠️ The date range picker is unstable, it's **not suitable** for usage in production. ⚠️⚠️
+
+> The date range picker will be made available in the coming month for production usage as part of a paid extension of the community edition (MIT license) of Material-UI.
+> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.). Pricing plans will start at \$129/developer for early access.
+
+The [date range pickers](https://material.io/components/pickers/) let users select a range of dates. They should be suitable for the context in which they appear.
 
 <Ad />
 
--- a/docs/pages/demo/daterangepicker/index.mdx
+++ b/docs/pages/demo/daterangepicker/index.mdx
@@ -17,7 +17,7 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
 
 > ⚠️⚠️ The date range picker is unstable, it's **not suitable** for usage in production. ⚠️⚠️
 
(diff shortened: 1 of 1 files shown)


docs/pages/demo/daterangepicker/index.mdx
```diff
@@ -13,9 +13,14 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
 
 <PageMeta component="DateRangePicker" />
 
-## Date picker
+## Date range picker
 
-[Date pickers](https://material.io/components/pickers/) let users select a date, or a range of dates. They should be suitable for the context in which they appear.
+> ⚠️⚠️ The date range picker is unstable, it's **not suitable** for usage in production. ⚠️⚠️
+
+> The date range picker will be made available in the coming month for production usage as part of a paid extension of the community edition (MIT license) of Material-UI.
```
mbrookes:
```suggestion
> The date range picker will be made available in the coming months for production use as part of a paid extension to the community edition (MIT license) of Material-UI.
```

docs/pages/demo/daterangepicker/index.mdx
```diff
@@ -13,9 +13,14 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
...
+> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.). Pricing plans will start at \$129/developer for early access.
```
mbrookes:
```suggestion
> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.).
```
1. Do we need pricing here?
2. Is paid support for MIT components in or out? (It isn't mentioned here.)

docs/pages/demo/daterangepicker/index.mdx
```diff
@@ -13,9 +13,14 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
...
+
+The [date range pickers](https://material.io/components/pickers/) let users select a range of dates. They should be suitable for the context in which they appear.
```
mbrookes: > They should be suitable for the context in which they appear.

For example?

mbrookes (dismissed):

docs/pages/demo/daterangepicker/index.mdx
oliviertassinari: 1. I think that it can help developers decide if they want to use the alpha anyway or move to alternatives.
2. How does the answer to this question impact the warning? It seems to be independent.

docs/pages/demo/daterangepicker/index.mdx
oliviertassinari: Are you suggesting the following?
```suggestion
The [date range pickers](https://material.io/components/pickers/) let users select a range of dates. 
```

docs/pages/demo/daterangepicker/index.mdx
mbrookes: 2. You're right, it's incomplete - it doesn't mention support at all, so MIT doesn't matter one way or the other.

docs/pages/demo/daterangepicker/index.mdx
mbrookes: I was asking what, specifically, "They should be suitable for the context in which they appear." referred to? Suitable in what sense? What "context"? But If the sentence isn't needed, then yes it can be deleted.

docs/pages/demo/daterangepicker/index.mdx
oliviertassinari: So we should rephrase the previous sentence? I think that we still need to create clear boundaries between the two worlds (community vs x)

```diff
-The date range picker will be made available in the coming month for production usage as part of a paid extension of the community edition (MIT license) of Material-UI.
+The date range picker will be made available in the coming month for production usage as part of a paid extension of the MIT licensed modules of Material-UI.
```

docs/pages/demo/daterangepicker/index.mdx
```diff
@@ -13,9 +13,14 @@ import * as CustomRangeInputs from './CustomRangeInputs.example';
...
 <PageMeta component="DateRangePicker" />
 
-## Date picker
+## Date range picker
 
-[Date pickers](https://material.io/components/pickers/) let users select a date, or a range of dates. They should be suitable for the context in which they appear.
+> ⚠️⚠️ The date range picker is unstable, it's **not suitable** for usage in production. ⚠️⚠️
+
+> The date range picker will be made available in the coming months for production use as part of a paid extension to the community edition (MIT license) of Material-UI.
+> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.). Pricing plans will start at \$129/developer for early access.
```
oliviertassinari:
```suggestion
> This paid extension will include advanced components (rich data grid, date range picker, tree view drag & drop, etc.). Pricing for early access will start with an affordable plan.
```

dmtrKovalenko (approved):

PR #1907, opened 2020-06-17
Feat: Render Month

### LINKED ISSUE #1170: A renderMonth function
Is there any possibility of an implementation of a renderMonth function? This would work very similarly to how the renderDay function works, although allowing custom rendering for different selected months.

phsantiago:
This PR closes #1170

## Description
I used a implementation very similar to renderDay. It's missing the tests and types, if someone want to help in that part, feel totally free ❤️

### DIFF
--- a/docs/pages/demo/datepicker/ServerRequest.example.jsx
+++ b/docs/pages/demo/datepicker/ServerRequest.example.jsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Badge } from '@material-ui/core';
 import { TextField } from '@material-ui/core';
-import { DatePicker, Day } from '@material-ui/pickers';
+import { DatePicker, Day, Month } from '@material-ui/pickers';
 import { makeJSDateObject } from '../../../utils/helpers';
 import { CalendarSkeleton } from '@material-ui/pickers/CalendarSkeleton';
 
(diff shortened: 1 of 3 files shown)


docs/pages/demo/datepicker/ServerRequest.example.jsx
```diff
@@ -45,6 +45,9 @@ function ServerRequest() {
         // loading
         renderInput={props => <TextField {...props} />}
         renderLoading={() => <CalendarSkeleton />}
+        renderMonth={(month, MonthComponentProps) => {
```
oliviertassinari, line 48: It doesn't seem to be related the example. Please leave it untouched.

lib/src/views/Calendar/MonthSelection.tsx
```diff
@@ -77,17 +79,16 @@ export const MonthSelection: React.FC<MonthSelectionProps> = ({
...
+        const monthProps = {
+          key: monthText,
+          value: monthNumber,
+          selected: monthNumber === currentMonth,
+          onSelect: onMonthSelect,
+          disabled: shouldDisableMonth(month),
+          children: monthText,
+        };
+
+        return renderMonth ? renderMonth(month, monthProps) : <Month {...monthProps} />;
```
oliviertassinari, line 91: Please keep the API consistant:

```suggestion
        return renderMonth ? renderMonth(monthProps) : <Month {...monthProps} />;
```

oliviertassinari (changes requested):
Thanks for looking at the issue. In order to move forward, I think that would need to:

- update the types, the prop types
- add a new test (testing library)
- figure out what API we want: https://github.com/mui-org/material-ui/issues/21453

oliviertassinari:
@phsantiago  Thanks for working on this problem. I'm closing as we have moved the component to https://github.com/mui-org/material-ui. You can find the same issue: https://github.com/mui-org/material-ui/issues/23524.

PR #1923, opened 2020-06-22
Improve date parts selection flow

### LINKED ISSUE #1651: Selecting a date should close the picker
## Environment

## Steps to reproduce

1. Open https://next.material-ui-pickers.dev/demo/datepicker#basic-usage
2. Click on the first calendar icon
3. Select a date
4. The popup is still open

## Expected behavior

The popup should close.

I think that we could consider a `disableCloseOnSelect` prop as the Autocomplete has for the current behavior.

### LINKED ISSUE #1709: DateTimePicker should call onChange when date is picked
## Environment

| @material-ui/pickers |    ^4.0.0-alpha.5     |

## Steps to reproduce

On desktop.

1. Open [DateTimePicker demo](https://next.material-ui-pickers.dev/demo/datetime-picker)
2. Open the "Clear Initial State" picker at the bottom. You can also use the default picker at the top and skip to step (6.)
3. Pick a date. Note how the text field is not filled -- `onChange()` was not called yet. This is fine, as the picker doesn't know the time yet. The picker will automatically open the hours view.
4. Pick hour and minute and close the picker. The selected date/time is now in the textbox as intended -- `onChange()` was called when you selected the minute.
5. Remember the date you picked and open the picker a second time.
6. Pick a (different) date.
7. The picker will automatically move on to the hours view.
7. Ignore that and close the picker by clicking outside, before selecting another time.

## Expected behavior

The value should now be `"<date picked last> <time picked the first time>"`.

## Actual behavior

The value is `"<date picked the first time> <time picked the first time>"`.
That is, the date picked the second time is completely ignored.

Moreover, if you open the picker again, you'll see the second date still highlighted, you can select a third date, close again, and that will be ignored too. So the "open" picker has an (inner) state that is detached from the textbox/outer value.

## Suggestion

`isFinish` should only be used to determine whether there's another view to move to and whether the picker should auto-close. It should *not* be used to determine whether to call `onChange()` (on the outer picker) or not. Instead the outer `onChange()` should be called whenever the inner state is complete, i.e. it has all the components that it needs.

As a simplification, it would also be fine to call `onChange()` even in step (3.) above, i.e. when only a date was selected. Within the picker's UI, the current time is used, so that can be sent to `onChange()` too. That should simplify things to "always call `onChange()` when the inner value is valid and the user selected any day/hour/minute". Or in other words, only refrain from syncing the inner value to the outer with `onChange()` when (a) it's invalid or (b) the user only opened the picker to look at the already-selected value without changing it.

### LINKED ISSUE #1909: Time Picker of static mode is only updated when smallest unit of time is changed
| @material-ui/pickers | v4.0.0-alpha.8 |
| material-ui          |     4.9.14    |
| TypeScript           |  3.4.5       |
| React                |    16.13.1     |
| Peer library         |         |  |

## Steps to reproduce

1. Render a static time picker as the demo: https://v4-0-0-alpha-8.material-ui-pickers.dev/demo/timepicker#static-mode
2. Change the hour in the picker
3. Click the edit icon (pen icon) in the picker 
4. The hour is not changed at all. The onChange method is not called when change the hour in the picker.

## Expected behavior
The hour should be changed

## Actual behavior
The hour should be changed after the hour is changed in the picker

## Live example
https://v4-0-0-alpha-8.material-ui-pickers.dev/demo/timepicker#static-mode

notes:
I also rendered the time picker with hour, minute and second. And the time picker will not call onChange() when hour and minute are changed in the picker. Not sure if this is by design, but I think user should be able to change any time unit and get the time updated.

dmtrKovalenko:
This PR closes #1909  
also closes #1709

## Description
This PR makes time selection flow clear and obvious because right now it is different between mobile and desktop versions. But it could be a problem (as we are changing them in runtime) and potentially lead to bugs. 

## User facing changes

* From now on we will dispatch `onChange` when each view is filled (when the next view needs to open). This allows the developer to understand which value is already filled and also improves UX with. `Cancel` button – if cancel clicked already filled part will be saved. (closes #1909)
* `autoOk` prop was removed and now picker will automatically close after selection **for Desktop** mode. It is possible to control this behavior with the new `disableCloseOnSelectProp` (closes #1651)


lib/src/views/Calendar/Year.tsx
```diff
@@ -27,6 +28,14 @@ export const useStyles = makeStyles(
       display: 'flex',
       justifyContent: 'center',
       padding: '8px 0',
+      '&:focus, &:hover': {
+        '& $yearButton': {
```
oliviertassinari: Would it be possible to reduce the specificity of this selector?
Why do we need two DOM nodes per year? Could we merge it to get only one? It would improve customizability and performance.

dmtrKovalenko: We need container to follow the `grid ` patter of year view. 
image

But agree it could be simplified

oliviertassinari:
> From now on we will dispatch onChange when each view is filled (when the next view needs to open). This allows the developer to understand which value is already filled and also improves UX with. Cancel button – if cancel clicked already filled part will be saved.

@dmtrKovalenko This sounds like it's going in the right direction. I have recently noticed the existance of a `onAccept` callback. It feels strange if not wrong. What's your thoughts on the following statement:

> The value/defaultValue/onChange props (trio of props for managing and exposing any state publickly) should represent the last value selected by the user.

If you agree, what are our options to remove the onAccept prop?

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -89,24 +88,19 @@ export function usePickerState<TInput, TDateValue>(
...
-
-          if (currentVariant !== 'mobile') {
-            acceptDate(newDate, autoAcceptRequested);
-          }
+
+        if (selectionState === 'partial') {
+          acceptDate(newDate, false);
+        }
+
+        if (selectionState == 'finish') {
```
oliviertassinari: It seems that we miss an eslint rule here:

```suggestion
        if (selectionState === 'finish') {
```

It will be solved with #1930

lib/src/views/Clock/ClockView.tsx
```diff
@@ -92,202 +93,203 @@ const getMinutesAriaText = (minute: string) => `${minute} minutes`;
...
+        getRequestedTimePoint: (when: 'start' | 'end') => MaterialUiPickersDate
+      ) => {
+        const isAfterComparingFn = createIsAfterIgnoreDatePart(
+          Boolean(disableIgnoringDatePartForTimeValidation),
+          utils
+        );
 
-          // prettier-ignore
-          return Boolean(
+        // prettier-ignore
```
oliviertassinari: Why ignore prettier?

```suggestion
```

#1936

e2e/test-utils.tsx
```diff
@@ -1,12 +1,14 @@
...
 import DateFnsAdapter from '@material-ui/pickers/adapter/date-fns';
 import TextField, { TextFieldProps } from '@material-ui/core/TextField';
 import { mount } from 'cypress-react-unit-test';
 import { DIALOG_WIDTH } from '../lib/src/constants/dimensions';
 import { LocalizationProvider, StaticDatePickerProps } from '@material-ui/pickers';
 
 // Time of the first commit to the pickers ❤️
-const momentInTime = new Date('2017-10-07T19:36:00.000Z');
+const momentInTime = parse('2017-10-07T19:36:00.000');
+parse;
```
oliviertassinari:
```suggestion

```

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -89,24 +88,19 @@ export function usePickerState<TInput, TDateValue>(
...
+          acceptDate(newDate, Boolean(autoOk));
+        }
+
+        if (selectionState == 'forceFinish') {
```
oliviertassinari:
```suggestion
        if (selectionState === 'forceFinish') {
```

dmtrKovalenko:
Related the `onAccept` prop – now it works like "when date is *finally* selected and we need to close picker". 
For example if it needs to make server request based on date change – it is probably the best place, because we are sure user is finished with datepicker

oliviertassinari:
OK thanks for the detail. I don't understand why we need it. Why is onChange not enough? From what I understand there are 2 different cases:

- the user uses the keyboard, it fires a change event like the native date time input, at each keystroke. Users can filter invalid dates out.
- the user uses the popup, it fires a change event only once the selection of the date is completed, either on click on the day button if autoOk is true, or on click on the OK button if autoOk is false.

`onChange` seems enough.

dmtrKovalenko:
`onAccept` means more like "The date was accepted and datepicker should close, so you could make a server request". Something like tat

dmtrKovalenko:
As appeared this PR is hardly linked with #1651, so decided to fix this issue in scope of this PR.

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -21,4 +22,56 @@ describe('<DatePicker />', () => {
 
     expect(screen.getByRole('button', { name: '2010' })).toBeDisabled();
   });
+
+  it("Doesn't close picker on selection in Mobile mode", async () => {
```
dmtrKovalenko:
```suggestion
  it("Doesn't close picker on selection in Mobile mode", () => {
```

docs/pages/guides/FormikValidationSchema.example.tsx
```diff
@@ -1,8 +1,8 @@
 // @ts-nocheck
 import React from 'react';
+import TextField from '@material-ui/core/TextField';
 import { date, object } from 'yup';
 import { Grid } from '@material-ui/core';
```
oliviertassinari:
```suggestion
import Grid from '@material-ui/core/Grid';
```

docs/pages/guides/StaticComponents.example.jsx
```diff
@@ -6,7 +6,7 @@ import { useStaticState, ClockView, Calendar } from '@material-ui/pickers';
 function StaticPickers() {
   const [value, handleDateChange] = useState(new Date());
 
-  // you can past mostly all available props, like minDate, maxDate, autoOk and so on
+  // you can past mostly all available props, like minDate, maxDate,   and so on
```
oliviertassinari:
```suggestion
  // you can past mostly all available props, like minDate, maxDate, autoOk and so on
```

e2e/component/DateTimePicker.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    // Date
+    cy.findByLabelText('next month')
+      .click()
+      .click();
+    cy.findByLabelText('Dec 22, 2015').click();
+    cy.get('input').should('have.value', '12/22/2015 7:36 PM');
+
+    // Hour
+    cy.findByRole('menu').trigger('mouseup', { buttons: 1, offsetX: 66, offsetY: 157 });
```
oliviertassinari, line 36: This action looks brittle, are they anyway we could improve it? For instance, maybe we could query the bounding box of the area and compute one angle?

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
```diff
@@ -86,11 +88,12 @@ export const DateTimePickerToolbar: React.FC<ToolbarComponentProps> = withDefaul
         {wrapperVariant !== 'desktop' && (
           <PickerToolbar
             toolbarTitle={toolbarTitle}
-            isLandscape={false}
             penIconClassName={classes.penIcon}
             className={classes.toolbar}
             isMobileKeyboardViewOpen={isMobileKeyboardViewOpen}
             toggleMobileKeyboardView={toggleMobileKeyboardView}
+            {...other}
+            isLandscape={false}
```
oliviertassinari, line 96: Should we allow the developers to change the orientation?

```suggestion
            isLandscape={false}
            {...other}
```

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -21,4 +22,56 @@ describe('<DatePicker />', () => {
 
     expect(screen.getByRole('button', { name: '2010' })).toBeDisabled();
   });
+
+  it("Doesn't close picker on selection in Mobile mode", () => {
```
oliviertassinari: We have been including "it" as part of the sentence

```suggestion
  it("doesn't close picker on selection in Mobile mode", () => {
```

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -21,4 +22,56 @@ describe('<DatePicker />', () => {
...
+        TransitionComponent={FakeTransitionComponent}
+        value={utilsToUse.date('2018-01-01T00:00:00.000Z')}
+        onChange={jest.fn()}
+        renderInput={props => <TextField {...props} />}
+      />
+    );
+
+    fireEvent.click(screen.getByLabelText('Choose date, selected date is Jan 1, 2018'));
+
+    await waitFor(() => screen.getByRole('dialog'));
```
oliviertassinari, line 53: I think that we should look into why it's necessary:

```suggestion
    screen.getByRole('dialog');
```

It hides something

lib/src/_shared/PickerToolbar.tsx
```diff
@@ -72,15 +72,13 @@ const PickerToolbar: React.SFC<PickerToolbarProps> = ({
...
   toolbarTitle,
-  ...other
 }) => {
   const classes = useStyles();
 
   return (
     <Toolbar
       data-mui-test="picker-toolbar"
       className={clsx(classes.toolbar, { [classes.toolbarLandscape]: isLandscape }, className)}
-      {...other}
```
oliviertassinari, line 83: Why removing the spread?

e2e/component/DateTimePicker.spec.tsx
oliviertassinari: This makes me think that until we get ChromeHeadless to run the tests into this can't be tested with JSDOM. We could work on bringing the test env to this repository, but I think that it will we be simpler to wait for the code migration.

oliviertassinari (commented):
We are almost good :)

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
dmtrKovalenko: No, DateTimePicker layout doesn't support landscape. At least for now, need to spend more time on this

lib/src/__tests__/DatePickerTestingLib.test.tsx
dmtrKovalenko: For now, all the tests are written with first capital letter, so we need to change or everything or nothing

lib/src/__tests__/DatePickerTestingLib.test.tsx
dmtrKovalenko: Yes, it is an issue with popper native `onCreate` function that works through timeout and need to update the state. (in order to allow focusing elements and prevent scroll)

lib/src/_shared/PickerToolbar.tsx
dmtrKovalenko: It is an internal component that doesn't need to support spreadoing

lib/src/__tests__/DatePickerTestingLib.test.tsx
oliviertassinari: If we assume that we have 2000 tests on material-ui, 10 tests on material-ui-x and 100 tests on material-ui-pickers, 5% of the tests are written with a capital first letter. The closest we get to 0%, the lower entropy, the better :)

lib/src/DateTimePicker/DateTimePickerToolbar.tsx
oliviertassinari: Make sense

lib/src/_shared/PickerToolbar.tsx
oliviertassinari: Fair enough

oliviertassinari (approved):

Issue #1631, opened 2020-04-09
Discrepancy with @material-ui/core


oliviertassinari:
The purpose of this issue is to **identify misalignment** between the main repository and this one, see what improvement we can bring in any direction.

This issue fits into this effort https://github.com/mui-org/material-ui/issues/19706.

The core idea around why this approach can help bring value:

1. Put constraints in place to maximize the consitency of the developer experience on the whole library (**user side**).
2. Make it easier for an engineer to contribute to the library, as a whole, **no matter which component** we are used to working on. Experimentations should go through small efforts / **RFC** and be rollout at 100% when successful, not scattered between components.
3. Share knowledge, tools, and great practices learned the hard way. Any small incremental improvement to the CI, tests, etc compound. It can be very hard to share and enforce different repositories.

### List of items

- [x] Export default right from the function declaration to increase code density. Meaning, no `export default BasicDatePicker;`.
- [x] Replace usage of `import React, { useState } from 'react';` with `import React as * from 'react'; React.useState`. https://github.com/mui-org/material-ui/pull/19802
- [x] Improve one level deep from:

```diff
-import { Button } from '@material-ui/core';
+import Button from '@material-ui/core/Button';
```

- [x] Remove `.displayName` for component. React dev tools can infer the display name from the name of the function, plus the display names are kept in production, increasing the bundle size and reducing obfuscation. For instance: https://github.com/mui-org/material-ui-pickers/blob/502118dbc268f77ca510bb78d6d588ea96ec4161/lib/src/DateRangePicker/DateRangePickerDay.tsx#L161
- [ ] Add display names when leveraging the `React.createContext()` API. The motivation is explained in https://github.com/mui-org/material-ui/pull/18468. For instance: https://github.com/mui-org/material-ui-pickers/blob/502118dbc268f77ca510bb78d6d588ea96ec4161/lib/src/LocalizationProvider.tsx#L6
- [ ] Always spread the props on the root element if public. always. When not possible, consider breaking the component down into smaller chunks or introducing a `xxxProps` prop. I have seen the following issues. For instance: 
- [x] Migrate from **enzyme** shallow (eww) to enzyme render (good) to **testing-library** (great). This is a work in progress on our side with the core components. We would upgrade the test anytime we find the opportunity, for instance when fixing a bug.
[...]

PR #1928, opened 2020-06-23
[core] Sort props ASC

oliviertassinari:
One step closer to solve #1631 and https://github.com/mui-org/material-ui/issues/19706.


lib/src/_shared/hooks/useIsLandscape.tsx
```diff
@@ -29,12 +29,15 @@ export function useIsLandscape(
     getOrientation()
   );
 
-  const eventHandler = React.useCallback(() => setOrientation(getOrientation()), []);
```
oliviertassinari, line 32: Simplification

dmtrKovalenko:
Do you make all this sortings manually? If so – we need some automate solution for that

oliviertassinari:
We have a constraint on sorting the prop types, otherwise eslint crashes. It would be cool if we can extended this static analysis.

lib/src/views/Calendar/SlideTransition.tsx
```diff
@@ -69,10 +69,10 @@ export const useStyles = makeStyles(
 
 export const SlideTransition: React.SFC<SlideTransitionProps> = ({
   children,
-  transKey,
+  className,
   reduceAnimations,
   slideDirection,
-  className = undefined,
```
oliviertassinari, line 75: Remove dead code.

dmtrKovalenko (approved):

PR #1930, opened 2020-06-23
[core] Align the eslint configuration with the main repository

oliviertassinari:
One step closer to https://github.com/mui-org/material-ui/issues/19706, the same as https://github.com/mui-org/material-ui-x/pull/23. I have copied & past the eslint configuration of the main repository.

TODO

- [ ] Fix the errors


oliviertassinari:
@dmtrKovalenko Do you think that you could take this task over me? I don't think that I will have the time to solve all the eslint errors. Making smaller chunks, like #1937 would be ideal. Thanks! :)

dmtrKovalenko:
@oliviertassinari, sure will take it. Thanks for the opening!

PR #1936, opened 2020-06-28
[core] Use prettier at 100%

oliviertassinari:
Discovered in https://github.com/mui-org/material-ui-pickers/pull/1923/files#r446568504.

### DIFF
--- a/lib/src/DateRangePicker/date-range-manager.ts
+++ b/lib/src/DateRangePicker/date-range-manager.ts
@@ -41,8 +41,5 @@ export function calculateRangePreview(options: CalculateRangeChangeOptions): Dat
   }
 
   const [previewStart, previewEnd] = newRange;
-  // prettier-ignore
-  return options.currentlySelectingRangeEnd === 'end'
-    ? [end, previewEnd]
-    : [previewStart, start];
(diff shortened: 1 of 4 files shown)


dmtrKovalenko (approved):

lib/src/DateRangePicker/date-range-manager.ts
```diff
@@ -41,8 +41,5 @@ export function calculateRangePreview(options: CalculateRangeChangeOptions): Dat
   }
 
   const [previewStart, previewEnd] = newRange;
-  // prettier-ignore
-  return options.currentlySelectingRangeEnd === 'end'
-    ? [end, previewEnd]
-    : [previewStart, start];
+  return options.currentlySelectingRangeEnd === 'end' ? [end, previewEnd] : [previewStart, start];
```
dmtrKovalenko, line 44: Don't like this formatting, it looks broken

oliviertassinari: I think that you make the point for why not using prettier ignore is valuable. It frees us from having discussions around what's the best formatting. Where do we start using it? When do we stop?

PR #1937, opened 2020-06-28
[core] Prepare migration with main repository, eslint v1

oliviertassinari:
This is a smaller chunk of #1930. I have figured that it would be much easier to have this effort go into smaller efforts.

### DIFF
--- a/docs/_shared/Example.tsx
+++ b/docs/_shared/Example.tsx
@@ -93,7 +93,7 @@ function Example({ source, testId, paddingBottom }: ExampleProps) {
   );
 
   return (
-    <>
+    <React.Fragment>
       <Collapse key="code" in={expanded}>
         <div className={classes.sourceToolbar}>
           <Tooltip title="Propose file change">
@@ -136,7 +136,7 @@ function Example({ source, testId, paddingBottom }: ExampleProps) {
 
         <ExampleComponent key={currentLib} />
       </div>
-    </>
+    </React.Fragment>
   );
 }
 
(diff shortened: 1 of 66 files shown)


dmtrKovalenko:
What the motivation of bringing this 2 parts in different PRs? I think they should never be merged independently. 

Maybe change the base branch to #1930?

docs/_shared/Example.tsx
```diff
@@ -93,7 +93,7 @@ function Example({ source, testId, paddingBottom }: ExampleProps) {
   );
 
   return (
-    <>
+    <React.Fragment>
```
dmtrKovalenko, line 96: What the motivation?

oliviertassinari:
> What the motivation of bringing this 2 parts in different PRs? I think they should never be merged independently.

I was planning on making more than 2 pull requests, more like a couple. The incentive is incremental changes: avoid conflicts on large pull requests, makes review easier. #1930 will finalize the effort, putting constraints in place so it holds in the long term. Step by step.

docs/_shared/Example.tsx
oliviertassinari: Why are we using `<React.Framgent>` in the codebase? I think that it's between these lines:

- Only this approach allows setting a `key`
- Only this approach allows new developers to easily search on Google what the element for. 
- We enforce one way for consistency. 

If you think that we should change the tradeoff, please open an RFC on the main repository to make the point. The objective here is to share the same eslint configuration on the whole codebase.

dmtrKovalenko (approved):

Issue #1795, opened 2020-05-20
[Datepicker] - add ok and cancel button


hadasmaimon:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯
I don't know why I don't find how to add the ok cancel button (It works in the past)
![image](https://user-images.githubusercontent.com/40765442/82452214-a7047000-9ab7-11ea-93d7-4224368d1a70.png)

## Expected Behavior 🤔
expected to have the ok and cancel buttons.
![image](https://user-images.githubusercontent.com/40765442/82452353-d0250080-9ab7-11ea-82a7-ef0264719749.png)

## Context 🔦

![image](https://user-images.githubusercontent.com/40765442/82452491-fea2db80-9ab7-11ea-8321-4361f2821dc3.png)

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Material-UI | v 4.0.0-alpha.7  |
| React       |         |
| Browser     |         |
| TypeScript  |         |
| etc.        |         |

dmtrKovalenko:
There is no way to display dialog buttons on the desktop with `<DatePicker />`. 
The only option is to use `MobileDatePicker`.

oliviertassinari:
A couple of buttons and there use cases that come to mind for the date picker:

- "OK" when we don't want the "autoOK" behavior
- "Today" bottom when we get lost
- "Clear" if we want to reset the value to null

That could potentially be useful. I don't know if it's something that should be built-in, or only provide a customization point for doing it.

dmtrKovalenko:
We could think about some option to render `anything` after the calendar. It could be helpful for buttons or displaying related helper information for users. Something like `renderBottomInfo`. 

For instance react-dates have the following prop 
```js
<DateRangePickerWrapper calendarInfoPosition="bottom" renderCalendarInfo={renderCalendarInfo} autoFocus />
```

http://airbnb.io/react-dates/?path=/story/drp-calendar-props--with-info-panel-bottom

oliviertassinari:
A random thought on the possible API:

```jsx
<DatePicker
  renderPopup={(props) => {
    <div>
      <StaticDatePicker {...props} />
      <Button>OK</Button>
    </div>
  }}
/>

hadasmaimon:
good thanks!

Jesusz0r:
[...]

Issue #578, opened 2018-08-16
Allow components to be styled using classnames


encodotom:
## Environment

| Tech                | Version |
|---------------------|---------|
| material-ui-pickers |  1.0.0-rc.11       |
| material-ui         |  1.3.1       |
| React               |    16.3.1     |
| Browser             |  All      |
| Platform            |     All    |

## Steps to reproduce
1. Try to override the styles of picker components (e.g. https://github.com/dmtrKovalenko/material-ui-pickers/blob/develop/lib/src/DatePicker/Calendar.jsx)
2. Find that these components can only be styled using a MaterialUI theme

## Expected behavior
As well as allowing users to style components using themes, material-ui allows users to pass in classnames which can be used to override the styles of components (see https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Button/Button.d.ts). This is _extremely_ beneficial for stacks which aren't using themes.

## Actual behavior
Picker components can only be styled using a MaterialUI theme.

dmtrKovalenko:
I am actually aware of that because this is such a giant bunch of work. 
I am not promise that it will be implemented so any help is totally appreciated

oliviertassinari:
I see that the components have a name defined: https://github.com/dmtrKovalenko/material-ui-pickers/blob/bdd7c1de801809a7798c55316a250cb9845e5fd2/lib/src/TimePicker/components/Clock.tsx#L172. You should get this feature for free when you upgrade to v4.0.0-beta.0. It was implemented in  https://github.com/mui-org/material-ui/pull/15140.

dmtrKovalenko:
@oliviertassinari it will leverage displayName?

oliviertassinari:
@dmtrKovalenko It uses the style sheet `name`. This value should already be unique in the `theme.overrides` scope.

dmtrKovalenko:
@oliviertassinari ohh, I got what you mean :) 
But I suppose this issue is mostly related to providing the prop `classnames`. But we differently need to create an example which will show how to override styles with global classnames :)

oliviertassinari:
Rereading the issue description, it seems that people are interested in having an API where they can provide a `classes` prop to each nested component. Right now, their only option is to use `theme.overrides`. Tomorrow they will be able to use the global class name scope. So, my bad. You won't get this feature for free. We still have to manually forward the props.

anderha:
Any update on this? Would be really great to have a `classes` prop

[...]

PR #1945, opened 2020-06-29
Add prop to render elements after calendar (#1795)

Jenselme:
This allows us to render anything after the calendar (buttons, text…)
by passing the renderCalendarInfo props to DateRangePicker.

Right now this PR is missing tests and documentation for the new props.

Before doing that, I'd like some inputs from you to see whether my approach seems correct to you. I'd also like to update the `DatePicker` to provide similar functionnality if it is. I tried to follow the ideas discussed in the related issue #1795 

My interogation is: I am rendering the extra information at the right place (after the `Calendar` component) or should I render it inside the `Calendar` component.

### DIFF
--- a/lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
@@ -17,7 +17,7 @@ import {
   isEndOfRange,
   DateValidationProps,
 } from '../_helpers/date-utils';
-import { DayProps as CalendarDayProps } from "../views/Calendar/Day";
+import { DayProps as CalendarDayProps } from '../views/Calendar/Day';
 
 export interface ExportedDesktopDateRangeCalendarProps<TDate> {
   /**
--- a/lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
@@ -17,6 +17,7 @@ import {
   isEndOfRange,
   DateValidationProps,
 } from '../_helpers/date-utils';
+import { DayProps as CalendarDayProps } from "../views/Calendar/Day";
 
 export interface ExportedDesktopDateRangeCalendarProps<TDate> {
   /**
@@ -30,6 +31,19 @@ export interface ExportedDesktopDateRangeCalendarProps<TDate> {
    * @example (date, DateRangeDayProps) => <DateRangePickerDay {...DateRangeDayProps} />
    */
   renderDay?: (date: TDate, DateRangeDayProps: DateRangeDayProps<TDate>) => JSX.Element;
+  /**
+   * Custom components. We currently only support Calendar to customize its rendering.
+   */
+  slotComponents?: {
+    Calendar?: React.ElementType;
+  };
+  /**
+   * Extra props passed to the custom components. You can use this to override any props
+   * on the base component too.
+   */
+  slotProps?: {
+    Calendar?: CalendarProps<TDate>;
+  };
 }
 
 interface DesktopDateRangeCalendarProps<TDate>
@@ -102,6 +116,8 @@ export function DateRangePickerViewDesktop<TDate>(props: DesktopDateRangeCalenda
     currentlySelectingRangeEnd,
     currentMonth,
     renderDay = (_, dateRangeProps) => <DateRangeDay {...dateRangeProps} />,
+    slotComponents = { Calendar },
+    slotProps = {},
     ...other
   } = props;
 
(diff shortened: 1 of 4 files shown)


lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -25,6 +25,7 @@ export interface ExportedDesktopDateRangeCalendarProps {
    * @default 2
    */
   calendars?: 1 | 2 | 3;
+  renderCalendarInfo?: React.ReactNode;
```
dmtrKovalenko: Will be better to have a function that will take the currentMonth as an argument.

dmtrKovalenko (changes requested):

oliviertassinari:
This concern is quite interesting. I think that we can use it as a case study, also with #1907 around how we can improve the customization API. I would propose we keep this pull request on hold until we resolve https://github.com/mui-org/material-ui/issues/21453.

I went a bit looking into react-day-picker and found a `components` API https://react-day-picker-next.netlify.app/docs/custom-components. It looks heavily inspired by https://react-select.com/components.

Jenselme:
> I went a bit looking into react-day-picker and found a components API https://react-day-picker-next.netlify.app/docs/custom-components. It looks heavily inspired by https://react-select.com/components.

Do you think the `components` API is more the way to go than the function?

I haven't seen https://github.com/mui-org/material-ui/issues/21453 before creating this but yes, a uniform API would be better. Let me know if I can help. Currently, I'll have to use a custom build of the project to move forward with my project, but I'd like to see this go upstream as fast as possible. Perhaps we can also move forward with the function right now and change it in a new version (if I understand correctly version 4 is still in alpha so we can do this).

oliviertassinari:
> Perhaps we can also move forward with the function right now and change it in a new version (if I understand correctly version 4 is still in alpha so we can do this).

@Jenselme Sorry, this is not possible, we would only merge changes once we are confident we have explored enough the alternative options.

dmtrKovalenko:
In my opinion render function here will be the best choice. Because it will be possible to render it dynamically for months without having to introduce new state. 

My idea: 

```
<DatePicker 
   renderCalendarInfo={(month) => <div> You will not be able to do something before {moth.sub({ month: 1 })} </div>
/>
```

Moreover we need to handle it for DateRangePicker where it is extremely useful. Something like


```
<DateRangePicker 
   renderLastCalendarInfo={(month) => <Button> I do not need return ticket </Button>} 
/>
```

It is the most required functionality I suppose

Jenselme:
@dmtrKovalenko so with two functions, one to allow rendering before the calendar and one to allow rendering after?

oliviertassinari:
What about this approach? This configuration can reduce the need for new custom APIs, we can reuse the existing component separation of the internal, progressively expose more to the public:

```tsx
import { DateRangePicker, Calendar } from '@material-ui/pickers';

function MyCalendar(props) {
  return <Calendar classes={{}} {...props} />;
}

return (
  <DateRangePicker components={{ Calendar: MyCalendar }} />
);
```

At this point, you can render and customize the component as you see fit. Maybe even leverage the `classes` API: #578.

dmtrKovalenko:
I am not sure, will it be "over" customized? 
There is the point when customization ability becomes a pain. Here as for me render function will fit the best

oliviertassinari:
@dmtrKovalenko I think that in an ideal world, we would expose an API that allows composition, the closer we are to the DOM notes, the better.

```jsx
<DatePicker>
  <DatePickerInput />
  <DatePickerPopup>
    <Calendar />
  </DatePickerPopup>
<DatePicker>
```

Now, in practice, this is likely overkill, it requires a lot of coordination.

I think that we can benefit from your existing reflection on what's the best wait to separate the different components (the current private components) and simply reexpose them. I have added a couple of more thoughts on the problem on https://github.com/mui-org/material-ui/issues/21453#issuecomment-653496935

oliviertassinari:
> I am not sure, will it be "over" customized?
There is the point when customization ability becomes a pain. Here as for me render function will fit the best

I think that over-customizability is a concern on its own and that can be resolved independently. No matter the direction we go in.

Jenselme:
@oliviertassinari @dmtrKovalenko did you decide for a path forward? I'd really like to complete this and see it merged in September ;-)

oliviertassinari:
@Jenselme Yes, we would be going with https://github.com/mui-org/material-ui/issues/21453#issuecomment-657278080. However, we might still change the names of these props. `slotComponents` could become `components` and `slotProps` -> `componentsProps`.

Jenselme:
@oliviertassinari any idea when this will be stable?

oliviertassinari:
@Jenselme What would be stable? The name? The name can be changed later on, it's a low-cost change, no need to optimize for it.

Jenselme:
> The name can be changed later on, it's a low-cost change, no need to optimize for it.

If it's just the names, I can update this PR yes.

oliviertassinari:
@Jenselme Also, please mind this comment: https://github.com/mui-org/material-ui-pickers/pull/2092#issuecomment-679342233.

Jenselme:
@oliviertassinari if I understood correctly, it should be something like this: https://github.com/mui-org/material-ui-pickers/pull/1945/commits/4b82dd04070574de3aca48cdb9e862f6ce52057d Am I right?

oliviertassinari:
Yes, it looks close :)

Jenselme:
I just updated the PR, that should be it. I'm not used to TS so I may have done mistakes with it. I know I had to add a `// @ts-ignore` in my example.

For testing, should I use "calendar customization" in `docs/pages/regression/Regression.tsx` to validate this?

I also noted that this PR won't be merged directly because of https://github.com/mui-org/material-ui/issues/21453#issuecomment-657278080 But I think the result is reviewable anyway.

oliviertassinari:
@Jenselme Thanks for working on the problem. I'm closing as we have moved the component to https://github.com/mui-org/material-ui.

PR #1949, opened 2020-07-02
[DateTimePicker] Always show arrow switcher in clock

### LINKED ISSUE #1925: Consistent display of ArrowSwitcher in clock views
## Environment

| @material-ui/pickers |    4.0.0-alpha.8     |
| material-ui          |    4.10.2     |
| React                |  16.13.1       |
| date-fns         |    2.14.0     |

## Steps to reproduce / Live example
```jsx
<DateTimePicker
        // starts with 'hours' view and displays ArrowSwitcher
        views={["hours", "minutes", "seconds"]}
        // does not display ArrowSwitcher for time views
        views={["year", "month", "date", "hours", "minutes", "seconds"]}
/>
```

https://codesandbox.io/s/datetimepicker-nq5qk?file=/src/App.js

## Expected behavior
ArrowSwitcher to be displayed for all time views.

## Actual behavior
ArrowSwitcher is sometimes displayed.

dmtrKovalenko:

This PR closes #1925


docs/pages/demo/datepicker/CustomDay.example.jsx
```diff
@@ -3,8 +3,8 @@ import React, { useState } from 'react';
 import isSameDay from 'date-fns/isSameDay';
 import endOfWeek from 'date-fns/endOfWeek';
 import startOfWeek from 'date-fns/startOfWeek';
-import isWithinInterval from 'date-fns/isWithinInterval';
 import TextField from '@material-ui/core/TextField';
+import isWithinInterval from 'date-fns/isWithinInterval';
```
oliviertassinari, line 7: A couple of thoughts:

- The diff isn't related to the problem we are trying to solve. It makes the review a bit harder.
- I think that we should welcome the initiative to leave the space in better shape when we first come in. Thanks for that.
- Why do we have rules as a warning? Outside of the transitioning period, I think that it should either be an error or off.
- What do you want to do with the `pretty-imports/sorted` rule? Do you want to keep it? If so, please make the case for it on https://github.com/mui-org/material-ui.

oliviertassinari (changes requested):
It sounds like we miss a test case for this

docs/pages/demo/datepicker/CustomDay.example.jsx
dmtrKovalenko: This rule is more like my own preference. I will not open an RFC for the usage of this rule :) 
looks like more developers tend to not sort imports at all.

The main idea of this rule is to be used with `autoFix`, honestly it should be a plugin for prettier and not an eslint rule.

oliviertassinari: There is a less strict import sorting rule right now, it's about using relative imports (`./`) after global ones. Having to run eslint autofix could be an issue. I think that we gave up on it on the main repository because it was too slow to run. I don't know if there are alternatives.

oliviertassinari (approved):

Issue #1911, opened 2020-06-19
Remove `selectedDate` from the `renderDay` 


dmtrKovalenko:
We probably don't need this argument at all, because we always has access to the currently selected using their own state. 

So proposed change for the API
```tsx
<DatePicker
  renderDay={(date, DayProps)  => <Day {...DayProps} />}
/>
```

> You are right, we changed calendar selected days to be an array when implementing date range picker. Here will be enough to only rename argument (because we have selected date available as state variable). But this makes me think that we don't need this argument at all

_Originally posted by @dmtrKovalenko in https://github.com/mui-org/material-ui-pickers/pull/1904_

oliviertassinari:
What are the pros and cons of both options? In https://github.com/mui-org/material-ui-pickers/pull/1907#discussion_r441875791, it seemed that consistency would help reduce bugs. I think that we should stick to the convention that render props and component props could be switched, with a single param, hence:

```jsx
<DatePicker renderDay={(DayProps) => <Day {...DayProps} />} />
```

oliviertassinari:
Let's stick to render "prop API <=> Component API": one argument with all the relevant options.

PR #1953, opened 2020-07-03
[DateRangePicker] Properly call `renderDayProp`

### LINKED ISSUE #1856: [DateRangePicker] renderDay not being called
## Environment

| @material-ui/pickers |  v4.0.0-alpha7  |
| material-ui          |    v4.10.0     |
| TypeScript           |   No      |
| React                |    16.13.1     |
| Browser              |    Chrome     |
| Peer library         |    moment     | 

## Steps to reproduce

1. Create Wrapping Component:
```
import React, { useState } from 'react'
import classNames from 'classnames'
import { StaticDateRangePicker as DateRangePicker, Day } from '@material-ui/pickers'
import { makeStyles } from '@material-ui/core/styles'
import TextField from '@material-ui/core/TextField'
import moment from 'moment'

const useStyles = makeStyles((theme) => ({}))

const StaticDateRangePicker = (props) => {
	const classes = useStyles()

	return (
		<DateRangePicker
			autoOk
			orientation="portrait"
			variant="static"
			openTo="date"
			disableOpenPicker
			value={props.range}
			onChange={(newRange) => {
				props.onChangedDate(newRange[0], newRange[1] ? newRange[1] : newRange[0])
			}}
			showTodayButton
			disableFuture
			views={['date']}
			renderInput={(startProps, endProps) => (
				<>
					<TextField {...startProps} />
					<TextField {...endProps} />
				</>
			)}
			renderDay={(day, selectedDates, DayComponentProps) => console.log("I'm a sailor")}
		/>
	)
}

export { StaticDateRangePicker }
```

Calling Component

```
render() {
 <StaticDateRangePicker
     range={[moment().subtract(1, "day"), moment()]}
     onChangedDate={(dateA, dateB) => {
	this.onSetDateRange(dateA, dateB)
     }} />
}
```

## Expected behavior

Expect my dream career title to be logged to the console

## Actual behavior

Appears the custom renderDay prop is being ignored, or some condition is overruling the default renderDay method

Anyone else seeing this?

dmtrKovalenko:

Closes #1856

### DIFF
--- a/lib/src/DateRangePicker/DateRangePickerViewMobile.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerViewMobile.tsx
@@ -7,6 +7,7 @@ import { MaterialUiPickersDate } from '../typings/date';
 import { Calendar, CalendarProps } from '../views/Calendar/Calendar';
 import { ExportedArrowSwitcherProps } from '../_shared/ArrowSwitcher';
 import { defaultMinDate, defaultMaxDate } from '../constants/prop-types';
+import { ExportedDesktopDateRangeCalendarProps } from './DateRangePickerViewDesktop';
 import {
   isWithinRange,
   isStartOfRange,
@@ -14,11 +15,12 @@ import {
   DateValidationProps,
 } from '../_helpers/date-utils';
 
-export interface ExportedMobileDateRangeCalendarProps {}
+export interface ExportedMobileDateRangeCalendarProps
+  extends Pick<ExportedDesktopDateRangeCalendarProps, 'renderDay'> {}
 
 interface DesktopDateRangeCalendarProps
   extends ExportedMobileDateRangeCalendarProps,
-    CalendarProps,
+    Omit<CalendarProps, 'renderDay'>,
     DateValidationProps,
     ExportedArrowSwitcherProps {
   date: DateRange;
@@ -39,6 +41,7 @@ export const DateRangePickerViewMobile: React.FC<DesktopDateRangeCalendarProps> 
   rightArrowButtonProps,
   rightArrowButtonText,
   rightArrowIcon,
+  renderDay = (_, props) => <DateRangeDay {...props} />,
   ...other
 }) => {
   const utils = useUtils();
@@ -67,17 +70,17 @@ export const DateRangePickerViewMobile: React.FC<DesktopDateRangeCalendarProps> 
         {...other}
         date={date}
         onChange={onChange}
-        renderDay={(day, _, DayProps) => (
-          <DateRangeDay
-            isPreviewing={false}
-            isStartOfPreviewing={false}
-            isEndOfPreviewing={false}
-            isHighlighting={isWithinRange(utils, day, date)}
-            isStartOfHighlighting={isStartOfRange(utils, day, date)}
-            isEndOfHighlighting={isEndOfRange(utils, day, date)}
-            {...DayProps}
-          />
-        )}
+        renderDay={(day, _, DayProps) =>
+          renderDay(day, {
+            isPreviewing: false,
+            isStartOfPreviewing: false,
+            isEndOfPreviewing: false,
+            isHighlighting: isWithinRange(utils, day, date),
+            isStartOfHighlighting: isStartOfRange(utils, day, date),
+            isEndOfHighlighting: isEndOfRange(utils, day, date),
+            ...DayProps,
+          })
+        }
       />
     </React.Fragment>
   );
--- a/lib/src/__tests__/DateRangePicker.test.tsx
+++ b/lib/src/__tests__/DateRangePicker.test.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
-import { utilsToUse } from './test-utils';
 import { DesktopDateRangePicker } from '../';
 import { screen, waitFor } from '@testing-library/react';
+import { utilsToUse, getAllByMuiTest } from './test-utils';
 import { TextField, TextFieldProps } from '@material-ui/core';
 import { createClientRender, fireEvent } from './createClientRender';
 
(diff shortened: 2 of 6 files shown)


lib/src/__tests__/DateRangePicker.test.tsx
```diff
@@ -31,4 +31,19 @@ describe('<DateRangePicker />', () => {
...
     expect(screen.getByRole('tooltip')).toBeInTheDocument();
   });
+
+  it('prop – `renderDay` should be called and render days', async () => {
+    render(
+      <DesktopDateRangePicker
+        open
+        renderInput={defaultRangeRenderInput}
+        onChange={jest.fn()}
+        renderDay={day => <div key={String(day)} data-mui-test="renderDayCalled" />}
```
oliviertassinari, line 41: We try to use the default test id attribute name when possible.

```suggestion
        renderDay={day => <div key={String(day)} data-testid="renderDayCalled" />}
```

cc @eps1lon Shouldn't we replace all usage of `data-mui-test` for `data-testid` in the core components too?

lib/src/DateRangePicker/DateRangePickerViewMobile.tsx
```diff
@@ -39,6 +41,7 @@ export const DateRangePickerViewMobile: React.FC<DesktopDateRangeCalendarProps>
   rightArrowButtonProps,
   rightArrowButtonText,
   rightArrowIcon,
+  renderDay = (_, props) => <DateRangeDay {...props} />,
```
oliviertassinari, line 44: It looks like we will need to refactor the render APIs. I think that the correct API should be:

```jsx
  renderDay = (props) => <DateRangeDay {...props} />,
```

As far as I have seen this pattern used in the community, it gets one argument, with the equivalent of the props.

I will open a RFC for that breaking change

oliviertassinari (commented):
Thanks for the test case :)

lib/src/__tests__/DateRangePicker.test.tsx
eps1lon: I had that in mind yes. I'd prefer to keep them out of the core components though. This way people aren't confused why they aren't available in prod builds.

oliviertassinari: I'm not sure we will be able to remove all the data test attributes from the source of the components (only using them in the test). Cases where it seems to be our only option:

- Icons: they are aria-hidden.
- Modal backdrop: no labels, no roles.

lib/src/DateRangePicker/DateRangePickerViewMobile.tsx
dmtrKovalenko: I strongly believe that we need exactly 2 arguments here. And that's why: 

```jsx
renderDay={(day, selectedDate, DayComponentProps) => {
        const date = makeJSDateObject(day); // skip this step, it is required to support date libs
        const isSelected =
          DayComponentProps.inCurrentMonth && highlightedDays.includes(date.getDate());

        return (
          <Badge
            key={date.toString()}
            overlap="circle"
            badgeContent={isSelected ? '🌚' : undefined}
          >
            <Day {...DayComponentProps} />
          </Badge>
        );
      }}
```

Actually an example of rendering custom day will require access to the day that is rendering. I cannot imagine an example where its not. (maybe only if user need to pass custom style). 

If we will pass the `date` as a prop it will need to spend more time to understand how to access the current date. And also will require some kind of notation

```jsx
(DayProps) => <Day {...DayProps} className={getClassByDay(DayProps.day} /> 
```

```jsx
({ day, ...other }) => <Day day={day} {...other} className={getClassByDay(day} /> 
```

I think this notation is the best

```jsx
(day, DayProps) => <Day {...DayProps} className={getClassByDay(day} /> 
```

dmtrKovalenko: > P.S. `selectedDays` must be removed #1911

oliviertassinari: I don't understand why `day` and `selectedDate` should come from here. It feels that:

- `day` should already be available as part of `DayComponentProps`: https://github.com/mui-org/material-ui-pickers/blob/8f895d5e438dcc8f7f04837763979a3009d777dc/lib/src/views/Calendar/Day.tsx#L80.
- `selectedDate` should already be available in the global scope: https://github.com/mui-org/material-ui-pickers/blob/8f895d5e438dcc8f7f04837763979a3009d777dc/docs/pages/demo/datepicker/ServerRequest.example.jsx#L11

dmtrKovalenko: I have described in my previous answer. Because accessing `renderDay` without using `day` is mostly useless. 
So if we know that `day` is always required we can simplify accessing it from `DayProps.day` to the first argument.  ([react-day-picker's `renderDay`](http://react-day-picker.js.org/api/DayPicker/#renderDay) as reference. )

oliviertassinari: I have looked at the codebase of the Autocomplete, we do have cases like this where we decide to inject extra arguments: https://github.com/mui-org/material-ui/issues/19254#issuecomment-643494821. I wonder if it's the best approach. It feels like we could have a single argument, that it would be simpler.

I really like the simplicity of this variant:

```jsx
(PickersDayProps) => 
  <Day
    {...PickersDayProps}
    className={clsx(PickersDayProps.className, getClassByDay(DayProps.day})
  /> 
```

We would need to update Autocompelte too but, It can feel simpler, to have a single place to look for available options. I also wonder if this is not a case simple enough for accepting a Component 🤷‍♂️. I think that we should really come up with a simple rule to decide with this tradeoff.

dmtrKovalenko: It makes users go to the docs/console.log the props to understand what is passing to the `props`. When there is only one argument – it is easier.

oliviertassinari: > It makes users go to the docs/console.log the props to understand what is passing to the props

Why? We can document the props of the Day component, and link this same page.

dmtrKovalenko: Yes, we can, and it is documented. But why developer should look and google for props documentation if he need to just render the same day as we do but do some custom things based on the `day of month`? 

I mean it is not so clear as just simple passing of both `day` and `props`.

oliviertassinari (approved):

lib/src/DateRangePicker/DateRangePickerViewMobile.tsx
oliviertassinari: The idea would be that we link the PickersDay component page, right in the DateRangePicker prop page. Making the information accessible with one click. We could also have a demo with the pattern.

I think that one advantage is about not having to wonder about when and which arguments should be exposed. A second advantage is that it's closer to how components work. 


If we don't go down this route. I would recommend that we force the render props to never have more than 4 arguments, never.

Issue #1895, opened 2020-06-15
Wrong value of isValid() when using DE


TomekStaszkiewicz:
## Environment

| Tech                 | Version |
| -------------------- | ------- |
| @material-ui/pickers |    4.0.0-alpha.8     |
| material-ui          |   4.10.2      |
| TypeScript           |         |
| React                |      16.12.0   |
| Browser              |         |
| Peer library         |         |  |

## Steps to reproduce
Open code sandbox link: https://codesandbox.io/s/billowing-snowflake-gfgts?file=/package.json
1. Use English, then edit it using input(NOT THE POPOVER) 
![Screenshot from 2020-06-15 11-46-03](https://user-images.githubusercontent.com/46328998/84643338-fb0b4480-aefd-11ea-80c9-280e57b50fb3.png)
2. When it's an invalid date, the textfield is shown as red
![Screenshot from 2020-06-15 11-46-14](https://user-images.githubusercontent.com/46328998/84643392-0d857e00-aefe-11ea-908a-c5d15a77d5c0.png)
if you change it back to correct date it turns green.
![Screenshot from 2020-06-15 11-46-21](https://user-images.githubusercontent.com/46328998/84643427-18d8a980-aefe-11ea-94f7-58ded4ad63c8.png)

3. Change to de and do the same things

![Screenshot from 2020-06-15 11-46-29](https://user-images.githubusercontent.com/46328998/84643455-22faa800-aefe-11ea-9aa0-badf97fecec6.png)

4. When it's an invalid date, it's red (as expected).
![Screenshot from 2020-06-15 11-46-36](https://user-images.githubusercontent.com/46328998/84643466-27bf5c00-aefe-11ea-8dd2-5e1ac5841c73.png)
But when it's valid again, it remains red.
If you check the console, you will see that the isValid() returns false for the correct date.

![Screenshot from 2020-06-15 11-46-45](https://user-images.githubusercontent.com/46328998/84643471-2a21b600-aefe-11ea-87fd-1c1bfb6ef3b9.png)

## Expected behavior

isValid() returning true for correct date
## Actual behavior

What is important is that this bug occurs only when changing the date by typing in the date. It does not occur when you change the date by clicking on the popover.

oliviertassinari:
@TomekStaszkiewicz Thanks for the detailed reproduction. Did you have a chance to look at why it behaves wrongly like this?

oliviertassinari:
I have simplified the reproduction down to:

```jsx
[...]

Issue #1934, opened 2020-06-24
Support for nested imports?


aamirafridi:
Using the Import like this for treeshaking

```
import {
  DesktopDateRangePicker,
  MobileDateRangePicker,
} from "@material-ui/pickers/DateRangePicker/DateRangePicker";
```

No TS error but JS errored `Cannot find module '@material-ui/pickers/DateRangePicker/DateRangePicker'`

aamirafridi:
I know that this works
```
import {
  DesktopDateRangePicker,
  MobileDateRangePicker,
} from "@material-ui/pickers";
```

oliviertassinari:
@aamirafridi We don't support nested imports for the pickers repository. We might with https://github.com/mui-org/material-ui/issues/19706. I think that the correct solution would be:

```jsx
import DateRangePicker, {
  DesktopDateRangePicker,
  MobileDateRangePicker,
} from "@material-ui/pickers/DateRangePicker";
```

The incentive so far for supporting it is that tree-shaking doesn't work in development, which can harm the DX.

aamirafridi:
Ok thanks :)

oliviertassinari:
Keeping it open to signal that it's not fixed and incentive for more comments.

oliviertassinari:
Done in https://next.material-ui.com/components/pickers/

PR #1955, opened 2020-07-04
[DatePicker] Add test for textbox aria-invalid

PaulSavignano:
- Tests aria-invalid attribute on the DatePicker's input element when the locale is updated and the date becomes invalid through incorrect text input.

- Add test test for #1895 (https://github.com/dmtrKovalenko/date-io/pull/406)

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -60,7 +60,7 @@
     "test": "jest",
     "test:typescript": "yarn rimraf src/__tests__/dist && tsc -p src/__tests__/tsconfig.json",
     "test:date-fns": "UTILS=date-fns yarn test",
-    "test:dayjs": "UTILS=datejs yarn test",
+    "test:dayjs": "UTILS=dayjs yarn test",
     "test:luxon": "UTILS=luxon yarn test",
     "test:moment": "UTILS=moment yarn test",
     "start": "rollup --config rollup.config.dev.js --watch & npx tsc --watch",
--- a/lib/src/__tests__/DatePickerTestingLib.test.tsx
+++ b/lib/src/__tests__/DatePickerTestingLib.test.tsx
@@ -4,7 +4,7 @@ import enLocale from 'date-fns/locale/en-US';
 import 'dayjs/locale/de';
 import TextField from '@material-ui/core/TextField';
 import { screen, waitFor } from '@testing-library/react';
-import { DateAdapter, getByMuiTest, utilsToUse, FakeTransitionComponent } from './test-utils';
+import { UtilClassToUse, getByMuiTest, utilsToUse, FakeTransitionComponent } from './test-utils';
 import { createClientRender, fireEvent } from './createClientRender';
 import {
   DatePicker,
@@ -130,10 +130,10 @@ describe('<DatePicker />', () => {
       const [value, setValue] = React.useState<unknown>(new Date('01/01/2020'));
 
       return (
-        <LocalizationProvider dateAdapter={DateAdapter} locale={locale}>
+        <LocalizationProvider dateAdapter={UtilClassToUse} locale={locale}>
           <Picker
             onChange={setValue}
-            renderInput={(props) => <TextField {...props} />}
+            renderInput={(props2) => <TextField {...props2} />}
             value={value}
             {...PickerProps}
           />
--- a/lib/src/__tests__/DatePickerTestingLib.test.tsx
+++ b/lib/src/__tests__/DatePickerTestingLib.test.tsx
@@ -1,9 +1,18 @@
 import * as React from 'react';
+import deLocale from 'date-fns/locale/de';
+import enLocale from 'date-fns/locale/en-US';
+import 'dayjs/locale/de';
 import TextField from '@material-ui/core/TextField';
 import { screen, waitFor } from '@testing-library/react';
-import { getByMuiTest, utilsToUse, FakeTransitionComponent } from './test-utils';
+import { DateAdapter, getByMuiTest, utilsToUse, FakeTransitionComponent } from './test-utils';
 import { createClientRender, fireEvent } from './createClientRender';
-import { DatePicker, MobileDatePicker, DesktopDatePicker } from '../index';
+import {
+  DatePicker,
+  DatePickerProps,
+  DesktopDatePicker,
+  LocalizationProvider,
+  MobileDatePicker,
+} from '../index';
 
 describe('<DatePicker />', () => {
   const render = createClientRender({ strict: false });
@@ -95,5 +104,78 @@ describe('<DatePicker />', () => {
 
     fireEvent.click(screen.getByLabelText('Jan 1, 2018'));
     expect(onChangeMock).not.toHaveBeenCalled();
+  });
+
+  describe('input validation', () => {
+    const locales = {
+      en: {
+        valid: 'January 2020',
+        invalid: 'Januar 2020',
+        dateFns: enLocale,
+      },
+      de: {
+        valid: 'Januar 2020',
+        invalid: 'Janua 2020',
+        dateFns: deLocale,
+      },
+    };
+
+    interface FormProps {
+      locale: any;
+      Picker: React.ElementType<DatePickerProps>;
+      PickerProps: Partial<DatePickerProps>;
+    }
+    const Form = (props: FormProps) => {
+      const { locale, Picker, PickerProps } = props;
+      const [value, setValue] = React.useState<unknown>(new Date('01/01/2020'));
+
+      return (
+        <LocalizationProvider dateAdapter={DateAdapter} locale={locale}>
+          <Picker
+            onChange={setValue}
+            renderInput={(props) => <TextField {...props} />}
+            value={value}
+            {...PickerProps}
+          />
+        </LocalizationProvider>
+      );
+    };
+
+    Object.keys(locales).forEach((locale2) => {
+      const { valid, invalid } = locales[locale2];
+      const locale = process.env.UTILS === 'date-fns' ? locales[locale2].dateFns : locale2;
+
+      it(`${locale2}: should set invalid`, () => {
+        render(
+          <Form
+            locale={locale}
+            Picker={DesktopDatePicker}
+            PickerProps={{ views: ['month', 'year'] }}
+          />
+        );
+        const input = screen.getByRole('textbox');
+        fireEvent.change(input, { target: { value: invalid } });
+        expect(input).toBeInvalid();
+      });
+
+      // Need to run with ICU loaded https://moment.github.io/luxon/docs/manual/install.html#node
+      if (process.env.UTILS === 'luxon') {
+        return;
+      }
+
+      it(`${locale2}: should set to valid when was invalid`, () => {
+        render(
+          <Form
+            locale={locale}
+            Picker={DesktopDatePicker}
+            PickerProps={{ views: ['month', 'year'] }}
+          />
+        );
+        const input = screen.getByRole('textbox');
+        fireEvent.change(input, { target: { value: invalid } });
+        fireEvent.change(input, { target: { value: valid } });
+        expect(input).toBeValid();
+      });
+    });
   });
 });
--- a/lib/src/__tests__/DatePickerTestingLib.test.tsx
+++ b/lib/src/__tests__/DatePickerTestingLib.test.tsx
@@ -107,19 +107,6 @@ describe('<DatePicker />', () => {
   });
 
   describe('input validation', () => {
-    const locales = {
-      en: {
-        valid: 'January 2020',
-        invalid: 'Januar 2020',
-        dateFns: enLocale,
-      },
-      de: {
-        valid: 'Januar 2020',
-        invalid: 'Janua 2020',
-        dateFns: deLocale,
-      },
-    };
-
     interface FormProps {
       locale: any;
       Picker: React.ElementType<DatePickerProps>;
@@ -141,11 +128,26 @@ describe('<DatePicker />', () => {
       );
     };
 
-    Object.keys(locales).forEach((locale2) => {
-      const { valid, invalid } = locales[locale2];
-      const locale = process.env.UTILS === 'date-fns' ? locales[locale2].dateFns : locale2;
+    const tests = [
+      {
+        locale: 'en',
+        valid: 'January 2020',
+        invalid: 'Januar 2020',
+        dateFns: enLocale,
+      },
+      {
+        locale: 'de',
+        valid: 'Januar 2020',
+        invalid: 'Janua 2020',
+        dateFns: deLocale,
+      },
+    ];
 
-      it(`${locale2}: should set invalid`, () => {
+    tests.forEach((test) => {
+      const { valid, invalid } = test;
+      const locale = process.env.UTILS === 'date-fns' ? test.dateFns : test.locale;
+
+      it(`${test.locale}: should set invalid`, () => {
         render(
           <Form
             locale={locale}
@@ -163,7 +165,7 @@ describe('<DatePicker />', () => {
         return;
       }
 
-      it(`${locale2}: should set to valid when was invalid`, () => {
+      it(`${test.locale}: should set to valid when was invalid`, () => {
         render(
           <Form
             locale={locale}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,6 +157,12 @@ workflows:
             - 'Build and analyze bundlesize'
 
       - jest_tests:
+          name: 'Dayjs jest tests'
+          lib: dayjs
+          requires:
+            - 'Build and analyze bundlesize'
+
+      - jest_tests:
           name: 'Moment jest tests'
           lib: moment
           requires:
--- a/lib/src/__tests__/DateTimePickerTestingLib.test.tsx
+++ b/lib/src/__tests__/DateTimePickerTestingLib.test.tsx
@@ -8,6 +8,7 @@ import { DesktopDateTimePicker, StaticDateTimePicker } from '../DateTimePicker';
 describe('<DateTimePicker />', () => {
   // Doesn't work
   if (process.env.UTILS === 'dayjs') {
+    it('noop', () => {});
     return;
   }
 
--- a/lib/src/__tests__/DateTimePickerTestingLib.test.tsx
+++ b/lib/src/__tests__/DateTimePickerTestingLib.test.tsx
@@ -6,6 +6,11 @@ import { createClientRender } from './createClientRender';
 import { DesktopDateTimePicker, StaticDateTimePicker } from '../DateTimePicker';
 
 describe('<DateTimePicker />', () => {
+  // Doesn't work
+  if (process.env.UTILS === 'dayjs') {
+    return;
+  }
+
   const render = createClientRender({ strict: false });
 
   it('prop: mask – should take the mask prop into account', () => {
--- a/lib/src/__tests__/KeyboardDatePicker.test.tsx
+++ b/lib/src/__tests__/KeyboardDatePicker.test.tsx
@@ -5,6 +5,11 @@ import { mount } from './test-utils';
 import { DesktopDatePicker, DatePickerProps } from '../DatePicker/DatePicker';
 
 describe('e2e -- DatePicker keyboard input', () => {
+  // Doesn't work
+  if (process.env.UTILS === 'dayjs') {
+    return;
+  }
+
   const onChangeMock = jest.fn();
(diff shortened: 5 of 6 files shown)


lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+    fireEvent.change(input, { target: { value: 'Januar 2020' } });
+
+    const ariaInvalid = input.getAttribute('aria-invalid');
+    expect(ariaInvalid).toEqual('true');
+  });
+
+  it('Should add aria-invalid of false to input when date is valid', () => {
+    render(<Form picker={DesktopDatePicker} />);
+
+    const input = screen.getByLabelText('test input') as HTMLInputElement;
```
dmtrKovalenko: I think this can be simplified by `screen.getByRole('textbox')`

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+    render(<Form picker={DesktopDatePicker} />);
+
+    const localeButton = screen.getByText('Fr');
+    const input = screen.getByLabelText('test input') as HTMLInputElement;
+
+    fireEvent.click(localeButton);
+    fireEvent.change(input, { target: { value: 'janvie 2020' } });
+
+    const ariaInvalid = input.getAttribute('aria-invalid');
+    expect(ariaInvalid).toEqual('true');
```
dmtrKovalenko: This check probably could be simplified by using `expect(input).toBeInvalid()`

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+  return (
+    <LocalizationProvider dateAdapter={DateFnsUtils} locale={localeMap[locale]}>
+      <Picker
+        value={date}
+        onChange={setDate}
+        views={['year', 'month']}
+        renderInput={props => {
+          const inputProps = {
+            ...props.inputProps,
+            'aria-label': 'test input',
```
dmtrKovalenko: Probably we can avoid it in favour of role.

```suggestion
```

dmtrKovalenko (dismissed):
Thanks, this test case is definitely nice to have. Added a couple of comments how to make tests more readable.

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+      <div>
+        <Button onClick={() => setLocale('en')}>En</Button>
+        <Button onClick={() => setLocale('fr')}>Fr</Button>
+      </div>
+    </LocalizationProvider>
+  );
+};
+
+describe('<DatePicker />', () => {
+  it('Should add aria-invalid of true to input when date is invalid', () => {
```
oliviertassinari: To match the convention of the codebase. 

```suggestion
  it('should add aria-invalid of true to input when date is invalid', () => {
```

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
-import * as React from 'react';
-import { DatePicker } from '../DatePicker';
+import React, { useState } from 'react';
```
oliviertassinari: Per https://github.com/mui-org/material-ui-pickers/issues/1631

```suggestion
import * as React from 'react';
```

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+import frLocale from 'date-fns/locale/fr';
+import DateFnsUtils from '@date-io/date-fns';
+import enLocale from 'date-fns/locale/en-US';
+import Button from '@material-ui/core/Button';
 import TextField from '@material-ui/core/TextField';
-import { screen } from '@testing-library/react';
-import { createClientRender } from './createClientRender';
+import LocalizationProvider from '../LocalizationProvider';
```
oliviertassinari: In the main repository, our imports are inconsistent between using private import path and public ones.
What if every time we have the choice, we go with a public import cc @mui-org/core-team. So in our case

```suggestion
import LocalizationProvider from '@material-ui/pickers/LocalizationProvider';
```

This will encourage us to focus more on end to end tests (to cover developers & end-users).

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+import { MaterialUiPickersDate } from '../typings';
```
oliviertassinari: What's the public API variation of this type?

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,24 +1,94 @@
...
+        <Button onClick={() => setLocale('en')}>En</Button>
+        <Button onClick={() => setLocale('fr')}>Fr</Button>
+      </div>
+    </LocalizationProvider>
+  );
+};
+
+describe('<DatePicker />', () => {
+  it('Should add aria-invalid of true to input when date is invalid', () => {
+    render(<Form picker={DesktopDatePicker} />);
```
oliviertassinari: We try to keep the tests as single focused and isolated as possible. In our case, do we need the locale switcher for this test? I don't think so, I think that we should go one level of abstraction down. 

Also, what about running the tests with each date engine. Is this something we should aim for, or is it overkill at this level? Maybe the former?

oliviertassinari (commented):
Great start :)

lib/src/__tests__/DatePickerTestingLib.test.tsx
PaulSavignano: #1631 shows;
`Replace usage of import React, { useState } from 'react'; with import React from 'react'; React.useState.`

Is the wildcard * the preferred syntax?

oliviertassinari: Thanks for raising, this standard originates from https://github.com/mui-org/material-ui/pull/19802. I have updated the issue.

lib/src/__tests__/DatePickerTestingLib.test.tsx
eps1lon: Agree. I try to keep this in mind since it'll help run the test suite for production builds as well among other things. So far it wasn't technically required but for readability it would help a lot.

lib/src/__tests__/DatePickerTestingLib.test.tsx
dmtrKovalenko:
```jsx
import { MaterialUIPickersDate } from '@material-ui/pickers'
```

oliviertassinari: What about we reexport?

```jsx
import { DateType } from '@date-io/type';
```

```suggestion
import { DateType } from '@material-ui/pickers'
```

and then:

```jsx
const [date, setDate] = useState<DateType | null>(new Date('01/01/2020'));
```

oliviertassinari: Now, because the topic is broader than this pull request, I have opened an RFC, so we can take the whole codebase into consideration: #1962. But at least here, we have a great example here.

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,84 +1,116 @@
...
-    const input = screen.getByRole('textbox');
-
-    fireEvent.click(localeButton);
-    fireEvent.change(input, { target: { value: 'janvie 2020' } });
-    fireEvent.change(input, { target: { value: 'janvier 2020' } });
-
-    expect(input).toBeValid();
+  describe('locale validity', () => {
+    // iterate the different location providers and test each with each locale
+    Object.keys(dateAdapters).forEach(dateAdapterKey => {
```
oliviertassinari: Interesting approach! I wonder what's the best approach for this. For instance, the CI is configured to run with each engine:

Capture d’écran 2020-07-06 à 17 45 12

https://app.circleci.com/pipelines/github/mui-org/material-ui-pickers/1616/workflows/c55a745b-11df-4e4f-9324-9a37c8231a9e

lib/src/__tests__/DatePickerTestingLib.test.tsx
PaulSavignano: Agreed regarding each date engine.  Hijri is not validating correctly with the views={['month', 'year']}.  Maybe that calendar requires a different format. Jalali validates correctly, however, when the input is changed to invalid text and then changed back to valid text the field remains invalid.

Thoughts on Hijri and Jalali?

lib/src/__tests__/DatePickerTestingLib.test.tsx
PaulSavignano: Is the ci capturing Hijri and Jalali?

docs/pages/guides/upgrading-to-v3.mdx
```diff
@@ -79,7 +79,7 @@ That's why we recreated it from scratch.
 Now `onChange` prop for `<Keyboard(Date|Time)Picker />` has different signature from the pure pickers:
 
 ```ts
-type KeybordOnChange = (date: MaterialUiPickersDate | null, value: string | undefined) => void;
+type KeybordOnChange = (date: DateType | null, value: string | undefined) => void;
```
oliviertassinari: I'm not sure we should change the type in this pull request. I know @dmtrKovalenko Is working on the problem. What about we focus on the test for now? :)

```suggestion
type KeybordOnChange = (date: MaterialUiPickersDate | null, value: string | undefined) => void;
```

Smaller efforts are more likely to be merged.

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,81 +1,116 @@
...
+import enLocale from 'date-fns/locale/en-US';
 import TextField from '@material-ui/core/TextField';
-import { getByMuiTest } from './test-utils';
-import { screen, waitFor } from '@testing-library/react';
-import { utilsToUse, FakeTransitionComponent } from './test-utils';
-import { createClientRender, fireEvent } from './createClientRender';
-import { DatePicker, MobileDatePicker, DesktopDatePicker } from '@material-ui/pickers';
+import LocalizationProvider, {
+  LocalizationProviderProps,
+} from '@material-ui/pickers/src/LocalizationProvider';
```
oliviertassinari: From what I understand, we don't support nested imports yet as we do with the other packages: #1934.

```suggestion
import { LocalizationProvider, LocalizationProviderProps } from '@material-ui/pickers';
```

docs/pages/guides/upgrading-to-v3.mdx
PaulSavignano: Sounds good!

lib/src/__tests__/DatePickerTestingLib.test.tsx
oliviertassinari: I don't know, but maybe it's too ambitious? I think that we can aim for smaller changes. What about we focus on adding a test case for the related issue #1895? Also, it seems that we could already use `utilsToUse` for running the tests with different libraries.

PaulSavignano: Sounds good.  I'll make the update.

dmtrKovalenko:
Am I the only who cannot see the changes of this PR?
image

PaulSavignano:
@dmtrKovalenko, During a revert to an earlier commit my local git became out of sync and would not pull in changes from the next branch.  I was unable to merge and ran a reset.  

While running tests with luxon and moment the test for valid after invalid fails with locale de.  I can look into this further if you'd like.

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,15 +1,43 @@
...
+import deLocale from 'date-fns/locale/de';
+import enLocale from 'date-fns/locale/en-US';
 import TextField from '@material-ui/core/TextField';
 import { getByMuiTest } from './test-utils';
-import { screen, waitFor } from '@testing-library/react';
-import { utilsToUse, FakeTransitionComponent } from './test-utils';
+import { MaterialUiPickersDate } from '../typings';
 import { createClientRender, fireEvent } from './createClientRender';
-import { DatePicker, MobileDatePicker, DesktopDatePicker } from '@material-ui/pickers';
+import { render as rtlRender, screen, waitFor } from '@testing-library/react';
```
oliviertassinari: I don't think that using the render method exposed the testing library is an option. What about the strict mode, cleanup, extra props logic?

```suggestion
import { screen, waitFor } from '@testing-library/react';
```

Instead, we can nest two Localization Provider

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,15 +1,43 @@
...
+  LocalizationProvider,
+  MobileDatePicker,
+} from '@material-ui/pickers';
+
+interface FormProps {
+  locale: any;
+  picker: React.FC<DatePickerProps>;
+  pickerProps: Partial<DatePickerProps>;
+}
+const Form: React.FC<FormProps> = ({ locale = 'en', picker: Picker, pickerProps }) => {
```
oliviertassinari: What about we move this component as close as possible to the next tests? We try to keep each test as isolated as possible. We could have it close to the `const locales = {` variable declaration.

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -1,15 +1,43 @@
 import * as React from 'react';
+import deLocale from 'date-fns/locale/de';
+import enLocale from 'date-fns/locale/en-US';
 import TextField from '@material-ui/core/TextField';
 import { getByMuiTest } from './test-utils';
-import { screen, waitFor } from '@testing-library/react';
-import { utilsToUse, FakeTransitionComponent } from './test-utils';
+import { MaterialUiPickersDate } from '../typings';
```
oliviertassinari: Please rebase on the `next` branch to be able to leverage #1966.

lib/src/__tests__/test-utils.tsx
```diff
@@ -66,8 +66,8 @@ const getUtilClass = () => {
   }
 };
 
-export const UtilClassToUse: any = getUtilClass();
-export const utilsToUse: IUtils<MaterialUiPickersDate> = new UtilClassToUse();
+export const DateAdapter: any = getUtilClass();
```
oliviertassinari: Cool, one step in the right direction for #1969.

oliviertassinari (commented):
I have left a few comments, we are almost there :).

lib/src/__tests__/test-utils.tsx
PaulSavignano: I see the MaterialUiPickersDate type was removed in the next branch.  Thoughts on typing the useState in the Form wrapper?

oliviertassinari: See #1969

lib/src/__tests__/DatePickerTestingLib.test.tsx
PaulSavignano: I believe cleanup happens automatically; https://testing-library.com/docs/react-testing-library/api#cleanup

oliviertassinari: Interesting. @eps1lon do we need to handle the cleanup logic in `createClientRender`?

eps1lon: If you're using `jest`, yes. Though I might have an idea why this is problematic. I suggest manually cleaning up (and importing from /pure) so that it's easier to debug cleanup logic until we come up with a robust solution to clean up scheduled tasks.

oliviertassinari: Cool, thanks for the insight from a testing-library maintainer 😁!

oliviertassinari (commented):
Regarding the fails in the CI, a couple of thoughts:

- The x `console.warn` make it hard to figure out what's the assertion that fails. I think that it's important we enforce #1798.
- There are `console.error` that are raised that doesn't fail the tests :o.
- Moment and Luxon are failing on the DE locale test. Which make sense as the locale isn't configured to be loaded for them:

Capture d’écran 2020-07-14 à 12 48 05

- It might be better to run all the date adapters in the same CI run, so we can look at the failing cases without having to open x windows

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -79,6 +91,87 @@ describe('<DatePicker />', () => {
...
+          onChange={setDate}
+          renderInput={props => <TextField {...props} />}
+          value={date}
+          {...pickerProps}
+        />
+      </LocalizationProvider>
+    );
+  };
+
+  const dateAdapters = {
```
PaulSavignano, line 114: @oliviertassinari Is this approach in alignment with how you'd like to see the adapters tested in one ci?

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -79,6 +91,87 @@ describe('<DatePicker />', () => {
...
+    de: {
+      valid: 'Januar 2020',
+      invalid: 'Janua 2020',
+    },
+  };
+
+  const dateFnsLocales = {
+    en: enLocale,
+    de: deLocale,
+  };
```
PaulSavignano: @oliviertassinari Is this approach in alignment with how you'd like to see the adapters tested in one ci?

oliviertassinari: I guess we need some way to configure the locale for each date library

PaulSavignano: I think only date-fns requires a unique locale.  The others accept a string locale.

PaulSavignano: Checking back here regarding moving this along.  What update would you like to see updated here?  Should we keep with the object of adapters and loop or update the ci?

oliviertassinari: I think that the practical solution is to write the tests once but have the style engine configurable with an env variable, as Dmitriy designed it in the first place, the only improvement opportunity I see is to have all the engines runs in the same build so we can with a single glance run them all in the local env or see all the fails in the CI without opening x windows. 

So regarding this pull request, I think that we should simply configure the locale for each date engine, as you have started with date-fns.

oliviertassinari: Also, note that we need to run dayjs in the CI, we didn't made this change yet.

PaulSavignano: Sounds good.  I'll add dayjs to the ci.

lib/package.json
```diff
@@ -60,7 +60,7 @@
     "test": "jest",
     "test:typescript": "yarn rimraf src/__tests__/dist && tsc -p src/__tests__/tsconfig.json",
     "test:date-fns": "UTILS=date-fns yarn test",
-    "test:dayjs": "UTILS=datejs yarn test",
+    "test:dayjs": "UTILS=dayjs yarn test",
```
oliviertassinari, line 63: @dmtrKovalenko Dayjs tests were running with date-fns

oliviertassinari (approved):

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -96,4 +105,77 @@ describe('<DatePicker />', () => {
     fireEvent.click(screen.getByLabelText('Jan 1, 2018'));
     expect(onChangeMock).not.toHaveBeenCalled();
   });
+
+  describe('input validation', () => {
+    const locales = {
```
dmtrKovalenko: I'd propose to use `jest.each` for such cases. It outputs to the same tests but requires less code and significantly more readable

dmtrKovalenko (commented):
Thanks for this awesome tests!

lib/src/__tests__/DatePickerTestingLib.test.tsx
oliviertassinari: Regarding the usage of `jest.each`, we will move the tests to mocha, so I think that it will be simpler to rewrite the current ones to a similar pattern without be tightened to any custom Jest API. I have seen https://github.com/ryym/mocha-each still.

Issue #1759, opened 2020-05-08
Add ability to select same date in date range picker


kirill-konshin:
It possible to select same date to make one day range. It currently gives validation error: https://dev.material-ui-pickers.dev/demo/daterangepicker

image
![image](https://user-images.githubusercontent.com/1530410/81451622-c3102500-9139-11ea-9a41-5ea0e288441c.png)

Follow up for https://github.com/mui-org/material-ui-pickers/issues/364#issuecomment-626012154

oliviertassinari:
For benchmark https://www.google.com/flights and https://www.airbnb.com/ allows selecting the same date while https://www.booking.com/ doesn't.

btroop:
Would be great to add a prop to the api to allow or not allow the same date validation

dmtrKovalenko:
I am wondering about the better API for that. Maybe something like `minRangeLength` and `maxRangeLength`? That by default will be `1` and `null`.

oliviertassinari:
@dmtrKovalenko A side but related question. Should we move the validation outside of the DateRangePicker, as we did for the DatePicker?

In this approach, developers would have the following workaround: they could flag same-day selection as valid.

dmtrKovalenko:
It already has the same validation as `DatePicker` with `onError`. But `invalidRange` reason returns when the start and end dates are the same

oliviertassinari:
> It already has the same validation as DatePicker with onError

@dmtrKovalenko So people can already add a conditional to their onError handler?
if `reason ===  'invalidRange' && startDate === endDate` then ignore

dmtrKovalenko:
Yes, but it will look like

```jsx
const [error, setError] = React.useState([null, null])

<DateRange
  onError={([startReason, endReason], [start, end]) => { 
     if (startReason === 'invalidRange' && isSameDay(start, end)) {
       setError([null, null])
       return;
     }

     setError([startReason, endReason])
  })}
  renderInput={props => <TextField {...props} error={Boolean(error[0])} />   }
/>
```

oliviertassinari:
@dmtrKovalenko This looks perfect for the "lazy" solution, to let end-users fall into this trap and get visual/text feedback 👌. However, there is another side to the problem.

I assume the concern now is about: Should the UI prevent this case in the first place? It would give faster feedback in the UI, this sounds fair.

[...]

PR #1967, opened 2020-07-09
[DatePicker] Don't run onChange if same date selected

### LINKED ISSUE #1652: onChange should not trigger if selecting the same date
## Environment

## Actual behavior

https://codesandbox.io/s/holy-breeze-c6ucy?file=/src/App.js

## Expected behavior

https://codesandbox.io/s/material-demo-3sl56?file=/demo.js

We have recently fixed this aspect with the Autocomplete and the Select components, I figure it would be great to have it here too.

todor-a:
Fixes #1652 

For some reason the function `areValuesEqual` in `valueManager` was returning false when checking if two dates are equal (when in fact they were). This only happened the first time you choose the current value. If the picker's value was, let's say, 08/07/2020 and you selected it again, it would return false but if you select it a second time, it would return true. 

To solve it, I added another function - `areDaysEqual`.


oliviertassinari (dismissed):
Could you add a failing test case with testing-library that demonstrate the fix? Thanks

Also, I believe we should test all the components, date picker, date time picker, time picker, etc.

todor-a:
I'll get to it. Sorry, it's my first opensource contribution :)

lib/src/_shared/hooks/usePickerState.ts
```diff
@@ -103,8 +108,9 @@ export function usePickerState<TInput, TDateValue>(
         wrapperVariant: WrapperVariant,
         selectionState: PickerSelectionState = 'partial'
       ) => {
-        setPickerDate(newDate);
-
+        if (!valueManager.areDaysEqual(utils, newDate, pickerDate)) {
+          setPickerDate(newDate);
+        }
```
dmtrKovalenko:
```suggestion
        if (valueManager.areDaysEqual(utils, newDate, pickerDate)) {
          return;
        }

        setPickerDate(newDate);
```

dmtrKovalenko (changes requested):
I think overall logic is not correct, please add a test (with testing library) that confirms desired behavior

lib/src/_shared/hooks/usePickerState.ts
todor-a: I might be missing something but wouldn't this run only if you select the same date?

dmtrKovalenko: Yes, you are right but we still doesn't need to update the state. 
Another question – what for TimePicker and DateTimePicker? I think overall we need to add this check to the `Calendar` component.

oliviertassinari:
@tdranv Did you have the time to look at the feedback from the review? :)

todor-a:
I did but I'm having trouble with what you meant by "Could you add a failing test case with testing-library that demonstrate the fix?".

Can you elaborate a little more on that?

dmtrKovalenko:
@tdranv firstly please make this check on the `<Day />` side. We already know which day is currently selected and which will be selected. I think the perfect place for this check will be this handler: 

https://github.com/mui-org/material-ui-pickers/blob/c3c423179ceae40da6f3b5212e2618af44c2b161/lib/src/views/Calendar/Day.tsx#L181-L189

So proposed diff will be this: 

```diff
--- a/lib/src/views/Calendar/Day.tsx
+++ b/lib/src/views/Calendar/Day.tsx
@@ -179,7 +179,7 @@ const PureDay: React.FC<DayProps> = ({
   };
 
   const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
-    if (!disabled) {
+    if (!disabled && !selected) {
       onDaySelect(day, 'finish');
     }
```

And also please add a new test case to this file: 
https://github.com/mui-org/material-ui-pickers/blob/e6bfd12b759445cb8341d7669b2290412da93153/lib/src/__tests__/DatePickerTestingLib.test.tsx#L8

oliviertassinari:
> Can you elaborate a little more on that?

@tdranv Sure, the idea is that for a test to be relevant, it has to fail without the fix, otherwise we are testing something that already works.

todor-a:
I seem to have broken something else. I will take a look at it later. Thanks for being patient with me. :)

dmtrKovalenko:
From what I am seeing it looks like this behavior is not suitable for DateRangePicker. We must have the ability to select the same day for start and end. (#1759)

What I propose: add new prop `allowSameDateSelection` for the `<Day />` component. This will allow to configure this behavior for the users thanks to `renderDay` and also we will pass `allowSameDateSelection={true}` in the `DateRangeDay`. 

> UPD: thanks so much for your work ❤️

oliviertassinari:
@dmtrKovalenko Regarding #1759, what about we solve it with https://github.com/mui-org/material-ui-pickers/issues/1759#issuecomment-628894726?

For the date range picker, I think that the equivalent fix would be that if the same range is selected, no change event is fired.

todor-a:
I added the flag you suggested @dmtrKovalenko.

First, once again, thank you for being patient with me. I like Material UI a lot and wish to be part of it :) Here are the issues are encountered, as a first-time contributor:
- I find it hard to configure the development environment. Every time I run the tests I get all these errors, saying that `Your test suite must contain at least one test.` or `Test suite failed to run`. I assume that this is due to something I misconfigured but it would be useful to go into a little more detail on how to set up your environment in the How to Contribute file.
- I also couldn't run the linter/prettier. 
- I need some more time to get acquainted with how the project is structured (which was surprisingly easy for me!)

Anyway, working on this (despite it being such a small issue) is fun! Thanks for your guidance :)

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -78,4 +78,24 @@ describe('<DatePicker />', () => {
 
     expect(screen.queryByRole('dialog')).toBeInTheDocument();
   });
+
+  it('Does not call onChange if same date selected', async () => {
```
oliviertassinari:
```suggestion
  it('does not call onChange if same date selected', async () => {
```

lib/src/views/Calendar/Day.tsx
```diff
@@ -132,6 +137,7 @@ const PureDay: React.FC<DayProps> = ({
   className,
   day,
   disabled,
+  allowSameDateSelection = false,
```
oliviertassinari: sort props ASC

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -78,4 +78,24 @@ describe('<DatePicker />', () => {
...
+    const onChangeMock = jest.fn();
+
+    render(
+      <DesktopDatePicker
+        TransitionComponent={FakeTransitionComponent}
+        disableCloseOnSelect
```
oliviertassinari: Is this prop required for the test to fail without the patch?

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -78,4 +78,24 @@ describe('<DatePicker />', () => {
...
+        value={utilsToUse.date('2018-01-01T00:00:00.000Z')}
+        onChange={onChangeMock}
+        renderInput={props => <TextField {...props} />}
+      />
+    );
+
+    fireEvent.click(screen.getByLabelText('Choose date, selected date is Jan 1, 2018'));
```
oliviertassinari, line 94: I side note, I wonder if the asserting here isn't too brittle. What happens if we change the formation or the value? Would it be possible to query an element with role button instead?

lib/src/__tests__/DatePickerTestingLib.test.tsx
```diff
@@ -78,4 +78,24 @@ describe('<DatePicker />', () => {
 
     expect(screen.queryByRole('dialog')).toBeInTheDocument();
   });
+
+  it('Does not call onChange if same date selected', async () => {
+    const onChangeMock = jest.fn();
```
oliviertassinari, line 83: A side note on the naming convention. We have used this variation a lot in the main repository, would it make sense to continue?
```suggestion
    const handleChange = jest.fn();
```

oliviertassinari (commented):
Looks solid

lib/src/views/Calendar/Day.tsx
```diff
@@ -123,6 +123,11 @@ export interface DayProps extends ExtendMui<ButtonBaseProps> {
    * @default false
    */
   disableHighlightToday?: boolean;
+  /**
+   * Allow selecting the same date (fire onChange even if same date is selected).
+   * @default false
```
oliviertassinari, line 128: @eps1lon This change makes me wonder. On the main repository, we don't mention the default value in the TypeScript definitions, we have a tool that extracts them from the source.

Here, the upside is that we get the information right in the editor:

Capture d’écran 2020-07-14 à 20 28 03

But it can get quickly out-of-sync. What do you think about the tradeoff?

lib/src/__tests__/DatePickerTestingLib.test.tsx
todor-a: It is not.

lib/src/__tests__/DatePickerTestingLib.test.tsx
todor-a: Agree :)

lib/src/__tests__/DatePickerTestingLib.test.tsx
todor-a: I too think that this approach to querying can be brittle. But also fixing it should probably part of a separate issue/pull request?

lib/src/views/Calendar/Day.tsx
eps1lon: Great idea. I think we could statically verify if the default value matches. Though I wouldn't worry about it too much at first. In the end default values will rarely change in SemVer patches or minors because they're very likely breaking changes. 

I don't remember how we extract default values but if we extract them statically then we can also match them against TS types.

docs/pages/guides/css-overrides.mdx
```diff
@@ -43,6 +43,8 @@ This will also autocomplete classnames.
 
 Add the following to a TypeScript declaration file in your project, such as `overrides-mui.d.ts`:
 
+Add the following to a TypeScript declaration file in your project, such as `overrides-mui.d.ts`:
+
```
dmtrKovalenko:
```suggestion
```

dmtrKovalenko (commented):
@tdrandv Thanks, everything looks nice except [this commit](https://github.com/mui-org/material-ui-pickers/pull/1967/commits/0a60e638194d8d4ce15256a6a0b1669ec8dcd375). Can you please revert it and we could merge it 💪

docs/pages/guides/css-overrides.mdx
todor-a: Ops. Should be good now :)

dmtrKovalenko (approved):

oliviertassinari:
@tdranv For the tests, do you think that it would make sense to cover the other components, TimePicker, DateRangePicker, DateTimePicker?

lib/src/__tests__/DatePickerTestingLib.test.tsx
oliviertassinari: Yeah, right, a different pull request would make sense

todor-a:
Hm, I believe no for TimePicker. As far as DateRangePicker, a user can select the same date if the start and end dates are the same, as @dmtrKovalenko pointed out in a previous comment.

oliviertassinari:
What should happen if a end-user uses the date range picker and choose the same range than previously? Should it fire a change event or not?

todor-a:
> As far as DateRangePicker...

I probably did not say that correctly. What I mean was that a end-user can pick the same date for the start and end date. As far as your question: I believe it should not fire the event.

oliviertassinari:
@tdranv Ok cool. It was a great first pull request on Material-UI 👌🏻. Thank you for working on it!

oliviertassinari:
@tdranv If you would like to keep helping without diving too deep in the complexity of the components, there is ongoing work around unifying the codebase.

For instance, https://github.com/mui-org/material-ui/pull/21758 was recently merged on the main repository, it would be perfect if we could enforce the exact same eslint configuration here. I have started a couple of iterations already (doing it in small chunks), I'm doing that for an hour or two during the weekend. But if you want to speed things up, feel free too. Related to #1930.

todor-a:
Sure. I will have a look at it over the weekend. :)

PR #1977, opened 2020-07-11
[core] Update format to match main v2

oliviertassinari:
Build on top of #1930, one step further

PR #1976, opened 2020-07-11
[core] Add --report-unused-disable-directives to lint commands

oliviertassinari:
One chunk of #1930

PR #1979, opened 2020-07-11
test ci

oliviertassinari:
I want to test that the CI runs corretly


oliviertassinari:
Assumption confirmed, something is broken in the CI with Cypress cc @dmtrKovalenko

First noticed in #1975, #1976, #1977

PR #1987, opened 2020-07-12
test: add dayjs

PaulSavignano:
Add dayjs to ci tests

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -60,6 +60,7 @@
     "test": "jest",
     "test:typescript": "rimraf src/__tests__/dist && tsc -p src/__tests__/tsconfig.json",
     "test:date-fns": "UTILS=date-fns yarn test",
+    "test:dayjs": "UTILS=datejs yarn test",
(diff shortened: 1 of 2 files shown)


dmtrKovalenko (approved):

oliviertassinari (commented):
Is this enough? What about the configuration of the CI in https://github.com/mui-org/material-ui-pickers/blob/0818b5eeffd9b1a8d80ac7ab6afa14151a92efeb/.circleci/config.yml#L150?

dmtrKovalenko:
Ohh, I think it was only about an ability to run dayjs tests. 
On CI definitely we need to add one new workflow step

oliviertassinari:
Ok, so the question is, should we run dayjs in the CI too?

oliviertassinari:
> should we run dayjs in the CI too?

I assume that the answer is yes.

PaulSavignano:
Sounds good.  I'll add it to the ci.

PR #1994, opened 2020-07-15
Fix not rerendering days when inCurrentMonth changes

### LINKED ISSUE #1756: Wrong highlighting of days out of the month
versions:
    "@material-ui/pickers": "^4.0.0-alpha.6",
    "@material-ui/core": "^4.9.11",
 
expected: June 31 is out of the month and 1,2,3,4,5,6 is current month
actual:
image

### LINKED ISSUE #1993: MobileDateTimePicker with reducedAnimations is missing days
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-pickers/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

`MobileDateTimePicker` with `reducedAnimations` is missing dates that should be visible. Showing a `MobileDateTimePicker` for `new Date(2019, 7, 2)` will show the 2nd of August. Navigating back to July will show the last week missing entirely.

## Expected Behavior 🤔

The last week of July should be visible.

## Steps to Reproduce 🕹

Example: https://codesandbox.io/s/ecstatic-liskov-xcn9c

Steps:

1. Use `MobileDateTimePicker` with `reducedAnimations`.
2. Add `value` using `new Date(2019, 7, 2)`.
3. Navigate back to the previous month, July. Notice missing days.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| @material-ui/core | 4.10.2  |
| @material-ui/pickers | 4.0.0-alpha.9  |
| React       |  16.8.23       |

dmtrKovalenko:
Closes #1993 and also closes #1756 

The root of this issue was missing check for `inCurrentMonth` in `React.memo`. The same day with the same `key` was rendered in the new month, so react basically didn't change it only moved the node from one place to another one. 

Didn't reproduce without `reduceAnimation` because transition group is mounting and unmounting the whole component on animation.

### DIFF
--- a/lib/src/views/Calendar/Day.tsx
+++ b/lib/src/views/Calendar/Day.tsx
@@ -251,6 +251,7 @@ export const areDayPropsEqual = (prevProps: DayProps, nextProps: DayProps) => {
     prevProps.showDaysOutsideCurrentMonth === nextProps.showDaysOutsideCurrentMonth &&
     prevProps.disableHighlightToday === nextProps.disableHighlightToday &&
     prevProps.className === nextProps.className &&
+    prevProps.inCurrentMonth === nextProps.inCurrentMonth &&
     prevProps.onDayFocus === nextProps.onDayFocus &&
     prevProps.onDaySelect === nextProps.onDaySelect
   );
--- a/lib/src/views/Calendar/SlideTransition.tsx
+++ b/lib/src/views/Calendar/SlideTransition.tsx
@@ -77,7 +77,7 @@ export const SlideTransition: React.SFC<SlideTransitionProps> = ({
 }) => {
   const classes = useStyles();
   if (reduceAnimations) {
-    return <div className={className}>{children}</div>;
+    return <div className={clsx(classes.root, className)}>{children}</div>;
   }
 
   const transitionClasses = {


lib/src/views/Calendar/SlideTransition.tsx
```diff
@@ -77,7 +77,7 @@ export const SlideTransition: React.SFC<SlideTransitionProps> = ({
 }) => {
   const classes = useStyles();
   if (reduceAnimations) {
-    return <div className={className}>{children}</div>;
+    return <div className={clsx(classes.root, className)}>{children}</div>;
```
oliviertassinari, line 80: Should we spread too?

```suggestion
    return <div className={clsx(classes.root, className)} {...other}>{children}</div>;
```

oliviertassinari (changes requested):
Shouldn't we add a test case?

lib/src/views/Calendar/Day.tsx
```diff
@@ -251,6 +251,7 @@ export const areDayPropsEqual = (prevProps: DayProps, nextProps: DayProps) => {
     prevProps.showDaysOutsideCurrentMonth === nextProps.showDaysOutsideCurrentMonth &&
     prevProps.disableHighlightToday === nextProps.disableHighlightToday &&
     prevProps.className === nextProps.className &&
+    prevProps.inCurrentMonth === nextProps.inCurrentMonth &&
```
oliviertassinari, line 254: > Didn't reproduce without reduceAnimation because transition group is mounting and unmounting the whole component on animation.

What do you think of dropping the `areDayPropsEqual` logic? I believe that if we drop the ButtonBase component as proposed #1674, we would make this optimization with almost no ROI.

dmtrKovalenko:
I think this kind of bugs don’t need a pinning test, because you will delete this test once will see it in the future

lib/src/views/Calendar/Day.tsx
dmtrKovalenko: We could test it, but without this function everything is SO slow especially for dateRange picker.

lib/src/views/Calendar/SlideTransition.tsx
dmtrKovalenko: Here props are spreading to the TransitionGroup. So no, we don’t need to spread them

oliviertassinari: Right, because SlideTransition is private so it doesn't matter

lib/src/views/Calendar/Day.tsx
oliviertassinari: I'm confused. Here are my thoughts:

- What's the difference with the date range picker that explains the slowness compared to the date picker?
- We display two calendars instead of one.
- Ok, so the slowness is only an issue with desktop as we display a single calendar on mobile.
- On desktop we don't use `reduceAnimation`.
- You said that the logic is only working if `reduceAnimation`

I come to the conclusion that it shouldn't have any impact on date range picker. Where is the error?

dmtrKovalenko: No it also works for DateRangePicker preview and in month selection and keyboard control.

oliviertassinari (approved):
I have added a note about a good opportunity to simplify the codebase with removing `areDayPropsEqual` in #1674

> I think this kind of bugs don’t need a pinning test, because you will delete this test once will see it in the future

I agree because the probability that we will be able to drop this logic in the near future is probably 80% :)

lib/src/views/Calendar/Day.tsx
oliviertassinari: > keyboard control

The focus keybroad logic could be bypassing React entirely

dmtrKovalenko:
I am 99% sure we will not drop this logic. Hovering through JavaScript (Date range preview) is too slow if all 36 components rerenders.

oliviertassinari:
Good point, I didn't account for the date range selection, for which, we might not want to bypass React for rendering it.

In this case, we miss a test case at 99% ;)

oliviertassinari:
Especially considering that it fixes two GitHub issues.

Issue #1670, opened 2020-04-17
Improve treeshaking


dmtrKovalenko:
Make possible to tree-shake `Desktop|Mobile` wrappers. For instance, in the below report, DesktopDatePicker and MobileDatePicker have the same bundle size output: 25.9 kB, excluding the peer dependencies like @material-ui/core.

![image](https://user-images.githubusercontent.com/16926049/79562246-6928bc00-80b3-11ea-8639-500454109c78.png)

*https://bundlephobia.com/result?p=@material-ui/pickers@4.0.0-alpha.5*

dmtrKovalenko:
(Break exported modules by files) so we need to have 
`MobileDatePicker.tsx`
`DesktopDatePicker.tsx`
`DatePicker.tsx` 

In order to make possible to remove unused code of `MobileWrapper` for `DesktopDatePicker`

kwinto:
What tool makes such pretty reports?

dmtrKovalenko:
https://bundlephobia.com/result?p=@material-ui/pickers@4.0.0-alpha.5

oliviertassinari:
💯 for such effort.

Do we have a way to track/tests the changes we perform in at the pull request level so we are not required to publish on npm to have feedback (feedback every release vs within 10 minutes)? I see the [size snapshot](https://github.com/mui-org/material-ui-pickers/blob/next/lib/.size-snapshot.json) but it's at the package level.

Would it make sense to solve this problem after we migrate the picker to the mono-repository so we can benefit from the [module size tracker](https://material-ui.com/size-snapshot)?

I'm worried about some of the changes we do/plan to do if we have no visibility on the bundle size impact at the module level.

dmtrKovalenko:
Yeah, let’s going to wait for merge. For now I’ll test it with pure linking to CRA app and check bundlesize

aamirafridi:
This is my import
```javascript
import {
  DesktopDateRangePicker,
  LocalizationProvider,
  LocalizationProviderProps,
  MobileDateRangePicker,
  RangeInput,
} from "@material-ui/pickers";
```

I can see e.g. ClockView gets bundled as well. Any way to avoid this?

![image](https://user-images.githubusercontent.com/55896/87780558-df5dcb80-c826-11ea-9ee7-f14b8cc6c22a.png)

PR #2004, opened 2020-07-18
[core] Continue eslint sync with main repo

oliviertassinari:
Follow-up on #1930 and https://github.com/mui-org/material-ui/pull/21758.

I turned a couple of rules as "warn" as I didn't want to handle them now. We can take care of them progressively, one that is concerning and important to fix is the import/no-cycle one.

Same as https://github.com/mui-org/material-ui-x/pull/72. Once merged, we will need to work on solving all the warnings, this is a pre-requisite to be able to move the component in the main repository.


package.json
```diff
@@ -38,40 +38,32 @@
...
-    "*.{js,jsx,ts,tsx,json,css,md,mdx}": [
-      "prettier --write"
-    ]
-  },
   "eslintIgnore": [
     "**/node_modules",
     "**/coverage",
     "videos/",
+    "/docs/",
+    "/e2e/",
```
oliviertassinari: Legacy, no need to handle them

lib/src/DateRangePicker/DateRangePickerDay.tsx
```diff
@@ -96,19 +96,20 @@ const useStyles = makeStyles(
...
+    inCurrentMonth,
+    isEndOfHighlighting,
+    isEndOfPreviewing,
+    isHighlighting,
+    isPreviewing,
+    isStartOfHighlighting,
+    isStartOfPreviewing,
+    selected,
+    ...other
+  } = props;
```
dmtrKovalenko, line 112: What the original motiviation for this kind of change? Is this doing automatically?

lib/src/LocalizationProvider.tsx
```diff
@@ -35,7 +31,7 @@ LocalizationProvider.propTypes = {
   /**
    * Your component tree.
    */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
 } as any;
 
-export default withDefaultProps({ name: 'MuiPickersLocalizationProvider' }, LocalizationProvider);
+export default LocalizationProvider;
```
dmtrKovalenko, line 37: Accidental changes?

dmtrKovalenko (dismissed):
There is one rule that I am strongly want to discard and **never** use nor in the picker nor in the core code – default exports. We should not use them when we exactly know what we are importing – in all cases except examples. 

In short: when we exporting `CalendarView` we do want everybody to use them like **CalendarView** and not like `Calendar` or something else. This potential rename through the default exports is always implicit – why not avoid it? Moreover code with only named exports looks much more consistent because all the import lines have the same structure (starting with `import {`) 

https://medium.com/@rajeshnaroth/avoid-es6-default-exports-a24142978a7a
https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

lib/src/index.ts
```diff
@@ -14,15 +14,15 @@ export * from './DateRangePicker/DateRangePicker';
...
+export { default as PickersCalendarView } from './views/Calendar/CalendarView';
 
-export { Day as PickersDay } from './views/Calendar/Day';
+export { default as PickersDay } from './views/Calendar/Day';
 
-export { ClockView as PickersClockView } from './views/Clock/ClockView';
+export { default as PickersClockView } from './views/Clock/ClockView';
 
-export { Clock as PickersView } from './views/Clock/Clock';
+export { default as PickersClock } from './views/Clock/Clock';
```
dmtrKovalenko: This type of change looks weird. Probably current variant is not ideal as weel – we need to just reexport the right component names

lib/src/CalendarSkeleton.tsx
```diff
@@ -35,19 +35,20 @@ const monthMap = [
...
+    const { className, ...other } = props;
     const classes = useStyles();
     const calendarClasses = useCalendarStyles();
 
     return (
       <div className={clsx(classes.root, className)} {...other}>
         {monthMap.map((week, index) => (
           <div key={index} className={calendarClasses.week}>
-            {week.map((day, index) => (
+            {week.map((day, index2) => (
```
dmtrKovalenko, line 49: Why index2?

lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
 import { DateRange } from './RangeTypes';
 import { useUtils } from '../_shared/hooks/useUtils';
-import { makeStyles } from '@material-ui/core/styles';
 import { calculateRangePreview } from './date-range-manager';
-import { Calendar, CalendarProps } from '../views/Calendar/Calendar';
+import Calendar, { CalendarProps } from '../views/Calendar/Calendar';
```
dmtrKovalenko: That's what I am talking about
```js
import { Calendar, CalendarProps } from '...'
```

looks much better

lib/src/CalendarSkeleton.tsx
oliviertassinari: https://eslint.org/docs/rules/no-shadow

lib/src/DateRangePicker/DateRangePickerDay.tsx
oliviertassinari: The first-order motivation is to match the convention of the main repository. The second-order motivation was to allow quick scanning of the signature of the function. If you see "props" => it means it's a React component.

lib/src/LocalizationProvider.tsx
oliviertassinari: What problem does `withDefaultProps` solve? I have removed it as `LocalizationProvider` is already a global provider, it duplicates.

lib/src/index.ts
oliviertassinari: I think that the correct solution is to match https://github.com/mui-org/material-ui/blob/833861f8f47b77cbb033562f60354bc36d246f37/packages/material-ui/src/index.js#L15-L16:

```jsx
export { default as PickersClock } from './PickersClock';
export * from './PickersClock';
```

But that's for later.

oliviertassinari (commented):
Interesting topic on the named vs default imports, thanks for bringing it up. A couple of thoughts:

- I don't understand the reasoning of your argumentation. Would you like to make the case for it with an RFC on the main repository? Detailing the current problems and how this potential new standard could fix them?
- Both articles start by mentioning that both are valid, that it's a coding convention before anything else

> - This is opinionated nit picky territory.
> - Personal preferences

- Looking at the thread on Twitter: https://twitter.com/slicknet/status/1084101377297506304. It seems that default vs named export isn't the core of the solution of the problems, but the solution is much more about having a convention and following it systematically.
- Looking at the problems raised by the articles, it seems it can be solved with the convention we applying in the main repository for the React components: one component = one file, name of the component = name of the file. As for non-React modules, that's a different story.

dmtrKovalenko:
> I don't understand the reasoning of your argumentation 

I have provided 3 arguments:
1.  Implicit renames that could be buggy and provide unobvious 
2. The clearness in the module dependencies 
3. Easier refactoring 
4. More solid imports sections 
5. Ability to avoid unnecessarily "verbal" convention 

Ok, I`ll open RFC for the core repository. On Monday 😎

lib/src/DateRangePicker/DateRangePickerViewDesktop.tsx
oliviertassinari: For context, the reason I went for these changes is because eslint was complaining about duplicate exports with the same name (the default export filename was matching one of the named exports in the file), I went on removing the confusion with what was most popular in the codebase: default export.

oliviertassinari:
> Ok, I`ll open RFC for the core repository. On Monday 😎

Perfect

oliviertassinari:
On a side note, there is an important eslint rule that I couldn't fix and I had to put as a warning that will require to rethink the file structure: [import/no-cycle](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md). @dtassone Thanks for insisting on this rule. It turns out +23s on the main repository build was worth it for here :). 

I think that it will be a perfect opportunity to adopt the main repository file structure (folders, index.js, default export, etc). It might also something we want to batch with the tree-shaking reports: #1670.

oliviertassinari:
@dmtrKovalenko Can I merge before we get a conflict?

dmtrKovalenko:
Yes you can :)

lib/src/DateRangePicker/DateRangePickerDay.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
```
aamirafridi, line 1: is this file available from anywhere to import?

PR #2020, opened 2020-07-22
autoOk was removed

dandv:
Not sure when autoOk was removed, but I've just updated to the latest alpha10 from alpha7, and TypeScript flagged the property as an error.

If it wasn't removed, please disregard the PR. I no longer see `autoOk` at https://dev.material-ui-pickers.dev/api/DateTimePicker though.


dmtrKovalenko (approved):

linonetwo:
For people like me searched down here, how to replace autoOK usage?

linonetwo:
Wow this lib is part of material-ui, I will try out the new one!

PR #2016, opened 2020-07-21
[DateTimePicker] Allow same date selection

dmtrKovalenko:
Closes #2012

PR #2023, opened 2020-07-22
Check prettier on ci

dmtrKovalenko:

### DIFF
--- a/lib/src/DateRangePicker/DateRangePickerDay.tsx
+++ b/lib/src/DateRangePicker/DateRangePickerDay.tsx
@@ -136,10 +136,10 @@ export const PureDateRangeDay = ({
         })}
       >
         <Day
+          {...other}
           disableMargin
           allowSameDateSelection
           allowKeyboardControl={false}
-          {...other}
           day={day}
           selected={selected}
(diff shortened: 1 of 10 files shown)


lib/src/DateRangePicker/DateRangePickerDay.tsx
```diff
@@ -136,12 +136,12 @@ export const PureDateRangeDay = ({
         })}
       >
         <Day
-          allowKeyboardControl={false}
-          disableMargin
           {...other}
+          disableMargin
+          allowSameDateSelection
```
oliviertassinari, line 141: Unrelated to prettier

oliviertassinari (changes requested):
What about we:

1. Merge #2016
2. Rebase #2023 on top of next
3. Get a red CI, force merge
4. Push a prettier fix on next

This way, we get a clean workflow that confirms it works. Also, feel free to copy & paste the config from the main repo, it has been stress tested, it works, no need to innovate: https://github.com/mui-org/material-ui-x/pull/40.

dmtrKovalenko:
Lets do this

Issue #2027, opened 2020-07-23
DateRangePicker -> Custom input component demo slightly broken


Chee7ah:
- [X] The issue is present in the latest release.
- [X] I have searched the [issues](https://github.com/mui-org/material-ui-pickers/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

DateRangePicker closes immediately after the first date has been picked.
Happens in [alpha.9](https://v4-0-0-alpha-9.material-ui-pickers.dev/demo/daterangepicker#custom-input-component) and [alpha.10](https://v4-0-0-alpha-10.material-ui-pickers.dev/demo/daterangepicker#custom-input-component).

![G51d4zo4g4](https://user-images.githubusercontent.com/16325672/88260792-f0b93480-cc79-11ea-8657-bdd170e0353e.gif)

## Expected Behavior 🤔

DateRangePicker stays open till both ends of the range have been picked.
Works as expected in [alpha.8](https://v4-0-0-alpha-8.material-ui-pickers.dev/demo/daterangepicker#custom-input-component).

![uEhDmjuF5S](https://user-images.githubusercontent.com/16325672/88260671-b2237a00-cc79-11ea-94d9-21059d56c91d.gif)

## Steps to Reproduce 🕹

Steps:

1. Go to [https://v4-0-0-alpha-10.material-ui-pickers.dev/demo/daterangepicker#custom-input-component](https://v4-0-0-alpha-10.material-ui-pickers.dev/demo/daterangepicker#custom-input-component)
2. Fill in a date range with the picker
3. Note that picker closes immediately after first date is picked
4. Have to blur and refocus second input field to be able to fill the second date

## Context 🔦

Live example on the demo page

## Your Environment 🌎

The demo page

dmtrKovalenko:
Thanks for raising. There issue is related to usage of `startProps.ref` instead of `startProps.inputRef`. 

Can you check that it works for you? And if you have time -- PR with a fix of example will be appreciated ❤️

Chee7ah:
Checked, it fixes it. Opened PR #2028 with the change.

oliviertassinari:
Thanks for working on it, I think that we can close the issue once we have taken care of https://github.com/mui-org/material-ui-pickers/pull/2028#issuecomment-663123827.

PR #2028, opened 2020-07-23
[docs] Fix `inputRef` date picker customization demo

Chee7ah:
Fixes Issue #2027.

Applying the `inputRef` instead of `ref` to the `<input />` component in the demo fixes the issue of the picker closing prematurely after a single date pick instead of waiting for both dates of the range to be picked.

- [X] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui-org/material-ui/blob/master/CONTRIBUTING.md#sending-a-pull-request).

### DIFF
--- a/docs/pages/demo/daterangepicker/CustomRangeInputs.example.tsx
+++ b/docs/pages/demo/daterangepicker/CustomRangeInputs.example.tsx
@@ -11,7 +11,10 @@ function CustomRangeInputs() {
       onChange={(date) => handleDateChange(date)}
       renderInput={(startProps, endProps) => (
         <React.Fragment>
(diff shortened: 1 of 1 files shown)


oliviertassinari:
If we are going to fix the demo, I wonder if we shouldn't move forward with this approach at the same time: https://material-ui.com/components/autocomplete/#customizations - #1768. Maybe it will avoid us to break the demo a second time.

oliviertassinari:
What about we leverage the `focused` variable for the custom input demo? I think that it's an important UX element developers and designers need to be aware of.

dmtrKovalenko (approved):

PR #2045, opened 2020-07-28
Make all components generics 

dmtrKovalenko:
Unfortunately, our internal component's props keep returning `unknown`  for date values so we must refactor **all** the components to be generics. Here is the beginning of refactoring – but with all core changes for `makePickerWithState`


lib/src/views/Clock/ClockView.tsx
```diff
@@ -96,50 +98,57 @@ function getMinutesAriaText(minute: string) {
...
-}
+const getHoursAriaText = (hour: string) => `${hour} hours`
+
+const getSecondsAriaText = (seconds: string) => `${seconds} seconds`;
 
-function ClockViewRaw(props: ClockViewProps) {
+export function ClockView<TDate>(props: ClockViewProps<TDate>) {
   const {
-    allowKeyboardControl,
+    type,
```
oliviertassinari: sort asc

lib/src/views/Clock/Clock.tsx
```diff
@@ -9,19 +10,19 @@ import { VIEW_HEIGHT } from '../../constants/dimensions';
 import { ClockViewType } from '../../constants/ClockType';
 import { PickerOnChangeFn } from '../../_shared/hooks/useViews';
 import { getHours, getMinutes } from '../../_helpers/time-utils';
-import { withDefaultProps } from '../../_shared/withDefaultProps';
+import { useDefaultProps } from '../../_shared/withDefaultProps';
```
oliviertassinari, line 13: This is a step in the right direction 👍 . In hindsight, I wish I pushed more against `withDefaultProps` and `useStyles` in the past.

lib/src/views/Clock/Clock.tsx
```diff
@@ -95,7 +96,7 @@ export const useStyles = makeStyles(
   muiComponentConfig
 );
 
-export const Clock: React.FC<ClockProps> = withDefaultProps(muiComponentConfig, (props) => {
+export function Clock<TDate>(props: ClockProps<TDate>) {
```
oliviertassinari, line 99: Important question. What would it take to use withStyles here?

lib/src/_helpers/time-utils.ts
```diff
@@ -103,17 +103,17 @@ export const createIsAfterIgnoreDatePart = (
...
    * For input validation date part of passed object will be ignored if `disableIgnoringDatePartForTimeValidation` not specified.
    */
-  minTime?: unknown;
+  minTime?: ParsableDate<TDate>;
   /**
    * Max time acceptable time.
    * For input validation date part of passed object will be ignored if `disableIgnoringDatePartForTimeValidation` not specified.
    */
-  maxTime?: unknown;
+  maxTime?:  ParsableDate<TDate>;
```
oliviertassinari:
```suggestion
  maxTime?: ParsableDate<TDate>;
```

oliviertassinari (commented):
A note for later. 

I think that it would be great to explore the potential of only exposing a native date object or a string date in the public API. It seems that using a native date object has the potential to be a great way to remove the need for generics and to have examples that work with any date engines. The impact user-land seems small, people would only need to wrap the date with their library when they need to use it.

lib/src/views/Clock/Clock.tsx
dmtrKovalenko: I didn't get the question. What the problem of using `makeStyles` could be?

oliviertassinari: We will need to use the HOC API for integrating the components back into the lab. So we might as well do it now.

dmtrKovalenko: No we cannot use HOCs with pickers sources. Because it breaks generics

oliviertassinari: In this case, we need to reconsider the approach, what about we expose a native Date object to remove the need for generics? It's the approach used by https://github.com/adobe/react-spectrum/blob/c700898916bbd076bcc63e49d77c16d80623a8e7/packages/%40react-spectrum/calendar/src/Calendar.tsx (even if they are using date-fns), it seems simpler :)

oliviertassinari: Also, I wonder, isn't the Autocomplete using generic and withStyles at the same time?

dmtrKovalenko: Generics lives only on the typescript side – autocomplete written in js. 

it is possible to use withStyles. Simply by adding `withStyles() as typeof CalendarView` where `CalendarView` is generic. BUT why? I would be happy to make lab infrastructure to work without `withStylea`

oliviertassinari: > BUT why?

A couple of reasons that come to mind:

1. @mnajdova Is working on revamping how we write the components (at the style and unstyled level). I think that normalizing the approach will make her job easier. 
2. It bundles a couple of behavior that forces us to implement, which is great. `classes` API, theme default prop behavior.
3. It's required for the API classes markdown generation to work correctly.
4. It's required for the tests getClasses utility to work correctly.

> I would be happy to make lab infrastructure to work without withStylea

I think that we should leave this to 1. and focus on the migration for now. But if you have insight, please raise them to Marija :).

oliviertassinari (approved):

PR #2048, opened 2020-07-29
[docs] Fix guides links to point to next

### LINKED ISSUE #2046: Broken link in the docs
https://v4-0-0-alpha-10.material-ui-pickers.dev/api/DateRangePicker

The link provided for the prop `dateAdapter` is broken

broken link for **here**
> Allows to pass configured date-io adapter directly. More info **here**

aamirafridi:

PR to fix [this bug](https://github.com/mui-org/material-ui-pickers/issues/2046)

### DIFF
--- a/docs/prop-types.json
+++ b/docs/prop-types.json
@@ -1178,7 +1178,7 @@
     },
     "dateAdapter": {
       "defaultValue": null,
-      "description": "Allows to pass configured date-io adapter directly. More info [here](https://material-ui-pickers.dev/guides/date-adapter-passing)\n```jsx\ndateAdapter={new DateFnsAdapter({ locale: ruLocale })}\n```",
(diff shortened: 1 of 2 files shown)


dmtrKovalenko (approved):

oliviertassinari (commented):
What about the other cases where the URL is wrong? For instance:

https://github.com/mui-org/material-ui-pickers/blob/df3010bee046034178b5859ce4fd81755cb1d5e4/lib/src/views/Calendar/Calendar.tsx#L24

Also, to be noted that in the main repository, we remove the URL prefix to have the link work no matter the URL it's live on. This comes with one tradeoff, it doesn't work inside IntelliSense. 

I think that we should remove the prefixes to prepare the migration in the main repo.

PR #2053, opened 2020-07-31
Require Material-UI v5 alpha

eps1lon:
Getting an overview...


oliviertassinari (approved):

oliviertassinari:
We might want to update 

https://github.com/mui-org/material-ui-pickers/blob/ff3368340301d8e16b9b42781b6d02487ca07fc0/lib/package.json#L3

to 

```
5.0.0-alpha.11
```

eps1lon:
I wouldn't skip a version. That is always confusing.

oliviertassinari:
Ok, fair enough

oliviertassinari:
Released as v4.0.0-alpha.11

Issue #1650, opened 2020-04-12
The demos should work in isolation


oliviertassinari:
## Environment

| Tech                 | Version |
| -------------------- | ------- |
| @material-ui/pickers | v4.0.0-alpha.5 |

## Steps to reproduce

1. Go to https://next.material-ui-pickers.dev/demo/datepicker
2. Copy and paste the source in codesandbox
3. It crashes

> Can not find utils in context. It looks like you forgot to wrap your component in LocalizationProvider, or pass dateAdapter prop directly.

## Expected behavior

The demos should all be working in isolation, no matter the cost. I think that it's a critical point required to minimize churn during the first few minutes a developer try this option (competing libraries) out.

dmtrKovalenko:
What do you prefer? Wrap each demo in `LocalizationProvider` or pass `dateAdapter` prop?

oliviertassinari:
I think that we should have both:
- Have LocalizationProvider
- Create the utils from the most popular lib, likely date-fns.

Regarding testing with different date engines, I imagine it's much more a library maintainer concern than a developer concern. However, I think that the demos could accept an optional util prop coming from the outside.

Would it work? My only concern is with the overhead of these the demos. But I think that it's healthy, it incentive us to make configuration as easy as possible for developers.

oliviertassinari:
We will also need to remove all imports from docs folders, for instance
```jsx
import { makeJSDateObject } from "../../../utils/helpers";
```

dmtrKovalenko:
I mean using context provider in each example or pass dateAdapter right to the pickers? https://next.material-ui-pickers.dev/guides/date-adapter-passing

oliviertassinari:
@dmtrKovalenko What about this structure?

- **The documentation**. It injects a `dateAdapter` prop to the demo. It's something we do in the core in https://github.com/mui-org/material-ui/blob/ee9974955edb64081b7242baa1483c451178422b/docs/src/pages/components/drawers/ResponsiveDrawer.js#L169-L175
- **The demo**. It creates a default dateAdapter with date-fns, if the component has a parent `dateAdapter` provided, we use it, otherwise, we default to the local version. It make this prop available to the demos with a LocalizationProvider. Proof of concept:

```jsx
import React from "react";
import LocalizationProvider from "@material-ui/pickers/LocalizationProvider";
import DesktopDatePicker from "@material-ui/pickers/DesktopDatePicker";
[...]

PR #2054, opened 2020-08-01
chore: Fix various peer dependency warnings

eps1lon:
Please don't ignore these warnings. They're almost always correct.

It's especially important to align the main monorepo and the pickers monorepo. This should make integrating easier since the workspaces in pickers no longer rely on the whole dependency tree but can live on their own.


docs/package.json
```diff
@@ -17,6 +17,8 @@
     "node": ">=8.0.0"
   },
   "dependencies": {
+    "@babel/core": "^7.9.6",
```
eps1lon, line 20: Fixes "missing peer"-warning.

docs/package.json
```diff
@@ -17,6 +17,8 @@
...
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
```
eps1lon, line 21: Was previously declared in the root but only used in the docs workspace.

docs/package.json
```diff
@@ -25,6 +27,7 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.54",
     "@material-ui/pickers": "^4.0.0-alpha.1",
+    "@mdx-js/mdx": "^0.15.7",
```
eps1lon, line 30: Fixes "missing peer"-warning

package.json
```diff
@@ -70,5 +74,9 @@
     "/docs/_shared",
     "/e2e/",
     "**/build"
-  ]
+  ],
+  "resolutions": {
+    "**/@babel/core": "^7.9.6",
+    "**/webpack": "^4.43.0"
```
eps1lon, line 80: Technically `next` should declare it as a peer dependency with a default but that isn't supported in most package managers. In yarn v2 we will leverage `packageExtensions` instead of `resolutions`

package.json
```diff
@@ -58,6 +59,9 @@
     "husky": "^4.2.5",
     "lint-staged": "^10.2.11",
     "prettier": "^2.0.5",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
+    "typescript": "^3.9.6",
```
eps1lon, line 64: Fixes "missing peer"-warning.

package.json
```diff
@@ -33,9 +33,10 @@
     "url": "git+https://github.com/mui-org/material-ui-pickers.git"
   },
   "devDependencies": {
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
+    "@babel/preset-env": "^7.9.6",
     "@babel/runtime": "^7.8.4",
     "@cypress/webpack-preprocessor": "^4.1.0",
+    "@material-ui/core": "^4.11.0",
```
eps1lon, line 39: Fixes "missing peer"-warning.

package.json
```diff
@@ -33,9 +33,10 @@
     "url": "git+https://github.com/mui-org/material-ui-pickers.git"
   },
   "devDependencies": {
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
+    "@babel/preset-env": "^7.9.6",
```
eps1lon, line 36: Fixes "missing peer"-warning.

lib/package.json
```diff
@@ -99,18 +99,15 @@
...
     "babel-plugin-optimize-clsx": "^2.6.1",
     "babel-plugin-react-remove-properties": "^0.3.0",
     "codecov": "^3.7.1",
     "cross-env": "^7.0.2",
     "date-fns": "^2.12.0",
+    "dayjs": "^1.8.27",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
-    "eslint-plugin-react": "^7.19.0",
-    "eslint-plugin-react-hooks": "^4.0.5",
```
eps1lon, line 113: These are used in the `.eslintrc` in the root. Therefore they should be declared in the root package.json.

lib/package.json
```diff
@@ -99,18 +99,15 @@
...
     "@types/react-transition-group": "^4.2.4",
-    "@typescript-eslint/eslint-plugin": "^1.6.0",
-    "@typescript-eslint/parser": "^1.6.0",
     "babel-loader": "^8.1.0",
     "babel-plugin-optimize-clsx": "^2.6.1",
     "babel-plugin-react-remove-properties": "^0.3.0",
     "codecov": "^3.7.1",
     "cross-env": "^7.0.2",
     "date-fns": "^2.12.0",
+    "dayjs": "^1.8.27",
```
eps1lon, line 108: Fixes "missing peer"-warning. Worked incidentally since it was hoisted from the docs workspace.

lib/package.json
```diff
@@ -99,18 +99,15 @@
     "@types/react": "^16.9.35",
     "@types/react-dom": "^16.9.8",
     "@types/react-transition-group": "^4.2.4",
-    "@typescript-eslint/eslint-plugin": "^1.6.0",
-    "@typescript-eslint/parser": "^1.6.0",
```
eps1lon, line 103: These are used in the `.eslintrc` in the root. Therefore they should be declared in the root package.json which was already the case. 1.x was unused but  keeping them around can be dangerous with a lax package manger.

oliviertassinari:
@eps1lon Regarding the migration plan of pickers into the main repository, the current proposed candidate approach is:

1. We hard merge pickers and main into a WIP repository, basically something identical to https://github.com/oliviertassinari/mui-merge/tree/master/packages/material-ui-pickers but with the latest code.
1. We set up the CI on this new temporary repository.
1. We drop all the tooling infrastructure from pickers (making the peer dependencies issues you are going after here vanish).
1. We work on getting the CI green, moving all the components into the `/packages/material-ui-lab` folder, moving the documentation into the `/docs` folder.
1. Once the CI is green, we backport the changes we had to do the main infrastructure to the main repository.
1. We do 1. once more, but this time for real, with the main repo, hopefully, it will Just work™.
1. We transfer all the existing issues of the pickers in the main repository.

What do you think about it?

eps1lon:
I need to think about this a bit more. It seems like you're making things more complicated than necessary. There bigger, fundamental issues: how are the docs pages merged? Is pickers going core or lab?

This PR is important regardless of merging. Moving workspaces is just on use case that reveals issues with undeclared dependencies.

oliviertassinari:
> I need to think about this a bit more

Please do :)

> how are the **docs** pages merged?

The documentation can be migrated, to the format of the main repository. We would need to replace the mdx syntax. But the demos are close to being usable as-is. We would still need to fix #1650 (e.g. use *date-fns* by default in the demos). But that isn't a blocker, it can be done incrementally.

I think that the main concern is the same one as we have with the data grid. How do we organize the navigation for components that require more than 1 page? 

> Is pickers going core or lab?

Maybe in the lab first? So it doesn't block us in 6 months if we are ready to cut a stable release but pickers aren't.

dmtrKovalenko (approved):
Thanks 😊

Issue #1898, opened 2020-06-15
[RFC] Inline date-io inside the date picker


oliviertassinari:
Answering #1895 made me think again of a thought I had a few months ago. date-io has been around for 2 years, with the ambition to be used outside of the use cases of Material-UI.

I think that we can take a step back and look at the potential at this market, and how it has delivered. npm lists (approximately) the most downloaded packages, `@date-io/core` is a dependency of in https://www.npmjs.com/browse/depended/@date-io/core. I couldn't find one package that has over 1k downloads/month and doesn't have `@material-ui/pickers` as a dependency.

Do you have an example package that would match these criteria? If not, I think that we could conclude that there is very limited potential for it. And think about what would it look like if date-io was merged inside the main repository too. I'm curious to see what the outcome of https://github.com/gpbl/react-day-picker/issues/980 will be.

As we plan to expand the usage of date-io to more components inside Material-UI, I start to think that hosting the source at the same place would allow us to be more efficient, shaping better API, avoiding the release synchronization pain.

JRagone:
While I'm not a contributor to this repo, as a user of pickers I'm strongly in favor of a merge with `date-io`. I recently ran into #1440, and from the discussion in that thread, it seemed like @oliviertassinari was suggesting that there's no reason to upgrade the `date-io` version, because in all likelihood no other packages are dependent on it. If this is the case, i.e. pickers is strongly dependent on `date-io` and no other packages depend on `date-io`, then it seems merging the two would be a very logical step, as it resolves the aforementioned release synchronization pain.

oliviertassinari:
@JRagone The root issue of #1440 was solved (in the `next` branch) by hiding the dependency to developers. It won't happen again in the future. I don't think this is an argument we should consider for a merge.

oliviertassinari:
> pickers is strongly dependent on date-io and **no other packages** depend on date-io, then it seems merging the two would be a very logical step, as it resolves the aforementioned **release synchronization pain**.

Yes, this argument, I can get behind with 👍

oliviertassinari:
Let's do that in v5, we need to have a close feedback loop. We can probably duplicate the source, and iterate on them once in the same repo.

PR #2055, opened 2020-08-01
[core] Lift peer dependencies of date libs

eps1lon:
Lifts peer dependencies on date utils as optional. Otherwise you currently get a lot of missing peer dependency warnings when install v4:

```bash
warning "workspace-aggregator-65b683a5-1b32-4e5f-bd47-27b2d2dc6468 > docs > @material-ui/pickers > @date-io/dayjs@2.8.0" has unmet peer dependency "dayjs@^1.8.17".
warning "workspace-aggregator-65b683a5-1b32-4e5f-bd47-27b2d2dc6468 > docs > @material-ui/pickers > @date-io/luxon@2.8.0" has unmet peer dependency "luxon@^1.21.3".
warning "workspace-aggregator-65b683a5-1b32-4e5f-bd47-27b2d2dc6468 > docs > @material-ui/pickers > @date-io/moment@2.8.0" has unmet peer dependency "moment@^2.24.0".
```

Peer dependencies lifted:

- `@date-io/date-fns`: https://github.com/dmtrKovalenko/date-io/blob/v2.8.0/packages/date-fns/package.json#L32-L34
- `@date-io/dayjs`: https://github.com/dmtrKovalenko/date-io/blob/v2.8.0/packages/dayjs/package.json#L8-L10
- `@date-io/luxon`: https://github.com/dmtrKovalenko/date-io/blob/v2.8.0/packages/luxon/package.json#L11-L13
- `@date-io/moment`: https://github.com/dmtrKovalenko/date-io/blob/v2.8.0/packages/moment/package.json#L8-L10

Ideally we'd warn if none is fulfilled but in the end we don't know which adapter is used at install time. However, having too many warnings numbs users to ignore these which is worse

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -35,6 +35,10 @@
     "@material-ui/core": "^4.10.1",
     "@material-ui/lab": "^4.0.0-alpha.54",
     "@types/react": "^16.8.6",
+    "date-fns": "^2.0.0",
+    "dayjs": "^1.8.17",
+    "luxon": "^1.21.3",
+    "moment": "^2.24.0",
     "react": "^16.8.4",
     "react-dom": "^16.8.4"
   },
@@ -43,6 +47,18 @@
       "optional": true
     },
     "@material-ui/lab": {
+      "optional": true
+    },
+    "date-fns": {
+      "optional": true
+    },
+    "dayjs": {
+      "optional": true
+    },
+    "luxon": {
+      "optional": true
+    },
+    "moment": {
       "optional": true
     }
   },


oliviertassinari (commented):
Would it work? It seem that the peer dependencies warnings are coming from transitive dependencies?

Capture d’écran 2020-08-01 à 15 09 32

#1898 might be relevant in this context

eps1lon:
> Would it work?

What would work?

NMinhNguyen:
> Would it work? It seem that the peer dependencies warnings are coming from transitive dependencies?
> 
> 
> 
> Capture d’écran 2020-08-01 à 15 09 32
> 
> 
> 
> #1898 might be relevant in this context

I wonder if that's a Yarn bug, or maybe Yarn v1 bug?

eps1lon:
I don't see a bug there. Declared peer dependencies are not fulfilled.

NMinhNguyen:
> I don't see a bug there. Declared peer dependencies are not fulfilled.

I thought they would be fulfilled via `devDependencies` in the monorepo? And the fact you're specifying them as `peerDependencies` should help Yarn too.

oliviertassinari:
> What would work?

I wasn't aware that peer dependency warnings from depended on packages can be "lifted" like this. Happy to try it out :)

NMinhNguyen:
> > What would work?
> 
> 
> 
> I wasn't aware that peer dependency warnings from depended on packages can be "lifted" like this. Happy to try it out :) 

Ah I thought your screenshot was from after this diff was applied, and it was still generating warnings. In any case, lifting is something I discovered as part of experimenting with Yarn v2 where it's very sensitive to peer deps. You can read more about it here https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0

oliviertassinari (approved):

dmtrKovalenko (approved):

eps1lon:
FYI: Lifting peer dependencies only works in yarn v2. yarn v1 will still complain.

PR #2119, opened 2020-09-21
[YearSelection] Add order prop

MaximMorozCredentially:
When working with your library, I noticed that there is no way to list the years in reverse order, so I tried to fix that =)


oliviertassinari:
@MaximMorozCredentially Thanks for the effort, however, I don't think that we will accept such change until there is enough support for it from the community. We can't grow the API indefinitely. Feel free to open an issue about it on https://github.com/mui-org/material-ui to support the idea. For now, we are archiving this repository, so closing.

PR #2155, opened 2020-11-08
Feature: Allow custom weekday header formatting

djemeljanovs:
There should be an option to have more than a single character in weekday name headers.


oliviertassinari:
@djemeljanovs  Thanks for the pull request. Unfortunately, we can't accept it here. On-going work was moved to v5 in the main repository: https://next.material-ui.com/components/date-picker/.

If you could re-evaluate the need for the change there, it would be great :). In the event the problem still needs to be fixed, please start by opening an issue so we can understand the problem and challenge the best solution.

PR #2158, opened 2020-11-17
Update usage.mdx

getsetbro:
Import DateFnsAdapter but then in use as DateFnsUtils.


oliviertassinari:
@getsetbro Thanks for the pull request. Unfortunately, we can't accept it. On-going work was moved to v5: https://next.material-ui.com/components/date-picker/. If you could re-evaluate the need for the change there, it would be great :).

PR #2174, opened 2021-02-22
Allow React 17 in the peer dependencies

### LINKED ISSUE #2179: Bump supported React version to 17?
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-pickers/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

If it is possible, can we bump the supported React version to 17? AFAIK, React didn't break anything with this major update

## Motivation 🔦

Since v7 npm throws errors during installation if any of deps/peerDeps are not matching. One has to execute with `-force` flag, which is not straightforward.

doolse:
Npm 7 doesn't like installing without using --legacy-peer-deps, so it would be nice to ot have to do that

- [x ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui-org/material-ui/blob/master/CONTRIBUTING.md#sending-a-pull-request).

Fixes #2179

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -35,8 +35,8 @@
     "@date-io/core": "^1.3.6",
     "@material-ui/core": "^4.0.0",
     "prop-types": "^15.6.0",
-    "react": "^16.8.4",
-    "react-dom": "^16.8.4"
+    "react": "^16.8.4 || ^17",
+    "react-dom": "^16.8.4 || ^17"
   },
   "dependencies": {
     "@babel/runtime": "^7.6.0",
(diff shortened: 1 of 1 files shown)


oliviertassinari (approved):

thebinarysearchtree:
You can do `npm install --legacy-peer-deps @material-ui/pickers` but that isn't where the problems end. Once you have installed that, you will find that it doesn't work with the latest date libraries either. Essentially there is no working date picker in the current version of material ui. I am going to make my own because this project is dead and the next version of material ui puts the date pickers into the lab so they are assumed to not be stable.

